PR #10, opened 2022-05-11
feat: onSuccess callback prop

SamueleA:
Added an onSuccess prop that triggers a callback once a swap is successful

### DIFF
--- a/src/components/Widget.tsx
+++ b/src/components/Widget.tsx
@@ -93,10 +93,11 @@ export type WidgetProps = {
   dialog?: HTMLElement | null
   className?: string
   onError?: ErrorHandler
+  onSuccess?: () => void
 }
 
 export default function Widget(props: PropsWithChildren<WidgetProps>) {
-  const { children, theme, provider, jsonRpcEndpoint, dialog: userDialog, className, onError } = props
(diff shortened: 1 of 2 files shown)


zzmp:
Thanks, @SamueleA! Because the SwapWidget is capable of performing multiple transactions (including token permits and approvals), I think that the onSuccess callback needs to convey some information on the successful transaction. Could you open an issue to discuss this (and link this PR!) so that we can iron that out and move forward with transaction callbacks?

SamueleA:
@zzmp Thanks for replying! I've opened an issue as requested https://github.com/Uniswap/widgets/issues/11

SamueleA:
Made the onSuccess callback only trigger for SWAP and WRAP calls
[1eb9f9b](https://github.com/Uniswap/widgets/pull/10/commits/1eb9f9b23b8528acb42e8f3ab60bba7e2bb2dd07)

zzmp:
I'm not sure if we've documented it yet, but we now support onTxSuccess callbacks.
See https://github.com/Uniswap/widgets/blob/main/src/hooks/transactions/index.tsx#L64-L72.

PR #12, opened 2022-05-14
test(e2e): connect

zzmp:
- Configures e2e testing
- Adds a test for connection states
- Removes jotai from web-react Provider


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  e2e-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: ./.github/actions/setup
+
+      - run: yarn hardhat &
```
JFrankfurt, line 19: is this guaranteed to finish setup before a test needs it? might need to wait?

cosmos.config.json
```diff
@@ -1,7 +1,7 @@
 {
   "watchDirs": ["src/cosmos"],
   "webpack": {
-    "overridePath": "webpack.override.cjs"
+    "overridePath": "webpack.override.js"
```
JFrankfurt, line 4: 🎉

e2e/connect.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { JsonRpcProvider } from '@ethersproject/providers'
+import { act, render, RenderResult } from '@testing-library/react'
+import { tokens } from '@uniswap/default-token-list'
+
+import { SwapWidget } from '../src'
+
+const JSON_RPC_ENDPOINT = 'http://127.0.0.1:8545/'
+const provider = new JsonRpcProvider('http://127.0.0.1:8545/')
+
+class MockPromise<T = void> {
```
JFrankfurt, line 12: why is a mock promise necessary? (I'm not familiar w/ the `act` fn below, forgive me if this should be obv

hardhat.config.js
```diff
@@ -0,0 +1,13 @@
...
+}
+
+module.exports = {
+  networks: {
+    hardhat: {
+      chainId: 1,
+      forking: mainnetFork,
+    },
+  },
+}
```
JFrankfurt, line 13: not important, but imo we should pass in our own private key value(s) here -- I don't like using the hardhat defaults, just in case someone makes a mistake with them in the future. I'd rather make it explicit that we're using `SUPER_COMPROMISED_PRIVATE_KEY_1` etc.

src/components/Widget.tsx
```diff
@@ -93,10 +93,11 @@ export type WidgetProps = {
   dialog?: HTMLElement | null
   className?: string
   onError?: ErrorHandler
+  onConnect?: (provider: JsonRpcProvider) => void
```
JFrankfurt, line 96: is this supposed to be `onConnectWallet`? 
image

JFrankfurt: https://docs.uniswap.org/sdk/widgets/swap-widget/api

JFrankfurt (approved):
some nits and questions, but lgtm

CONTRIBUTING.md
```diff
@@ -19,6 +19,11 @@ Before developing locally, be sure to install the project's dependencies.
 1. `yarn test`
 - You may also run `yarn test --watch` for feedback in realtime.
 
+Integration tests are run through a separate command, and require a hardhat node to be running:
+
+1. `yarn hardhat &`
+1. `yarn test:e2e`
```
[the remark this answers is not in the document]
zzmp: I'm working on this rn for a follow-up PR. [In the past, I've done this with concurrently](https://github.com/Uniswap/interface/blob/e376885dfddd15370a24ef4163a776e6a2540add/package.json#L133-L135), but I think it'd be more correct (and possibly aid in deterministic testing in the future) to do it as part of jest setup.

src/components/Widget.tsx
zzmp: This is a separate callback that I added to facilitate sequencing in testing. `onConnectWallet` is triggered when a user attempts to connect a wallet through pressing the "Connect your wallet" button; this is triggered when any provider is successfully connected.

e2e/connect.test.tsx
zzmp: I find this pattern makes sequencing more clear, instead of having to have the boilerplate of a Promise within the test.

hardhat.config.js
zzmp: hardhat makes it very explicit that these are compromised when it starts up.

src/components/Widget.tsx
JFrankfurt: I really don't like adding undocumented API surface explicitly for the purpose of testing. If possible (imo) we should only test using the attributes naturally available to integrators and users. 

In this case, I'd make an assertion about expected UI state instead of adding a new prop to test.

PR #15, opened 2022-05-25
fix: set isPending false if wrap tx errors

kristiehuang:
before: (wrap button stays in spinning state even after user rejects tx)
Screen Shot 2022-04-13 at 6 19 11 PM

after:
Untitled

### DIFF
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -92,7 +92,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
       if (isAnimating(document)) {
         document.addEventListener('animationend', postWrap)
       } else {
-        postWrap()
+        setIsPending(false)
       }
     } catch (e) {
       // TODO(zzmp): Surface errors from wrap.
@@ -133,7 +133,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
       if (isAnimating(document)) {
         document.addEventListener('animationend', postSwap)
       } else {
-        postSwap()
+        setOpen(false)
       }
     } catch (e) {
       // TODO(zzmp): Surface errors from swap.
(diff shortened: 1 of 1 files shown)


src/components/Swap/SwapButton/index.tsx
```diff
@@ -84,6 +84,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
       })
       setDisplayTxHash(transaction.hash)
     } catch (e) {
+      setIsPending(false)
```
zzmp, line 87: Why don't 'postWrap' cover this case?

src/components/Swap/SwapButton/index.tsx
```diff
@@ -84,16 +89,13 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
       })
       setDisplayTxHash(transaction.hash)
     } catch (e) {
+      postWrap()
```
zzmp, line 92: This will occur before the event listener is added.

Why doesn't postWrap cover the case of a failed transaction already, without this PR?

src/components/Swap/SwapButton/index.tsx
```diff
@@ -83,21 +83,22 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
+      // Only reset pending after any queued animations to avoid layout thrashing, because a
+      // successful wrap will open the status dialog and immediately cover the button.
+      const postWrap = () => {
+        setIsPending(false)
+        document.removeEventListener('animationend', postWrap)
+      }
+      if (isAnimating(document)) {
+        document.addEventListener('animationend', postWrap)
+      } else {
+        postWrap()
```
zzmp, line 95: This can just be `setIsPending(false)`, because it doesn't need to remove an event listener.

zzmp (approved):

PR #23, opened 2022-06-02
test: utilities for unit testing components

zzmp:
- Adds utilities for unit testing components
- Typechecks all tests (by moving the build tsconfig to its own `tsconfig.build.json` file)


src/cosmos/Swap.fixture.tsx
```diff
@@ -8,18 +12,14 @@ import {
   SUPPORTED_LOCALES,
   SupportedChainId,
   SwapWidget,
-} from '@uniswap/widgets'
-import { useEffect } from 'react'
-import { useValue } from 'react-cosmos/fixture'
-
-import { DAI, USDC_MAINNET } from '../constants/tokens'
+} from '../index'
```
JFrankfurt, line 15: this seems fine, but curious what the change gets us?

src/cosmos/Swap.fixture.tsx
```diff
@@ -41,7 +41,7 @@ function Fixture() {
   const [width] = useValue('width', { defaultValue: 360 })
 
   const locales = [...SUPPORTED_LOCALES, 'fa-KE (unsupported)', 'pseudo']
-  const locale = useOption('locale', { options: locales, defaultValue: DEFAULT_LOCALE, nullable: false })
+  const locale = useOption<string>('locale', { options: locales, defaultValue: DEFAULT_LOCALE, nullable: false })
```
JFrankfurt, line 44: is locale not typed implicitly by the initialized value?

tsconfig.json
```diff
@@ -24,10 +24,11 @@
     "skipLibCheck": true,
     "strict": true,
     "strictNullChecks": true,
+    "suppressImplicitAnyIndexErrors": true,
```
JFrankfurt, line 27: this feels like overkill (related to dropping the SupportedChainId?) can you just use ts-ignore where you need it instead?

src/cosmos/Swap.fixture.tsx
zzmp: `@uniswap/widgets` is self-referential, so it's outside of the type system. I can probably add it as a path, though, which might make this look a little nicer.

PR #38, opened 2022-06-17
feat: Add WalletConnect QR code to ConnectWalletDialog with latest web3-react

kristiehuang:
Adds QR code functionality to WalletConnect. Also updates web3-react to latest version.

Known minor issues:
* After MetaMask disconnection, widget will still use MetaMask's JSON RPC connector to get price quotes. Doesn't cause issues in user experience tho. This is bc of web3-react's `useIsActive()` hook not working entirely as we'd expect it to
* Does not re-render when switching between accounts (also bc of web3-react's `useAccount()` hook)

Demo of WalletConnect QR working for both the popup modal & the tile QR:

qr


package.json
```diff
@@ -154,8 +156,8 @@
     "@uniswap/v3-core": "1.0.0",
     "@uniswap/v3-periphery": "^1.1.1",
     "@walletconnect/ethereum-provider": "^1.7.1",
-    "@web3-react/metamask": "8.0.24-beta.0",
-    "@web3-react/walletconnect": "8.0.31-beta.0",
+    "@web3-react/metamask": "8.0.26-beta.0",
```
zzmp, line 159: These (@walletconenct/ethereum-provider, @web3-react/metamask, @web3-react/walletconnect) should be moved to dependencies (not devDeps).

src/cosmos/useJsonRpcEndpoint.ts
```diff
@@ -7,24 +7,24 @@ if (INFURA_KEY === undefined) {
   console.warn(`INFURA_KEY must be a defined environment variable to use JsonRpcEndpoints in the cosmos viewer`)
 }
 
-export const INFURA_NETWORK_URLS: { [key in SupportedChainId]?: string } = INFURA_KEY
+export const INFURA_NETWORK_URLS: { [key in SupportedChainId]?: string | string[] } = INFURA_KEY
```
zzmp, line 10: Rm `string` - just `string[]`.

src/cosmos/useJsonRpcEndpoint.ts
```diff
@@ -35,5 +35,5 @@ export default function useJsonRpcEndpoint() {
   return useOption('jsonRpcEndpoint', {
     options: endpoints,
     defaultValue: INFURA_NETWORK_URLS[SupportedChainId.MAINNET] ? SupportedChainId[SupportedChainId.MAINNET] : NONE,
-  })
+  })?.at(0)
```
zzmp, line 38: What does `at` compile to? If it compiles to `at`, and not `[]`, then prefer to use `[]` directly (as it is more widely available).

zzmp (changes requested):
Should merging be deferred until https://github.com/NoahZinsmeister/web3-react/issues/591 is resolved, so that this doesn't bloat the widget size?

zzmp:
Design question: Should the "Connect wallet to swap" button be more button-like? Should the text change at all?

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -56,43 +59,56 @@ function getWallet(provider?: JsonRpcProvider | Eip1193Provider) {
...
   const web3 = useMemo(() => {
     if (connector === EMPTY || !(active || activating)) {
       return EMPTY_WEB3
     }
-    return { connector, library, chainId, accounts, account, active, activating, error }
-  }, [account, accounts, activating, active, chainId, connector, error, library])
+    return { connector, library, chainId, accounts, account, active, activating }
+  }, [account, accounts, activating, active, chainId, connector, library])
 
   const updateWeb3 = (updateContext: Web3ContextType) => {
```
zzmp, line 113: I'm afraid this might cause frequent updates. Can the updater be moved to a separate context?

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -46,18 +58,93 @@ interface ButtonProps {
...
+      ;(tileConnector.provider?.connector as unknown as EventEmitter | undefined)?.off('display_uri', uriListener)
+    }
+  })
+
+  async function formatQrCodeImage(uri: string) {
+    let result = ''
+    const dataString = await QRCode.toString(uri, { margin: 0, type: 'svg' })
+    if (typeof dataString === 'string') {
+      result = dataString.replace(
+        '<svg',
```
zzmp, line 127: Is there a safer way to do these html transforms?

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -46,18 +58,93 @@ interface ButtonProps {
...
+
+    return () => {
+      tileConnector.events.off(URI_AVAILABLE)
+      ;(tileConnector.provider?.connector as unknown as EventEmitter | undefined)?.off('display_uri', uriListener)
+    }
+  })
+
+  async function formatQrCodeImage(uri: string) {
+    let result = ''
+    const dataString = await QRCode.toString(uri, { margin: 0, type: 'svg' })
```
zzmp, line 124: The QRCode that we end up with is not as pretty as the WC QR code. Do you know why?

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -46,18 +58,93 @@ interface ButtonProps {
...
   caption?: string
+  connection?: Web3Connection
   onClick: () => void
 }
 
-function WalletConnectButton({ walletName, logoSrc, caption, onClick }: ButtonProps) {
+const wcQRUriAtom = atom<string | undefined>(undefined)
+
+function WalletConnectButton({ walletName, logoSrc, caption, connection: wcTileConnection, onClick }: ButtonProps) {
+  const [tileConnector] = wcTileConnection as [WalletConnect, Web3ReactHooks]
```
zzmp, line 68: You switch between wcTileConnection and wcConnectionTile - stick with one name. They're in different contexts, so they can be re-used.

zzmp, line 68: The important thing here is wc, not tile.

zzmp (changes requested):

src/components/ConnectWallet/ConnectWalletDialog.tsx
kristiehuang: wdym by safer?

unfortunately this is literally how WalletConnect does it in their code - https://github.com/WalletConnect/walletconnect-legacy/blob/4563223a9fe2ff099e7a66788813697e09b1e35c/packages/modal/src/browser/components/QRCodeDisplay.tsx#L12

kristiehuang:
> Should merging be deferred until [NoahZinsmeister/web3-react#591](https://github.com/NoahZinsmeister/web3-react/issues/591) is resolved, so that this doesn't bloat the widget size?

is there a plan for this to be resolved? I don't want to be blocked by an issue that's on the backlog

src/hooks/connectWeb3/useActiveWeb3React.tsx
kristiehuang: hm - could you clarify this? wdym by "risk extra renders by instantiating objects multiple times"

kristiehuang:
> Design question: Should the "Connect wallet to swap" button be more button-like? Should the text change at all?

pinged Phil about this!

src/components/ConnectWallet/ConnectWalletDialog.tsx
JFrankfurt: I think what you have currently complies with the designs: https://www.figma.com/file/A7w2snzxyFK2qj2aFpkkm1/Widgets?node-id=6057%3A39728

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -46,18 +58,93 @@ interface ButtonProps {
...
-        </ThemedText.Subhead1>
-      </ButtonContents>
+      <StyledMainButtonRow>
+        <ButtonContents>
+          <img src={logoSrc} alt={walletName} key={walletName} width={32} />
+          <ThemedText.Subhead1>
+            <Trans>{walletName}</Trans>
+          </ThemedText.Subhead1>
+          <ThemedText.Caption color="secondary">
+            <Trans>{caption}</Trans>
```
JFrankfurt, line 143: `<Trans>{dynamicContent}</Trans>` doesn't work.

src/components/ConnectWallet/index.tsx
```diff
@@ -1,25 +1,27 @@
-import { connections } from 'hooks/connectWeb3/useConnect'
+import { useWeb3React } from '@web3-react/core'
```
JFrankfurt, line 1: this is unused

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -101,16 +117,84 @@ export function ActiveWeb3Provider({
...
+  return [connector, hooks]
+}
+
+function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: string | JsonRpcProvider) {
+  // TODO(kristiehuang): implement RPC URL fallback, then jsonRpcEndpoint can be optional
+
+  // WalletConnect relies on Buffer, so it must be polyfilled.
+  if (!('Buffer' in window)) {
+    window.Buffer = Buffer
+  }
```
JFrankfurt, line 141: image

I think you should add this to the global window type

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -101,16 +117,84 @@ export function ActiveWeb3Provider({
...
+        chainId,
+      }
+      updateActiveWeb3ReactCallback(updateContext)
     }
-  }, [error])
+  }, [account, accounts, activating, chainId, isActive, library, updateActiveWeb3ReactCallback, wallet])
 
-  return <Web3Context.Provider value={{ web3, updateWeb3 }}>{children}</Web3Context.Provider>
+  return useWallet
 }
```
JFrankfurt, line 200: @zzmp can you take a look at this?

@kristiehuang The name here makes it seem like this is a hook (`use______`), and it's using hooks internally, but the usage in `ConnectWalletDialog` is as a factory for connection handlers.

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -46,18 +58,93 @@ interface ButtonProps {
...
 }
 
-function WalletConnectButton({ walletName, logoSrc, caption, onClick }: ButtonProps) {
+const wcQRUriAtom = atom<string | undefined>(undefined)
+
+function WalletConnectButton({ walletName, logoSrc, caption, connection: wcTileConnection, onClick }: ButtonProps) {
+  const [walletConnect] = wcTileConnection as [WalletConnect, Web3ReactHooks]
+  const [error, setError] = useState(undefined)
+
+  const [QRUri, setQRUri] = useAtom(wcQRUriAtom)
```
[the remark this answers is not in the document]
kristiehuang: no :(

src/hooks/connectWeb3/useActiveWeb3React.tsx
kristiehuang: I'm not getting this warning - are you using `"@types/node": "^16.7.13"` ?

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -87,18 +174,20 @@ function NoWalletButton() {
 }
 
 export function ConnectWalletDialog() {
-  const [mmConnection, wcConnection] = connections
+  const [mmConnection, wcTileConnection, wcPopupConnection] = connections
   // TODO(kristiehuang): what happens when I try to connect one wallet without disconnecting the other?
```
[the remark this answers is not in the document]
kristiehuang: works fine - but depends on w3r's `usePriorityConnector` hook, which returns "the first of the initializedConnectors that is active"

i.e. if `const initializedConnectors = [metaMask, walletConnect]`
* I connect WalletConnect first, then also connect MetaMask -- it switches to MetaMask wallet
* I connect MetaMask first, then also connect WalletConnect -- no switch is made

src/hooks/connectWeb3/useActiveWeb3React.tsx
kristiehuang: could I just rename it to `useWalletCallback`?

zzmp: Usage is correct, but naming is off (and I think you inherited this).
@kristiehuang could you change `useWallet`->`activateWallet`, and `useConnect`->`useConnectCallback`?

zzmp:
This does not build for me locally. Note that the checks are not being run because this is not a PR into `main`.

src/hooks/connectWeb3/useActiveWeb3React.tsx
zzmp: > > Should merging be deferred until [NoahZinsmeister/web3-react#591](https://github.com/NoahZinsmeister/web3-react/issues/591) is resolved, so that this doesn't bloat the widget size?
> 
> is there a plan for this to be resolved? I don't want to be blocked by an issue that's on the backlog

We've gotten pushback from integrators in the past about the size of this package, so I would investigate this before moving forward. It shouldn't block this PR, but it may need to block wallet-connect-flow from merging into main, or it may need to be noted in the release notes.

zzmp:
> > Should merging be deferred until [NoahZinsmeister/web3-react#591](https://github.com/NoahZinsmeister/web3-react/issues/591) is resolved, so that this doesn't bloat the widget size?
> 
> is there a plan for this to be resolved? I don't want to be blocked by an issue that's on the backlog

We've gotten pushback from integrators in the past about the size of this package, so I would investigate this before moving forward. It shouldn't block this PR, but it may need to block wallet-connect-flow from merging into main, or it may need to be noted in the release notes.

src/components/ConnectWallet/index.tsx
```diff
@@ -1,25 +1,26 @@
...
-  onConnectWallet?: (e?: React.MouseEvent<HTMLButtonElement>) => void
+  onClickConnectWallet?: (e?: React.MouseEvent<HTMLButtonElement>) => void
 }
 
-export default function Wallet({ disabled, account, onConnectWallet }: WalletProps) {
+export default function Wallet({ disabled, onClickConnectWallet }: WalletProps) {
   // Attempt to connect eagerly on mount
   useEffect(() => {
-    connections.forEach(([wallet, _]) => wallet.connectEagerly())
+    connections.forEach(([wallet, _]) => wallet?.connectEagerly?.())
```
zzmp, line 15: Can you add a comment here that this will immediately trigger lazy loading of the wallets.

src/hooks/connectWeb3/useActiveWeb3React.tsx
zzmp: This is being recreated on each render, and then the value of the provider is being instantiated on each render.

Instead, make updateWeb3 a callback
```
const updateWeb3 = useCallback((updateContext: Web3ContextType) => {
```

and memoize the value so it is not recreated every time the provider renders
```
const value = useMemo(() => ({ web3, updateWeb3 }), [web3, updateWeb3])
return <Web3Context.Provider value={value}>{children}</Web3Context.Provider>
```

zzmp (changes requested):

src/components/ConnectWallet/ConnectWalletDialog.tsx
zzmp: The colors on the QR code should match the theme, and not just be white/black.
I'm fine with this being done in a follow-up PR, but it should be done before merging to `main`.

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -101,16 +117,84 @@ export function ActiveWeb3Provider({
...
+        })
+    )
+  )
+}
+
+function useActiveWalletProvider(): Web3Provider | undefined {
+  return getPriorityConnector(...connections).usePriorityProvider() as Web3Provider
+}
+
+export function useWalletCallback(connection: Web3Connection) {
```
vm, line 170: Do we need to use this hook instead of using the `Web3ReactProvider` and `useConnector` hook that's provided by `web3-react`? I think that could reduce a lot of the complexity

vm: It would also remove the need for `useActiveWalletProvider`

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -56,43 +59,56 @@ function getWallet(provider?: JsonRpcProvider | Eip1193Provider) {
...
   provider?: Eip1193Provider | JsonRpcProvider
 }
 
 export function ActiveWeb3Provider({
   jsonRpcEndpoint,
-  provider,
+  provider: propsProvider,
   children,
 }: PropsWithChildren<ActiveWeb3ProviderProps>) {
+  const metaMaskConnection = useMemo(
```
vm, line 82: if the connectors are completely memoized I would move them outside of the component, though I think we could switch to using the `Web3ReactProvider` as commented below

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -56,43 +59,56 @@ function getWallet(provider?: JsonRpcProvider | Eip1193Provider) {
...
+  connections = [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup]
+
   const network = useMemo(() => getNetwork(jsonRpcEndpoint), [jsonRpcEndpoint])
-  const wallet = useMemo(() => getWallet(provider), [provider])
+  const activeProvider = useActiveWalletProvider()
+  const wallet = useMemo(() => getWallet(propsProvider ?? activeProvider), [propsProvider, activeProvider])
 
   // eslint-disable-next-line prefer-const
-  let { connector, hooks } = wallet.hooks.useIsActive() || network === EMPTY_STATE ? wallet : network
+  let { connector, hooks } = wallet !== EMPTY_STATE ? wallet : network
```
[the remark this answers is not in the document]
kristiehuang: this gets handled a couple lines up

```
const activeProvider = useActiveWalletProvider() // which uses the priority selector
const wallet = useMemo(() => getWallet(propsProvider ?? activeProvider), [propsProvider, activeProvider])
```

src/components/ConnectWallet/index.tsx
kristiehuang: @zzmp is that different from the `// Attempt to connect eagerly on mount` comment?

@vm if the user isn't already connected to the page, I don't think we want to call activate unless the user clicks `Connect wallet`

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -45,19 +57,90 @@ const StyledNoWalletText = styled(ThemedText.Subhead1)`
...
+      walletConnect.events.off(URI_AVAILABLE)
+      ;(walletConnect.provider?.connector as unknown as EventEmitter | undefined)?.off('display_uri', uriListener)
+    }
+  })
+
+  async function formatQrCodeImage(uri: string) {
+    let result = ''
+    const dataString = await QRCode.toString(uri, { margin: 0, type: 'svg' })
+    if (typeof dataString === 'string') {
+      result = dataString.replace('<svg', `<svg class="walletconnect-qrcode_tile" alt="WalletConnect" width="100"`)
```
zzmp, line 125: The viewbox is set to `0 0 53 53`, so the width must be a multiple of 53 (ie 106) to prevent smushing the image (sorry idk the technical word). Does `QRCode` have an option to use `50`?

zzmp (changes requested):

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -45,19 +57,90 @@ const StyledNoWalletText = styled(ThemedText.Subhead1)`
...
+        <ButtonContents>
+          <img src={logoSrc} alt={walletName} width={32} />
+          <ThemedText.Subhead1>
+            <Trans>{walletName}</Trans>
+          </ThemedText.Subhead1>
+          <ThemedText.Caption color="secondary">
+            <Trans>Scan to connect your wallet. Works with most wallets.</Trans>
+          </ThemedText.Caption>
+        </ButtonContents>
+        <div dangerouslySetInnerHTML={{ __html: qrCodeSvg }}></div>
```
zzmp, line 142: Instead of doing this dangerously, can you output an image (using QRCode.toDataUrl) and then set the src of an img element?

PR #48, opened 2022-06-30
fix: make swap details row clickable

kristiehuang:
Previously, only the arrow button was clickable to open swap details. Now, this entire row is clickable:

swap gif


src/components/Expando.tsx
```diff
@@ -9,6 +9,10 @@ import styled from 'styled-components/macro'
 
 const HeaderColumn = styled(Column)`
   transition: gap 0.25s;
+
+  :hover {
+    cursor: pointer;
+  }
```
JFrankfurt, line 15: don't need the `:hover` here

JFrankfurt (approved):

PR #49, opened 2022-06-30
fix: use back button instead of X

kristiehuang:
We use an X icon to close dialog tabs, which creates a double X when the widget is hosted in an integrator modal.
* change X to a back arrow
* change animation from vertical to horizontal slide

Previously:
Screen Shot 2022-06-30 at 7 59 48 PM

Now:
slide-in (1)

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -63,7 +63,7 @@ interface IconButtonProps {
 export const IconButton = forwardRef<HTMLButtonElement, IconButtonProps & ComponentProps<typeof BaseButton>>(
   function IconButton({ icon: Icon, iconProps, ...props }: IconButtonProps & ComponentProps<typeof BaseButton>, ref) {
     return (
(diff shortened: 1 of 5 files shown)


src/components/Button.tsx
```diff
@@ -63,7 +63,7 @@ interface IconButtonProps {
 export const IconButton = forwardRef<HTMLButtonElement, IconButtonProps & ComponentProps<typeof BaseButton>>(
   function IconButton({ icon: Icon, iconProps, ...props }: IconButtonProps & ComponentProps<typeof BaseButton>, ref) {
     return (
-      <SecondaryButton {...props} ref={ref}>
+      <SecondaryButton {...props} ref={ref} style={{ transform: 'translateY(1.75px)' }}>
```
JFrankfurt, line 66: put this in a `styled` wrapper instead of doing the inline style

nit - is it possible to accomplish this with some other tool? translate can be a little unexpected. It's better to use margin/padding/positioning in most cases. Not important to figure this out if you're busy.

JFrankfurt (approved):

PR #50, opened 2022-07-01
feat: switch to using w3r's Web3ReactProvider

kristiehuang:
Switch from using our own `ActiveWeb3Provider` to w3r's Web3ReactProvider


src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -1,74 +1,38 @@
...
-const [EMPTY_CONNECTOR, EMPTY_HOOKS] = initializeConnector<Connector>(() => EMPTY)
-const EMPTY_STATE = { connector: EMPTY_CONNECTOR, hooks: EMPTY_HOOKS }
-const EMPTY_WEB3: Web3ContextType = { connector: EMPTY }
-const EMPTY_CONTEXT = { web3: EMPTY_WEB3, updateWeb3: (updateContext: Web3ContextType) => console.log(updateContext) }
-const Web3Context = createContext(EMPTY_CONTEXT)
-
 export default function useActiveWeb3React() {
-  const { web3 } = useContext(Web3Context)
-  return web3
+  return useWeb3React()
```
kristiehuang, line 14: temporary! Will replace imports later

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -131,6 +23,15 @@ function toWeb3Connection<T extends Connector>([connector, hooks]: [T, Web3React
...
   return [connector, hooks]
 }
+function getWallet(provider: JsonRpcProvider | Eip1193Provider) {
+  if (JsonRpcProvider.isProvider(provider)) {
+    return toWeb3Connection(initializeConnector((actions) => new JsonRpcConnector(actions, provider)))
+  } else if (JsonRpcProvider.isProvider((provider as any).provider)) {
+    throw new Error('Eip1193Bridge is experimental: pass your ethers Provider directly')
+  } else {
+    return toWeb3Connection(initializeConnector((actions) => new EIP1193({ actions, provider })))
+  }
```
kristiehuang, line 33: this fxn doesn't work

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -163,38 +64,62 @@ function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: strin
...
+    () => toWeb3Connection(initializeConnector<MetaMask>((actions) => new MetaMask({ actions }))),
+    []
+  )
+  const walletConnectConnectionQR = useMemo(() => getWalletConnectConnection(false, jsonRpcEndpoint), [jsonRpcEndpoint]) // WC via tile QR code scan
+  const walletConnectConnectionPopup = useMemo(
+    () => getWalletConnectConnection(true, jsonRpcEndpoint),
+    [jsonRpcEndpoint]
+  ) // WC via built-in popup
+
+  // const networkConnection = useMemo(() => {
```
kristiehuang, line 87: and my network connection isn't working quite right either

src/components/Swap/Toolbar/index.tsx
```diff
@@ -28,8 +28,8 @@ export default memo(function Toolbar() {
   const isAmountPopulated = useIsAmountPopulated()
   const { type: wrapType } = useWrapCallback()
   const caption = useMemo(() => {
-    if (!active || !chainId) {
-      if (activating) return <Caption.Connecting />
+    if (!account || !chainId) {
```
kristiehuang, line 31: changed to use `account` instead of `active`

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -163,38 +62,63 @@ function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: strin
...
+export function ActiveWeb3Provider({
+  provider: propsProvider,
+  jsonRpcEndpoint,
+  children,
+}: PropsWithChildren<ActiveWeb3ProviderProps>) {
+  const metaMaskConnection = useMemo(
+    () => toWeb3Connection(initializeConnector<MetaMask>((actions) => new MetaMask({ actions }))),
+    []
+  )
+  const walletConnectConnectionQR = useMemo(() => getWalletConnectConnection(false, jsonRpcEndpoint), [jsonRpcEndpoint]) // WC via tile QR code scan
```
vm, line 79: why do you need different connectors for these two?

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -163,38 +62,63 @@ function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: strin
...
+  const walletConnectConnectionPopup = useMemo(
+    () => getWalletConnectConnection(true, jsonRpcEndpoint),
+    [jsonRpcEndpoint]
+  ) // WC via built-in popup
+
+  const networkConnection = useMemo(() => {
+    let network: string[] | JsonRpcProvider[]
+    if (JsonRpcProvider.isProvider(jsonRpcEndpoint)) {
```
vm, line 87: why is this necessary? if the passed in value is a string or JsonRpcProvider should be all set. if it's undefined you should skip making a network connection or error

src/hooks/connectWeb3/useActiveWeb3React.tsx
kristiehuang: being able to make our own QR code tile requires disabling the `qrcode` option during WalletConnect instantiation

Screen Shot 2022-07-01 at 4 33 02 PM
Screen Shot 2022-07-01 at 4 33 10 PM

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -163,38 +62,63 @@ function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: strin
...
+    []
+  )
+  const walletConnectConnectionQR = useMemo(() => getWalletConnectConnection(false, jsonRpcEndpoint), [jsonRpcEndpoint]) // WC via tile QR code scan
+  const walletConnectConnectionPopup = useMemo(
+    () => getWalletConnectConnection(true, jsonRpcEndpoint),
+    [jsonRpcEndpoint]
+  ) // WC via built-in popup
+
+  const networkConnection = useMemo(() => {
+    let network: string[] | JsonRpcProvider[]
```
vm, line 86: networkRpc

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -163,38 +62,63 @@ function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: strin
...
+    if (JsonRpcProvider.isProvider(jsonRpcEndpoint)) {
+      network = [jsonRpcEndpoint]
+    } else {
+      network = [jsonRpcEndpoint ?? '']
+    }
+    const urlMap = { [SupportedChainId.MAINNET]: network }
+    const connection = toWeb3Connection(initializeConnector<Network>((actions) => new Network({ actions, urlMap })))
+    connection[0].activate() // Activate network immediately after instantiation
```
vm, line 94: i think you can do this where you try to eagerly connect all the wallets, except try to use activate if eager connection not available

src/hooks/connectWeb3/useActiveWeb3React.tsx
kristiehuang: if I just do `const urlMap = { [SupportedChainId.MAINNET]: [jsonRpcEndpoint] }`, we get type errors

Screen Shot 2022-07-01 at 4 35 53 PM

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -1,128 +1,17 @@
...
-    } else {
-      ;[connector, hooks] = initializeConnector((actions) => new EIP1193({ actions, provider }))
-    }
-    connector.activate()
-    return { connector, hooks }
-  }
-  return EMPTY_STATE
-}
 
 export let connections: [Connector, Web3ReactHooks][] = []
```
vm, line 14: you should have this be a function called `getConnections` instead of having it live as mutable at the top level

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -163,38 +62,63 @@ function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: strin
...
+  if (propsProvider) {
+    connections = [
+      getWallet(propsProvider),
+      metaMaskConnection,
+      walletConnectConnectionQR,
+      walletConnectConnectionPopup,
+      networkConnection,
+    ]
+  } else {
+    connections = [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection]
```
vm, line 107: all the stuff to create the connections should live in the function `createConnections` and then just call that here. you can basically memoize the connections then also, so you don't need to memoize all the specific connectors.

vm (changes requested):
commented

src/hooks/connectWeb3/useActiveWeb3React.tsx
kristiehuang: good point on the undefined -- but if I don't handle these separately I get type errors:
Screen Shot 2022-07-01 at 4 39 26 PM

src/hooks/connectWeb3/useActiveWeb3React.tsx
vm: wow that's kinda brutal

src/hooks/connectWeb3/useActiveWeb3React.tsx
vm: you need it to be typed as `(string | JsonRpcProvider)[]`

kristiehuang: right but Network's urlMap only takes in `JsonRpcProvider | url | url[] | JsonRpcProvider[]`

So essentially I need `networkRpc` to be `string[] | JsonRpcProvider[]`, not `(string | JsonRpcProvider)[]`

vm: okay i see

src/hooks/connectWeb3/useActiveWeb3React.tsx
```diff
@@ -163,38 +62,63 @@ function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: strin
...
+    () => getWalletConnectConnection(true, jsonRpcEndpoint),
+    [jsonRpcEndpoint]
+  ) // WC via built-in popup
+
+  const networkConnection = useMemo(() => {
+    let network: string[] | JsonRpcProvider[]
+    if (JsonRpcProvider.isProvider(jsonRpcEndpoint)) {
+      network = [jsonRpcEndpoint]
+    } else {
+      network = [jsonRpcEndpoint ?? '']
```
vm, line 90: why the empty string?

src/components/ConnectWallet/index.tsx
```diff
@@ -12,12 +13,12 @@ interface WalletProps {
 export default function Wallet({ disabled, onClickConnectWallet }: WalletProps) {
   // Attempt to connect eagerly on mount
   useEffect(() => {
-    connections.forEach(([wallet, _]) => wallet?.connectEagerly?.())
+    connections.forEach(([wallet, _]) => (wallet?.connectEagerly ? wallet?.connectEagerly() : wallet?.activate()))
```
vm, line 16: can wallet be undefined?

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { WalletConnect } from '@web3-react/walletconnect'
+import { Buffer } from 'buffer'
+import { SupportedChainId } from 'constants/chains'
+import { PropsWithChildren, useMemo } from 'react'
+import { useCallback } from 'react'
+
+export let connections: [Connector, Web3ReactHooks][] = []
+export type Web3Connection = [Connector, Web3ReactHooks]
+
+function toWeb3Connection<T extends Connector>([connector, hooks]: [T, Web3ReactHooks, Web3ReactStore]): [
```
vm, line 17: this can be removed since it's the return type of `initializeConnector`

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -0,0 +1,122 @@
...
+  } else {
+    return toWeb3Connection(initializeConnector((actions) => new EIP1193({ actions, provider })))
+  }
+}
+
+function getWalletConnectConnection(useDefault: boolean, jsonRpcEndpoint?: string | JsonRpcProvider) {
+  // TODO(kristiehuang): implement RPC URL fallback, then jsonRpcEndpoint can be optional
+
+  // WalletConnect relies on Buffer, so it must be polyfilled.
+  if (!('Buffer' in window)) {
```
vm, line 39: can you polyfill this outside of this function?

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -0,0 +1,122 @@
...
+    window.Buffer = Buffer
+  }
+
+  let rpcUrl: string
+  if (JsonRpcProvider.isProvider(jsonRpcEndpoint)) {
+    rpcUrl = jsonRpcEndpoint.connection.url
+  } else {
+    rpcUrl = jsonRpcEndpoint ?? '' // TODO(kristiehuang): use fallback RPC URL
```
vm, line 47: i would handle fallback by checking `if (!jsonRpcEndpoint)`

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -0,0 +1,122 @@
...
+    []
+  )
+  const walletConnectConnectionQR = useMemo(() => getWalletConnectConnection(false, jsonRpcEndpoint), [jsonRpcEndpoint]) // WC via tile QR code scan
+  const walletConnectConnectionPopup = useMemo(
+    () => getWalletConnectConnection(true, jsonRpcEndpoint),
+    [jsonRpcEndpoint]
+  ) // WC via built-in popup
+
+  const networkConnection = useMemo(() => {
+    if (jsonRpcEndpoint) {
```
vm, line 88: yea basically move this check inside the network connector function

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -0,0 +1,122 @@
...
+  if (integratorConnection && networkConnection) {
+    connections = [
+      integratorConnection,
+      metaMaskConnection,
+      walletConnectConnectionQR,
+      walletConnectConnectionPopup,
+      networkConnection,
+    ]
+  } else if (integratorConnection) {
+    connections = [integratorConnection, metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup]
```
vm, line 105:
```
let connections = [
  integratorConnection,
  metaMaskConnection,
  walletConnectConnectionQR,
  walletConnectConnectionPopup,
  networkConnection,
]

if (integrationConnection) {
   connections = [integratorConnection, ...integrationConnection]
}

if (networkConnection) {
  // this should always be true!
  connections.push(networkConnection)
}
```

src/hooks/connectWeb3/useWeb3React.tsx
kristiehuang: where would this typically go?

vm: usually in `polyfills.ts` file which is imported in `index.tsx`

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }, [jsonRpcEndpoint])
+
+  connections = [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup]
+  if (integratorConnection) connections = [integratorConnection, ...connections]
+  if (networkConnection) connections.push(networkConnection)
+
+  return <Web3ReactProvider connectors={connections}>{children}</Web3ReactProvider>
+}
+
+export function useConnectCallback(connection: Web3Connection) {
```
vm, line 96: i think you may not need this, can probably just call activate? curious what led to this initially

vm (approved):

PR #52, opened 2022-07-05
feat: desaturate & opacity for unapproved tokens

kristiehuang:


src/components/TokenSelect/index.tsx
```diff
@@ -137,9 +142,25 @@ export default memo(function TokenSelect({ value, collapsed, disabled, onSelect
     },
     [onSelect, setOpen]
   )
+
+  const {
+    [Field.INPUT]: { currency, amount: amountIn },
+    trade,
+  } = useSwapInfo()
+  const isInputToken = value && currency?.equals(value)
+  const approvalState = useApprovalStateForSpender(amountIn, useSwapRouterAddress(trade.trade), useIsPendingApproval)
```
JFrankfurt, line 151: double checking that this works for approvals and permits

src/components/TokenImg.tsx
```diff
@@ -52,4 +52,6 @@ export default styled(TokenImg)<{ size?: number }>`
   border-radius: 100%;
   height: ${({ size }) => size || 1}em;
   width: ${({ size }) => size || 1}em;
+
+  ${({ unapproved }) => unapproved && 'filter: grayscale(100%) opacity(60%)'}
```
zzmp, line 56: use `'saturate(0) opacity(0.4)'` instead (I may have the opacity value wrong, but this is what is already in the code). Desaturation is the stated goal, so `saturate` is more semantic than `grayscale`.

src/components/TokenSelect/TokenButton.tsx
```diff
@@ -41,10 +41,11 @@ interface TokenButtonProps {
   value?: Currency
   collapsed: boolean
   disabled?: boolean
+  isUnapprovedToken?: boolean
   onClick: () => void
 }
 
-export default function TokenButton({ value, collapsed, disabled, onClick }: TokenButtonProps) {
+export default function TokenButton({ value, collapsed, disabled, isUnapprovedToken, onClick }: TokenButtonProps) {
```
zzmp, line 48: Same - use `approved: boolean` or `approval: ApprovalState`.

zzmp: Whatever you use here should be passed directly to `TokenImg`.

zzmp (changes requested):
This is already done in https://github.com/Uniswap/widgets/blob/main/src/components/Swap/Input.tsx#L38-L41.

Is this meant to replace that? Could this instead be done by copying or re-using the `Input.tsx` code for `Output.tsx`?

PR #54, opened 2022-07-05
feat: add autorouter API URL param

kristiehuang:
The Swap Widget computes the optimal trade route by running the @uniswap/smart-order-router package locally in the browser. This can take a few seconds to load & is slow.

* Allow an integrator to optionally provide a base url to their own hosted instance of the router-api. 
* If baseUrl provided, use routerAPI with fallback to clientside SOR package. Else, use clientside SOR.
* Also, we deprecate usage of useV2/V3Quoters in favor of using the clientside SOR router.

Provider needs to support eth_feeHistory


src/hooks/routing/useRoutingAPITrade.ts
```diff
@@ -0,0 +1,136 @@
...
+} {
+  const [currencyIn, currencyOut]: [Currency | undefined, Currency | undefined] = useMemo(
+    () =>
+      tradeType === TradeType.EXACT_INPUT
+        ? [amountSpecified?.currency, otherCurrency]
+        : [otherCurrency, amountSpecified?.currency],
+    [amountSpecified, otherCurrency, tradeType]
+  )
+
+  const useClientSideRouter = !Boolean(routerApiBaseUrl) // False if URL is '' or undefined
```
kristiehuang, line 38: useClientSideRouter no longer is a user setting; just check if routerApiBaseUrl is empty. Will remove param entirely once I verify with Alex about the user setting toggle when he's back from OOO

src/hooks/routing/useRoutingAPITrade.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  return useMemo(() => {
+    if (!currencyIn || !currencyOut) {
+      return {
+        state: TradeState.INVALID,
+        trade: undefined,
+      }
+    }
+
+    if (isFetching) {
```
kristiehuang, line 74: use `isFetching` instead of `isLoading && !quoteResult` to check for TradeState.LOADING

https://redux.js.org/tutorials/essentials/part-7-rtk-query-basics#using-query-hooks-in-components:~:text=isFetching%3A%20a%20boolean%20indicating%20if%20the%20hook%20is%20currently%20making%20any%20request%20to%20the%20server

src/hooks/routing/useRoutingAPITrade.ts
```diff
@@ -0,0 +1,136 @@
...
+//   putDimensions() {
+//     return
+//   }
+
+//   putMetric(key: string, value: number, unit?: MetricLoggerUnit) {
+//     sendTiming('Routing API', `${key} | ${unit}`, value, 'client')
+//   }
+// }
+
+// setGlobalMetric(new GAMetric())
```
kristiehuang, line 136: remove analytics / metrics?

src/hooks/useClientSideV3Trade.ts
```diff
@@ -23,6 +23,7 @@ const DEFAULT_GAS_QUOTE = 2_000_000
  * @param amountSpecified the exact amount to swap in/out
  * @param otherCurrency the desired output/payment currency
  */
+// TODO: delete this function & deprecate it after swapping out USDC usage for SOR
```
kristiehuang, line 26: will do this in a separate PR

src/hooks/routing/useRoutingAPITrade.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const queryArgs = useRoutingAPIArguments({
+    tokenIn: currencyIn,
+    tokenOut: currencyOut,
+    amount: amountSpecified,
+    tradeType,
+    baseUrl: routerApiBaseUrl,
+    useClientSideRouter,
+    provider: library as JsonRpcProvider,
+  })
```
kristiehuang, line 63: add baseUrl & provider params

src/hooks/swap/useRouterTrade.ts
```diff
@@ -0,0 +1,40 @@
...
+} {
+  const autoRouterSupported = useAutoRouterSupported()
+  const isWindowVisible = useIsWindowVisible()
+
+  const [debouncedAmount, debouncedOtherCurrency] = useDebounce(
+    useMemo(() => [amountSpecified, otherCurrency], [amountSpecified, otherCurrency]),
+    200
+  )
+
+  return useRoutingAPITrade(
```
kristiehuang, line 34: thinking about merging useRouterTrade (formerly useBestTrade) with useRoutingAPITrade -- no real reason for these to be separate hooks imo

src/components/Swap/index.tsx
```diff
@@ -41,6 +41,7 @@ function getTransactionFromMap(
 }
 
 export interface SwapProps extends TokenDefaults, FeeOptions {
+  routerApiBaseUrl?: string
```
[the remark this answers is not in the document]
zzmp: Alternatively, `smartOrderRouterUrl`. But I don't think it needs to include `Api` or `Base` - these are redundant with `router` and `Url`.

src/hooks/routing/useAutoRouterSupported.ts
```diff
@@ -0,0 +1,8 @@
+import useActiveWeb3React from 'hooks/useActiveWeb3React'
+
+import { AUTO_ROUTER_SUPPORTED_CHAINS } from './clientSideSmartOrderRouter'
```
zzmp, line 3: Change this to a helper function following the example in https://github.com/Uniswap/interface/pull/3915/commits/4c492e3b09b888444130e8491510f176d22c1cfe. You should do this in a separate PR, and it may be helpful to do this for all instances of `.includes` checks against lists.

src/hooks/routing/useAutoRouterSupported.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export default function useAutoRouterSupported(): boolean {
```
zzmp, line 5: This can be exported as a hook directly from `./clientSideSmartOrderRouter`, no need for a new file.

src/hooks/swap/useSwapInfo.tsx
```diff
@@ -31,7 +31,7 @@ interface SwapInfo {
 }
 
 // from the current swap inputs, compute the best trade and return it.
-function useComputeSwapInfo(): SwapInfo {
+function useComputeSwapInfo(routerApiBaseUrl?: string): SwapInfo {
```
zzmp, line 34: ->`routerUrl`

src/hooks/useClientSideV3Trade.ts
zzmp: > after swapping out USDC usage for SOR

What is this referring to?

src/state/routing/slice.ts
```diff
@@ -0,0 +1,117 @@
+import { BaseProvider, JsonRpcProvider } from '@ethersproject/providers'
+import { createApi, fetchBaseQuery, FetchBaseQueryError } from '@reduxjs/toolkit/query/react'
+import { Protocol } from '@uniswap/router-sdk'
+// eslint-disable-next-line no-restricted-imports
```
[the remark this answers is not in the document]
zzmp: I don't think this file is lazy-loaded. It's not `await import`ed anywhere. AFAIK importing `ChainId` will import the entire `@uniswap/smart-order-router` with the initial package.

However, I think if you import just the _type_, then it will be lazy-loaded, as the type is only used during build

    import type { ChainId } from '@uniswap/smart-order-router'

src/state/routing/slice.ts
```diff
@@ -0,0 +1,99 @@
...
+import { ChainId } from '@uniswap/smart-order-router'
+import ms from 'ms.macro'
+import qs from 'qs'
+
+import { GetQuoteResult } from './types'
+
+const protocols: Protocol[] = [Protocol.V2, Protocol.V3]
+
+const DEFAULT_QUERY_PARAMS = {
```
zzmp, line 12: Is the intention to eventually expose this to users?

src/state/routing/slice.ts
```diff
@@ -0,0 +1,99 @@
...
+  protocols: protocols.map((p) => p.toLowerCase()).join(','),
+  // example other params
```
zzmp, line 14: Instead of an example, can you link to the documentation/source for these query params?

zzmp (changes requested):
overall "design" looks great

src/hooks/useClientSideV3Trade.ts
kristiehuang: `useUSDCPrice` still uses useV2Quoter & useV3Quoter to calculate the price in USDC of the input currency - but now that we have a multichain SOR, we can deprecate those in favor of using SOR

src/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -25,9 +26,11 @@ async function getQuote(
     tokenOut: { address: string; chainId: number; decimals: number; symbol?: string }
     amount: BigintIsh
   },
-  routerParams: AlphaRouterParams,
+  providerUrl: string,
   routerConfig: Partial<AlphaRouterConfig>
 ): Promise<{ data: GetQuoteResult; error?: unknown }> {
+  const provider = new JsonRpcProvider(providerUrl)
```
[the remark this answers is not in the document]
kristiehuang: we unfortunately can't -- provider (type BaseProvider from ethers) is non-serializable & redux store doesn’t like that if we try to pass a non-serializable obj into queryFn

cc @justindomingue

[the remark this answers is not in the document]
zzmp: We can't use the JsonRpcProvider from our web3 connectin?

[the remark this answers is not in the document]
justindomingue: doesn't this run outside of the react lifecycle?

src/hooks/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,164 @@
...
+    tradeType,
+    isError,
+    route,
+    queryArgs,
+    gasUseEstimateUSD,
+    isSyncing,
+  ])
+}
+
+// We want to remove this from widget right?
```
justindomingue, line 151: can we remove this comment? This code shouldn't run in that file anyway

src/hooks/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,164 @@
...
+    [debouncedAmountSpecified, debouncedOtherCurrency, tradeType]
+  )
+
+  const chainId = currencyIn?.chainId as ChainId
+  if (chainId && !AUTO_ROUTER_SUPPORTED_CHAINS.includes(chainId)) {
+    throw new Error(`Router does not support this chain (chainId: ${chainId}).`)
+  }
+
+  // TODO(kristiehuang): after merging in fallback jsonRpcEndpoints, cloudflare-eth.com does not support eth_feeHistory, which we need for the router :/
+  // is there any downside to just using the (free) flashbots RPC endpoints instead? https://docs.flashbots.net/flashbots-protect/rpc/ratelimiting
```
justindomingue, line 64: confused why we're using flashbot's rpc

src/hooks/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,164 @@
...
+  const [currencyIn, currencyOut]: [Currency | undefined, Currency | undefined] = useMemo(
+    () =>
+      tradeType === TradeType.EXACT_INPUT
+        ? [debouncedAmountSpecified?.currency, debouncedOtherCurrency]
+        : [debouncedOtherCurrency, debouncedAmountSpecified?.currency],
+    [debouncedAmountSpecified, debouncedOtherCurrency, tradeType]
+  )
+
+  const chainId = currencyIn?.chainId as ChainId
+  if (chainId && !AUTO_ROUTER_SUPPORTED_CHAINS.includes(chainId)) {
```
justindomingue, line 59: should we make `useAutoRouterSupported()` take in a chainId and move this logic in there?

src/hooks/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (isFetching) {
+      return {
+        state: TradeState.LOADING,
+        trade: undefined,
+      }
+    }
+
+    let otherAmount = undefined
+    if (quoteResult) {
+      if (tradeType === TradeType.EXACT_INPUT && currencyOut) {
```
justindomingue, line 111: i think L111 to L116 would be more readable as 

```
otherAmount = CurrencyAmount.fromRawAmount(tradeType === TradeType.EXACT_INPUT ? currencyOut : currencyIn, quoteResult.quote)
```

src/hooks/routing/useRouterTrade.ts
kristiehuang: an open question - that's why I left a question here if there'd be any downside to using the flashbots rpc?

the jsonRpcEndpoint prop is optional, but there's a PR in progress to provide fallback endpoints if the integrator doesn't provide one. Rn, the fallback endpoint in that PR is 'cloudflare-eth.com' (ethereumnodes.com). However, cloudflare-eth.com doesn't support the eth_feeHistory method which we need for the router

src/hooks/routing/clientSideSmartOrderRouter.ts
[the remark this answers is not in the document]
zzmp: it does, but we don't get rid of these providers - they're just garbage-collected.
I think it will be compute-intensive (and therefore maybe laggy) if we re-instantiate them every time.

Alternatively, can we have a file-level cache where we re-use a provider until the providerUrl changes?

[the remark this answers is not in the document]
kristiehuang: redux maintainer says about re-instantiating object every time queryFn is run — “it’s cached so you wouldn’t actually do that unless you’re making a network request. which only happens when you don’t have a cached value (i.e. nothing’s already subscribed to it)”

src/hooks/useClientSideV3Trade.ts
kristiehuang: https://github.com/Uniswap/widgets/pull/58

src/hooks/routing/useRouterTrade.ts
kristiehuang: going to do this in a separate future PR - re: https://github.com/Uniswap/widgets/pull/54#discussion_r917935671

src/cosmos/Swap.fixture.tsx
```diff
@@ -57,6 +57,8 @@ function Fixture() {
   const tokenList = useOption('tokenList', { options: tokenLists, defaultValue: 'Default', nullable: false })
   console.log(tokenList)
 
+  const routerUrl = 'https://api.uniswap.org/v1/'
```
zzmp, line 60: This is a const string, so it should be ROUTER_URL.

src/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -11,6 +13,11 @@ export const AUTO_ROUTER_SUPPORTED_CHAINS: ChainId[] = Object.values(ChainId).fi
   Number.isInteger(chainId)
 )
 
+export function useAutoRouterSupported(): boolean {
+  const { chainId } = useActiveWeb3React()
+  return Boolean(chainId && AUTO_ROUTER_SUPPORTED_CHAINS.includes(chainId))
```
zzmp, line 18: This should be an exported function instead of exporting AUTO_ROUTER_SUPPORTED_CHAINS: `isAutoRouterSupportedChain(chainId?: ChainId): boolean`.

src/state/routing/slice.ts
```diff
@@ -0,0 +1,111 @@
...
+import { GetQuoteResult } from './types'
+
+const protocols: Protocol[] = [Protocol.V2, Protocol.V3]
+
+// routing API quote query params: https://github.com/Uniswap/routing-api/blob/main/lib/handlers/quote/schema/quote-schema.ts
+const DEFAULT_QUERY_PARAMS = {
+  protocols: protocols.map((p) => p.toLowerCase()).join(','),
+}
+
+const serializeRoutingCacheKey = ({ endpointName, queryArgs }: { endpointName: string; queryArgs: any }) => {
```
zzmp, line 20: Is this still necessary, with serializableCheck turned off?

kristiehuang: yes - we still need to write a new serializable fxn in order to generate a query cache key; we can now just safely silence the errors about storing the serializable provider in state

src/hooks/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,141 @@
...
+    throw new Error(`Router does not support this token's chain (chainId: ${currencyChainId}).`)
+  }
+
+  const queryArgs = useRouterArguments({
+    tokenIn: currencyIn,
+    tokenOut: currencyOut,
+    amount: debouncedAmountSpecified,
+    tradeType,
+    routerUrl,
+    provider: library as JsonRpcProvider,
```
justindomingue, line 69: should we somehow verify that `library` is in fact a `jsonrpcprovider` instead of casting?

justindomingue: and throw error if not

src/hooks/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  const route = useMemo(
+    () => computeRoutes(currencyIn, currencyOut, tradeType, quoteResult),
+    [currencyIn, currencyOut, quoteResult, tradeType]
+  )
+
+  // get USD gas cost of trade in active chains stablecoin amount
+  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(quoteResult?.gasUseEstimateUSD) ?? null
+
+  const isSyncing = currentData !== data
```
justindomingue, line 87: do you understand this bit? I feel like `syncing` may be the wrong word?

I think that currentData !== data means the query args changed.

src/hooks/routing/useRouterTrade.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  return useMemo(() => {
+    if (!currencyIn || !currencyOut) {
+      return {
```
justindomingue, line 91: might as well return `INVALID_TRADE` here?

src/state/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+  [multicall.reducerPath]: multicall.reducer,
+  [routing.reducerPath]: routing.reducer,
+})
+export const store = configureStore({
+  reducer,
+  middleware: (getDefaultMiddleware) =>
+    // in routing, we pass in a non-serializable provider object to queryFn to avoid re-instantiating on every query
+    // rtk-query stores original args in state, so we need to turn off serializableCheck
+    // this is OK because we don't use time-travel debugging nor persistance
+    getDefaultMiddleware({ thunk: true, serializableCheck: false }).concat(routing.middleware),
```
justindomingue, line 17: you should probably only disable serializableCheck for the routing slice.

justindomingue: you canuse something like

```
serializableCheck: {
 ignoredPath: ['routing'] // ?? maybe routing.reducerPath
}
```

src/state/routing/slice.ts
justindomingue: could you instead just call` defaultSerializeQueryArgs` from redux toolkit, but pass in your args w/o provider?

```
defaultSerializeQueryArgs( ({ provider, ...rest}) => ({...rest})())
```

src/hooks/routing/useRouterTrade.ts
kristiehuang: library is of type `(JsonRpcProvider & { provider?: ExternalProvider }) | Web3Provider` and Web3Provider extends JsonRpcProvider

kristiehuang: so we should be good

justindomingue (approved):

src/hooks/routing/useRouterTrade.ts
```diff
@@ -69,7 +70,7 @@ export function useRouterTrade<TTradeType extends TradeType>(
     provider: library as JsonRpcProvider,
   })
 
-  const { isFetching, isError, data, currentData } = useGetQuoteQuery(queryArgs ?? skipToken, {
+  const { isLoading, isError, data } = useGetQuoteQuery(queryArgs ?? skipToken, {
```
justindomingue, line 73: hm so one small thing, data v. currentdata

>data - The latest returned result regardless of hook arg, if present.
currentData - The latest returned result for the current hook arg, if present.

one issue you might encounter is if queryargs change, data will not be invalidated, but currentdata will

src/hooks/routing/useRouterTrade.ts
```diff
@@ -84,21 +85,25 @@ export function useRouterTrade<TTradeType extends TradeType>(
...
-
-  return useMemo(() => {
-    if (!currencyIn || !currencyOut) {
-      return {
-        state: TradeState.INVALID,
-        trade: undefined,
+  const trade = useMemo(() => {
+    if (route) {
+      try {
+        return route && transformRoutesToTrade(route, tradeType, gasUseEstimateUSD)
```
justindomingue, line 91: is the non-null check on `route` required given L89?

src/hooks/routing/useRouterTrade.ts
```diff
@@ -84,21 +85,25 @@ export function useRouterTrade<TTradeType extends TradeType>(
...
 
-  const isSyncing = currentData !== data
-
-  return useMemo(() => {
-    if (!currencyIn || !currencyOut) {
-      return {
-        state: TradeState.INVALID,
-        trade: undefined,
+  const trade = useMemo(() => {
+    if (route) {
```
justindomingue, line 89: i tend to prefer early returns to reduce indentation

```
if (!route) return
..
```

src/hooks/routing/useRouterTrade.ts
```diff
@@ -84,21 +85,25 @@ export function useRouterTrade<TTradeType extends TradeType>(
...
 
-    if (isFetching) {
-      return {
-        state: TradeState.LOADING,
-        trade: undefined,
+  return useMemo(() => {
+    if (!currencyIn || !currencyOut) return INVALID_TRADE
+
+    if (!trade && !isError) {
+      if (isDebouncing) {
```
justindomingue, line 103: i find nested if statements confusing to read, maybe rewrite as 

```
if (!trade && !isError) {
  return { state: isDebouncing ? TradeState.SYNCING : TradeState.LOADING }
}

ianlapham (approved):
Looks good now good work!

justindomingue (approved):

PR #57, opened 2022-07-12
fix: clientside router should use existing JsonRpcProvider obj

kristiehuang:
Instead of reinstantiating provider obj from providerUrl, we want to pass in our existing JsonRpcProvider obj to routingApi. While rtk-query will cache args, it doesn't destroy the provider args - meaning our javascript garbage collector will collect it, which mean it may live in memory for longer than we want. (??)

Since we don't need time-travel debugging nor persistence, we can store non-serializable objects in state.
* We turn off the serializable check for routing
* Use a custom `serializeQueryArgs` fxn to generate a query cache key

### DIFF
--- a/src/state/routing/slice.ts
+++ b/src/state/routing/slice.ts
@@ -19,7 +19,7 @@ const DEFAULT_QUERY_PARAMS = {
 }
 
 const serializeRoutingCacheKey: SerializeQueryArgs<any> = ({ endpointName, queryArgs }) => {
-  // same as default serializeQueryArgs, but we add extra case for serialization if key is provider
+  // same as default serializeQueryArgs, but we add extra case if key is our non-serializable JsonRpcProvider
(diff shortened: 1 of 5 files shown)


src/state/routing/slice.ts
```diff
@@ -15,9 +18,29 @@ const DEFAULT_QUERY_PARAMS = {
   protocols: protocols.map((p) => p.toLowerCase()).join(','),
 }
 
+const serializeRoutingCacheKey: SerializeQueryArgs<any> = ({ endpointName, queryArgs }) => {
+  // same as default serializeQueryArgs, but we add extra case if key is our non-serializable JsonRpcProvider
+  return `${endpointName}(${JSON.stringify(queryArgs, (key, value) => {
```
zzmp, line 23: I don't think you still need to serialize this, now that you've disabled serialization checks.

PR #61, opened 2022-07-20
feat: add widget disconnection via hover

kristiehuang:
Add way for user to disconnect wallet

disc1


src/components/ConnectWallet/ConnectedWalletChip.tsx
```diff
@@ -14,22 +20,44 @@ export default function ConnectedWalletChip({ disabled }: { disabled?: boolean }
...
+      }
+    })
+  }
 
   return (
     <>
       <AccountButton
         hidden={disabled}
-        // onClick={() => console.log('open account modal')}
+        onClick={() => hover && disconnectWallet()}
```
cmcewen, line 40: i'm not sure if you need hover to be positive here - i can't imagine how you would get into case where the state wouldn't update, but it seems like the safer thing is to disconnect on click always right?

cmcewen (approved):

PR #64, opened 2022-07-22
fix: disabled token button text color should use correct colors

kristiehuang:
| before | after |
| ----- | -----  | 
| Screen Shot 2022-07-22 at 11 46 43 AM  Screen Shot 2022-07-22 at 11 46 47 AM  |  Screen Shot 2022-07-22 at 11 40 16 AM Screen Shot 2022-07-22 at 11 40 10 AM |

### DIFF
--- a/src/components/TokenSelect/TokenButton.tsx
+++ b/src/components/TokenSelect/TokenButton.tsx
@@ -45,11 +45,8 @@ interface TokenButtonProps {
 }
 
 export default function TokenButton({ value, collapsed, disabled, onClick }: TokenButtonProps) {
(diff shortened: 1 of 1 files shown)


cmcewen (commented):
so just reading this (and the last PR), think we have a bit of over-memo . generally it only makes sense to do it if the calculation is expensive, or if you're concerned about referential equality between objects you return. this has some good background: https://medium.com/swlh/should-you-use-usememo-in-react-a-benchmarked-analysis-159faf6609b7

i think both `buttonBackgroundColor` and `contentColor` should be un-memo'd

cmcewen (approved):

PR #65, opened 2022-07-25
feat: add onTx status callbacks

kristiehuang:
Adding onTxSubmit/Success/Fail callbacks

| onTxSubmit | onTxSuccess | onTxFail |
| ------------ | ------------- | -------- |
| Screen Shot 2022-07-25 at 7 17 49 PM | Screen Shot 2022-07-26 at 10 59 55 AM | Screen Shot 2022-07-26 at 11 21 55 AM |

### DIFF
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -123,16 +123,19 @@ export default memo(function SwapButton({ disabled, onTxSubmit, onTxSuccess, onT
       })
       setDisplayTxHash(transaction.hash)
 
-      // Set the block containing the response to the oldest valid block to ensure that the
-      // completed trade's impact is reflected in future fetched trades.
-      transaction.wait(1).then((receipt) => {
-        if (receipt?.status === 0) {
-          onTxFail?.(new Error('Transaction failed'), receipt)
-        } else {
-          onTxSuccess?.(receipt.transactionHash, receipt)
-        }
-        setOldestValidBlock(receipt.blockNumber)
-      })
+      transaction
+        .wait(1)
+        .then((receipt) => {
+          if (receipt?.status === 0) {
+            onTxFail?.(new Error('Transaction failed'), receipt)
+          } else {
+            onTxSuccess?.(receipt.transactionHash, receipt)
+          }
+          // Set the block containing the response to the oldest valid block to ensure that the
+          // completed trade's impact is reflected in future fetched trades.
+          setOldestValidBlock(receipt.blockNumber)
+        })
+        .catch((e) => onTxFail?.(e, transaction.hash))
 
       // Only reset open after any queued animations to avoid layout thrashing, because a
       // successful swap will open the status dialog and immediately cover the summary dialog.
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -25,10 +25,11 @@ import useApprovalData, { useIsPendingApproval } from './useApprovalData'
 interface SwapButtonProps {
   disabled?: boolean
   onTxSubmit?: (txHash: string, data: any) => void
+  onTxSuccess?: (txHash: string, data: any) => void
   onTxFail?: (error: Error, data: any) => void
 }
 
-export default memo(function SwapButton({ disabled, onTxSubmit, onTxFail }: SwapButtonProps) {
+export default memo(function SwapButton({ disabled, onTxSubmit, onTxSuccess, onTxFail }: SwapButtonProps) {
   const { account, chainId } = useActiveWeb3React()
   const {
     [Field.INPUT]: {
@@ -125,6 +126,11 @@ export default memo(function SwapButton({ disabled, onTxSubmit, onTxFail }: Swap
       // Set the block containing the response to the oldest valid block to ensure that the
       // completed trade's impact is reflected in future fetched trades.
       transaction.wait(1).then((receipt) => {
+        if (receipt?.status === 0) {
+          onTxFail?.(new Error('smtn went wrong'), transaction)
+        } else {
+          onTxSuccess?.(receipt.transactionHash, receipt)
+        }
         setOldestValidBlock(receipt.blockNumber)
       })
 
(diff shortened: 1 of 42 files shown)


src/components/Swap/SwapButton/index.tsx
```diff
@@ -118,11 +123,19 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
+          if (receipt?.status === 0) {
+            onTxFail?.(new Error('Transaction failed'), receipt)
+          } else {
+            onTxSuccess?.(receipt.transactionHash, receipt)
+          }
+          // Set the block containing the response to the oldest valid block to ensure that the
+          // completed trade's impact is reflected in future fetched trades.
+          setOldestValidBlock(receipt.blockNumber)
+        })
+        .catch((e) => onTxFail?.(e, transaction.hash))
```
kristiehuang, line 138: If a transaction is cancelled by the user, it gets caught here. Does that count as a tx failing?

kristiehuang:
one thing I'm not sure about is all the different cases in which txs might fail -- too low gas, cancelled, contract execution error, etc -- and how to test them?

src/hooks/swap/useSendSwapTransaction.tsx
```diff
@@ -129,12 +132,13 @@ export default function useSendSwapTransaction(
               throw new Error(t`Transaction rejected.`)
             } else {
               // otherwise, the error was unexpected and we need to convey that
-              console.error(`Swap failed`, error, address, calldata, value)
+              console.error(`Swap failed`, error, calldata, value)
+              onTxFail?.(error, error?.transactionHash)
```
kristiehuang, line 136: this isn't necessarily an ethereum tx fail issue -- it would throw here if we haven't formed a valid tx and therefore can't submit it

should we still call `onTxFail` here?

src/components/Swap/SwapButton/index.tsx
```diff
@@ -118,11 +122,14 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
-        setOldestValidBlock(receipt.blockNumber)
-      })
+      transaction
+        .wait(1)
+        .then((receipt) => {
+          // Set the block containing the response to the oldest valid block to ensure that the
+          // completed trade's impact is reflected in future fetched trades.
+          setOldestValidBlock(receipt.blockNumber)
+        })
+        .catch((e) => onTxFail?.(e, transaction.hash)) // Do we run onTxFail on tx cancelled?
```
kristiehuang, line 132: this isn't necessarily an ethereum tx fail — the ethers library throws an error on cancelled tx, which is what we're catching here

src/hooks/transactions/index.tsx
```diff
@@ -79,14 +84,19 @@ export function TransactionsUpdater() {
   )
   const onReceipt = useCallback(
     ({ chainId, hash, receipt }) => {
+      if (receipt?.status === 0) {
+        onTxFail?.(new Error('Transaction failed'), receipt)
+      } else {
+        onTxSuccess?.(receipt.transactionHash, receipt)
+      }
```
kristiehuang, line 91: should handle ethereum onTxFail / onTxSuccess in our existing `onReceipt` callback

src/components/Swap/SwapButton/index.tsx
JFrankfurt: I don't think we should call `onTxFail` in this case. @akarys92 thoughts?

(aside-in the cancellation case we wouldn't have a `transaction.hash` here)

src/hooks/transactions/index.tsx
```diff
@@ -79,15 +84,35 @@ export function TransactionsUpdater() {
...
+    if (newTxHashes.length !== oldTxHashes.length) {
+      // if added new tx
+      newTxHashes.forEach((txHash) => {
+        if (!oldTxHashes.includes(txHash)) {
+          onTxSubmit?.(txHash, pendingTransactions[txHash])
+        }
+      })
+      oldTxs.current = pendingTransactions
+    }
+  }, [pendingTransactions, onTxSubmit])
```
kristiehuang, line 115: using `pendingTransactions` as opposed to txs from `const txs = useAtomValue(transactionsAtom)` to mirror behavior in `onReceipt`

src/hooks/swap/useSendSwapTransaction.tsx
```diff
@@ -129,9 +129,8 @@ export default function useSendSwapTransaction(
               throw new Error(t`Transaction rejected.`)
             } else {
               // otherwise, the error was unexpected and we need to convey that
-              console.error(`Swap failed`, error, address, calldata, value)
-
-              throw new Error(t`Swap failed: ${swapErrorToUserReadableMessage(error)}`)
+              console.error(`Swap failed`, error, calldata, value)
+              throw new Error(t`Swap failed: ${swapErrorToUserReadableMessage(error)}`) // FIXME: this prints to console as [object Object]
```
kristiehuang, line 133: separate small PR here to fix this
https://github.com/Uniswap/widgets/pull/66

src/hooks/transactions/index.tsx
JFrankfurt: nit: can you update the variable names `new...` and `old...` for clarity? `currentPendingTransactions`, `oldPendingTransactions`, etc.

src/hooks/transactions/index.tsx
```diff
@@ -79,15 +84,35 @@ export function TransactionsUpdater() {
...
       })
     },
-    [updateTxs]
+    [updateTxs, onTxFail, onTxSuccess]
   )
 
+  const oldTxs = useRef({})
+  useEffect(() => {
+    const newTxHashes = Object.keys(pendingTransactions)
+    const oldTxHashes = Object.keys(oldTxs.current)
```
JFrankfurt, line 105: update to: `oldTxHashes = new Set(Object.keys(oldTxs.current))` and `oldTxHashes.has(txHash)` below

`.includes` is a loop through the array, while `.has` is [strictly better](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#description).

JFrankfurt (changes requested):

JFrankfurt (approved):

PR #66, opened 2022-07-27
fix: swapErrorToUserReadableMessage should return string

kristiehuang:
swapErrorToUserReadableMessage inside `useSendSwapTransaction` was printing to console as [object Object]


src/hooks/swap/useSendSwapTransaction.tsx
```diff
@@ -83,7 +83,7 @@ export default function useSendSwapTransaction(
                   })
                   .catch((callError) => {
                     console.debug('Call threw error', call, callError)
-                    return { call, error: swapErrorToUserReadableMessage(callError) }
+                    return { call, error: <Trans>{swapErrorToUserReadableMessage(callError)}</Trans> }
```
JFrankfurt, line 86: `<Trans>` values don't work with interpolations, unfortunately.

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -1,11 +1,56 @@
...
 import { ReactNode } from 'react'
+
 /**
  * This is hacking out the revert reason from the ethers provider thrown error however it can.
  * This object seems to be undocumented by ethers.
+ 
+ * @param error an error from the ethers provider
+ * @returns user-readable error string
+ */
+export function swapErrorToUserReadableMessage(error: any): string {
```
kristiehuang, line 11: largely duplicated code, but necessary because <Trans> values don't work with interpolations

src/hooks/swap/useSendSwapTransaction.tsx
```diff
@@ -83,7 +83,7 @@ export default function useSendSwapTransaction(
                   })
                   .catch((callError) => {
                     console.debug('Call threw error', call, callError)
-                    return { call, error: swapErrorToUserReadableMessage(callError) }
+                    return { call, error: swapErrorToUserReadableTrans(callError) }
```
JFrankfurt, line 86: Is this still going to return the `Trans` object? (so `[object Object]`)

JFrankfurt (approved):

PR #67, opened 2022-07-27
feat: add a controlled swap state to the widget

JFrankfurt:
WIP


src/state/swap.ts
```diff
@@ -23,6 +23,31 @@ export const swapAtom = atomWithImmer<Swap>({
...
 
+export interface IsControlledSwapState {
+  isControlledAmount?: boolean
+  isControlledToken?: boolean
+}
+
+export const isControlledSwapStateAtom = atom<IsControlledSwapState>({
+  isControlledAmount: false,
+  isControlledToken: false,
+})
```
kristiehuang, line 34: is possible for integrator to use controlled token state & uncontrolled amount state; need to disambiguate between the two

src/cosmos/ControlledSwap.fixture.tsx
```diff
@@ -0,0 +1,164 @@
...
+      onTxFail={(error: Error, data: any) => console.log('tx fail:', error, data)}
+      // controlled token state
+      // defaultTokenSelectorDisabled={false}
+      // inputToken={inputToken}
+      // onInputTokenChange={onInputTokenChange}
+      // outputToken={outputToken}
+      // onOutputTokenChange={onOutputTokenChange}
+      defaultTokenSelectorDisabled={true}
+      inputToken={disabledInputToken}
+      outputToken={disabledOutputToken}
```
kristiehuang, line 154: sorry that this is so confusing -- 

if we're using the native token selectors, we want to use the token `onChange` functions to set swap state instead of setting internal token states directly.
i.e. onChange calls a setState, which we pass back into the value prop

else if we're disabling native token selectors, we want to directly use the input/outputToken props to set internal widget state.

kristiehuang: so, very WIP, going to need to rename these variables - but I have a 

in/outputToken, onIn/OutputTokenChange (for native defaultTokenSelectorDisabled={false})

and `disabledIn/OutputToken` (for external defaultTokenSelectorDisabled={true})

PR #68, opened 2022-07-28
feat: add onReviewSwapClick handler

kristiehuang:
Add an onReviewSwapClick handler to allow integrator to add behavior when the review swap button is clicked.
- [X] docs: https://github.com/Uniswap/docs/pull/429


JFrankfurt:
This concept seems reasonable, but I don't like that we now have two similar handlers that behave significantly differently. If this is a design route we're going to take, then I think all `on[something]` handlers should have a similar api. (`onConnectWallet` should have a callback that initiates the native wallet connection flow.)

cc @akarys92 @zzmp @ianlapham Thoughts?

JFrankfurt (approved):

PR #70, opened 2022-07-29
feat: add switch chain button prompt

kristiehuang:
If user's connected chainId does not match the chainId of the currently-selected input token, prompt user to switch networks.

- not relevant if using widget's built-in token selector (which doesn't allow selecting tokens from the wrong network in the first place)

### How to test (because our built-in token selector doesn't allow you to select tokens from wrong network):
1. temporarily force token selector dropdown to only display mainnet tokens
  i) change L74 in TokenSelect/index.tsx to: `const listHasTokens = useMemo(() => list.some((token) => token.chainId === 1), [list])`
  ii) change L34 & L46 of hooks/useTokenList/index.tsx to: `const tokenMap = chainId && chainTokenMap?.[1]`
2. connect wallet + change wallet network to non-mainnet chain
3. select any token from token dropdown
4. profit: see 'switch network' prompt

Screen Shot 2022-07-29 at 3 01 20 PM
Screen Shot 2022-07-29 at 3 01 47 PM


src/components/Swap/SwapButton/useSwitchNetwork.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const onSwitchNetwork = useCallback(async () => {
+    setIsPending(true)
+    try {
+      const desiredChainIdHex = `0x${desiredChainId?.toString(16)}`
+      await connector.provider
+        ?.request({
+          method: 'wallet_switchEthereumChain',
+          params: [{ chainId: desiredChainIdHex }],
+        })
+        .then(() => connector.activate(desiredChainId))
```
JFrankfurt, line 20: 1. How does this handle if a user does not have the chain info added to their wallet yet? 
2. Have you taken a look at the code to accomplish this on the interface? I'm familiar w/ the approach you've taken here, but it looks like there's a new way to do this w/ the current version of web3-react? 

```
export const switchChain = async (connector: Connector, chainId: SupportedChainId) => {
  if (!isChainAllowed(connector, chainId)) {
    throw new Error(`Chain ${chainId} not supported for connector (${typeof connector})`)
  } else if (connector === walletConnectConnection.connector || connector === networkConnection.connector) {
    await connector.activate(chainId)
  } else {
    const info = getChainInfo(chainId)
    const addChainParameter = {
      chainId,
      chainName: info.label,
      rpcUrls: getRpcUrls(chainId),
      nativeCurrency: info.nativeCurrency,
      blockExplorerUrls: [info.explorer],
    }
    await connector.activate(addChainParameter)
  }
}
```

@vm thoughts?

vm: you can just call `connector.activate(chainId)` now instead of doing the provider request

vm: and you'll need to pass the chainInfo! very similar to interface code.

kristiehuang: small problem I only upgraded the w3r version in the `wallet-connect-flow` branch (which hasn't been merged yet) - and it comes with a couple changes that break how we use web3react currently

kristiehuang: current version of w3r doesn't support activate(chainId)

vm: can't we merge that first and then rebase this?

src/components/Swap/SwapButton/useSwitchNetwork.tsx
```diff
@@ -0,0 +1,43 @@
+import { Trans } from '@lingui/macro'
+import { Currency } from '@uniswap/sdk-core'
+import { useWeb3React } from '@web3-react/core'
+import { Action } from 'components/ActionButton'
+import { switchChain } from 'hooks/connectWeb3/useWeb3React'
+import { Spinner } from 'icons'
+import { useCallback, useMemo, useState } from 'react'
+
+export default function useSwitchNetwork(inputCurrency?: Currency) {
```
vm, line 9: this is clever, i may steal this for interface

src/components/Swap/SwapButton/useSwitchNetwork.tsx
```diff
@@ -0,0 +1,43 @@
...
+        setIsPending(false)
+      } catch {
+        // if user cancels switch network request in-wallet
+        setIsPending(false)
+      }
+    },
+    [connector]
+  )
+
+  const switchNetworkAction = useMemo((): Action | undefined => {
```
vm, line 27: this can just be a return without naming it

vm: just return `[switchChain, pending]` so that the UI state is not coupled

src/components/Swap/SwapButton/useSwitchNetwork.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const [isPending, setIsPending] = useState(false)
+  const onSwitchNetwork = useCallback(
+    async (desiredChainId: number) => {
+      setIsPending(true)
+      try {
+        await switchChain(connector, desiredChainId)
+        setIsPending(false)
+      } catch {
+        // if user cancels switch network request in-wallet
```
vm, line 20: that's not the only thing it's handling necessarily, so i'd probably remove the comment

src/components/Swap/SwapButton/useSwitchNetwork.tsx
```diff
@@ -0,0 +1,43 @@
...
+    },
+    [connector]
+  )
+
+  const switchNetworkAction = useMemo((): Action | undefined => {
+    if (!inputCurrency) return undefined
+    const desiredChainId = inputCurrency.chainId
+    if (chainId && chainId !== desiredChainId) {
+      return isPending
+        ? { message: <Trans>Switch network in your wallet</Trans>, icon: Spinner }
```
vm, line 32: move the copy and icon stuff into the UI component

src/components/Swap/SwapButton/index.tsx
```diff
@@ -49,6 +50,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
   const deadline = useTransactionDeadline()
 
   const { type: wrapType, callback: wrapCallback } = useWrapCallback()
+  const switchNetworkAction = useSwitchNetwork(inputCurrency)
```
vm, line 53: technically chain !== network (is my understanding). since in web3-react, network refers to infura

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -72,11 +118,12 @@ interface ActiveWeb3ProviderProps {
 
 export function ActiveWeb3Provider({
   provider,
-  jsonRpcUrlMap,
+  jsonRpcUrlMap: propsJsonRpcUrlMap,
```
vm, line 121: i would leave this name as is

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -72,11 +118,12 @@ interface ActiveWeb3ProviderProps {
...
   defaultChainId: propsDefaultChainId,
   children,
 }: PropsWithChildren<ActiveWeb3ProviderProps>) {
   const onError = console.error
+  jsonRpcUrlMap = propsJsonRpcUrlMap
```
vm, line 126: i think this doesn't copy it, it just points a reference. you'd need to do a deep copy

vm (changes requested):
some comments

src/components/Swap/SwapButton/useSwitchNetwork.tsx
```diff
@@ -0,0 +1,43 @@
+import { Trans } from '@lingui/macro'
```
vm, line 1: this should live in `web3connection/hooks`

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -64,6 +65,51 @@ function getConnectionFromWalletConnect(
...
+      return ['https://polygon-rpc.com/']
+    case SupportedChainId.POLYGON_MUMBAI:
+      return ['https://rpc-endpoints.superfluid.dev/mumbai']
+    default:
+  }
+  // Our API-keyed URLs will fail security checks when used with external wallets.
+  throw new Error('RPC URLs must use public endpoints')
+}
+
+export const switchChain = async (connector: Connector, chainId: SupportedChainId) => {
```
vm, line 95: i would move this function into the hook file you have

src/components/Swap/SwapButton/index.tsx
kristiehuang: o did not know this

src/hooks/connectWeb3/useWeb3React.tsx
kristiehuang: can we not just point a reference?

vm: if it's just a reference then why rename the prop?

kristiehuang: not sure if this is the right way to do it but i want to use jsonRpcUrlMap in a different file, so need to export it. 

creating a `export let jsonRpcUrlMap` & setting its value to the passed-in prop here
so need to disambiguate between the exported var and the prop

src/hooks/connectWeb3/useSwitchChain.tsx
```diff
@@ -0,0 +1,75 @@
...
+      return ['https://polygon-rpc.com/']
+    case SupportedChainId.POLYGON_MUMBAI:
+      return ['https://rpc-endpoints.superfluid.dev/mumbai']
+    default:
+  }
+  // Our API-keyed URLs will fail security checks when used with external wallets.
+  throw new Error('RPC URLs must use public endpoints')
+}
+
+export const switchChain = async (connector: Connector, chainId: SupportedChainId) => {
```
vm, line 39: don't need to export

vm: also would make this a function for consistency

src/hooks/connectWeb3/useWeb3React.tsx
vm: yea wanna avoid this... shouldn't need to update jsonRpcUrlMap like that. if you do need to do it we should be using an atom

PR #123, opened 2022-08-03
Wcf/refactor wallet connection header

kristiehuang:


src/index.tsx
```diff
@@ -14,11 +15,12 @@ export type { DefaultAddress, TokenDefaults } from 'hooks/swap/useSyncTokenDefau
 export type { Theme } from 'theme'
 export { darkTheme, defaultTheme, lightTheme } from 'theme'
 
-export type SwapWidgetProps = SwapProps & WidgetProps
+export type SwapWidgetProps = HeaderProps & SwapProps & WidgetProps
 
 export function SwapWidget(props: SwapWidgetProps) {
   return (
     <Widget {...props}>
+      <Header title="Swap" {...props} />
```
JFrankfurt, line 23: If the Header component isn't used anywhere else at the moment I think we should drop the `title` prop and just put `"Swap"` directly into the dom.

JFrankfurt (approved):

src/components/Header.tsx
```diff
@@ -13,15 +20,33 @@ const HeaderRow = styled(Row)`
...
+    }
+  }, [props.onConnectWalletClick, onConnectWalletClick, setOnConnectWalletClick])
+
   return (
     <HeaderRow iconSize={1.2}>
-      <Row gap={0.5}>{title && <ThemedText.Subhead1>{title}</ThemedText.Subhead1>}</Row>
-      <Row gap={1}>{children}</Row>
+      <Row gap={0.5}>
+        <ThemedText.Subhead1>
+          <Trans>Swap</Trans>
```
zzmp, line 43: Should widgets expand past `Swap`, this will need to be configurable.
Can you keep passing `title` into the header?

src/index.tsx
```diff
@@ -14,11 +15,12 @@ export type { DefaultAddress, TokenDefaults } from 'hooks/swap/useSyncTokenDefau
 export type { Theme } from 'theme'
 export { darkTheme, defaultTheme, lightTheme } from 'theme'
 
-export type SwapWidgetProps = SwapProps & WidgetProps
+export type SwapWidgetProps = HeaderProps & SwapProps & WidgetProps
```
zzmp, line 18: This should be an interface extending the existing props:

```
export interface SwapWidgetProps extends HeaderProps, SwapProps, WidgetProps {}
```

This is slightly easier on the type system: it reduces time for type checking and it surfaces better types downstream.

src/components/Header.tsx
```diff
@@ -13,15 +20,33 @@ const HeaderRow = styled(Row)`
...
+  onConnectWalletClick?: () => void | Promise<boolean>
 }
 
-export default function Header({ title, children }: HeaderProps) {
+export default function Header(props: HeaderProps) {
+  const { isActive } = useWeb3React()
+  const onSupportedNetwork = useOnSupportedNetwork()
+  const isDisabled = !(isActive && onSupportedNetwork)
+
+  const [onConnectWalletClick, setOnConnectWalletClick] = useAtom(onConnectWalletClickAtom)
```
zzmp, line 32: You should be able to set this without referencing the already-set value:

```
const setOnConnectWalletClick = useSetAtom(onConnectWalletClickAtom)
useEffect(() => setOnConnectWalletClick(props.onConnectWalletClick), [props.onConnectWalletClick, setOnConnectWalletClick])
```

src/components/Header.tsx
```diff
@@ -13,15 +20,33 @@ const HeaderRow = styled(Row)`
...
-      <Row gap={0.5}>{title && <ThemedText.Subhead1>{title}</ThemedText.Subhead1>}</Row>
-      <Row gap={1}>{children}</Row>
+      <Row gap={0.5}>
+        <ThemedText.Subhead1>
+          <Trans>Swap</Trans>
+        </ThemedText.Subhead1>
+      </Row>
+      <Row gap={1}>
+        <Wallet disabled={props.hideConnectionUI} />
+        <Settings disabled={isDisabled} />
```
zzmp, line 48: Settings is specific to the Swap widget (similar to the title) - this either needs to be a prop, or a callback to activate the Swap's settings.

zzmp (changes requested):
This change would couple src/components/Header to src/components/Swap. Header needs to be updated so that it does not couple Swap, in case we want to use it for other widgets in the future.

PR #124, opened 2022-08-04
build: downgrade node in .nvmrc

vm:
Hardhat requires a maximum of v16, which is in our dependencies.


kristiehuang (changes requested):
also need to downgrade @types/node version

JFrankfurt:
lame!!

current version of hardhat supports 18

PR #127, opened 2022-08-05
build: strip styled-components dep and DefaultTheme from the bundle

JFrankfurt:
- Labels styled-components as a peerDependency, so that [it is not bundled, as directed by the docs](https://styled-components.com/docs/faqs#i-am-a-library-author-should-i-bundle-styledcomponents-with-my-library)
- Strips the `DefaultTheme` declaration merge from the bundle, so that it does not conflict with downstream usage


.releaserc.json
```diff
@@ -1,15 +1,15 @@
 {
```
JFrankfurt, line 1: ran prettier on `.` and it picked this up. Leaving it in.

package.json
```diff
@@ -105,7 +104,8 @@
     "react": ">=17.0.1",
     "react-dom": ">=17.0.1",
     "react-redux": ">=7.2.2",
-    "redux": "^4.1.2"
+    "redux": "^4.1.2",
+    "styled-components": ">=5"
```
vm, line 108: why >= instead of ^?

JFrankfurt: following the example of the styled-components docs: 
> Also, note that in the peerDependencies section the version string has been made a more permissive >= n. This allows future versions of styled-components to work automatically and you can simply narrow the range with a patch update to your library if a breaking change is eventually added.

zzmp: `^` restricts to that major version, while `>=` allows higher major versions.

If a library does not use breaking changes, this is desirable, to allow the consumer to upgrade the dependency and still use (in this case) the widget. This is done with React a lot (because the breaking changes are usually more transparent to consumers).

rollup.config.js
```diff
@@ -37,7 +37,7 @@ const transpile = {
   input: 'src/index.tsx',
   external: (source) => {
     // @ethersproject/* modules are provided by ethers
-    return source.startsWith('@ethersproject/')
+    return [...source.startsWith('@ethersproject/'), 'styled-components']
```
zzmp, line 40: This change shouldn't be necessary, as this should already be marked external because it is not a dependency.

zzmp (changes requested):

rollup.config.js
JFrankfurt: That makes sense, but I'm confused now why the styled-components docs for library maintainers specifically call this out as a necessary step *after* the dev/peer changes. 🤔

zzmp: In our case, this is done through https://www.npmjs.com/package/rollup-plugin-node-externals in the rollup config, similar to the suggested `rollup-plugin-peer-deps-external` solution in the styled-component docs.

vm (dismissed):

zzmp (approved):

PR #129, opened 2022-08-08
feat(onTokenSelectorClick): add onTokenSelectorClick event handler

JFrankfurt:
docs: https://github.com/Uniswap/docs/pull/457


src/components/Swap/index.tsx
```diff
@@ -51,26 +51,14 @@ export interface SwapProps extends TokenDefaults, FeeOptions {
...
-    }
-  }, [props.onReviewSwapClick, onReviewSwapClick, setOnReviewSwapClick])
-
-  const [onConnectWalletClick, setOnConnectWalletClick] = useAtom(onConnectWalletClickAtom)
-  useEffect(() => {
-    if (props.onConnectWalletClick !== onConnectWalletClick) {
-      setOnConnectWalletClick((old: (() => void | Promise<boolean>) | undefined) => (old = props.onConnectWalletClick))
-    }
-  }, [props.onConnectWalletClick, onConnectWalletClick, setOnConnectWalletClick])
+  useIntegratorEventHandlers(props)
```
JFrankfurt, line 61: I moved all the removed code as-is into `useIntegratorEventHandlers`

src/components/TokenSelect/index.tsx
```diff
@@ -129,7 +131,20 @@ export default memo(function TokenSelect({ value, collapsed, disabled, onSelect
...
+      return promise
+        .then((open) => {
+          setOpen(open)
+        })
+        .catch(() => {
+          setOpen(false)
+        })
+    }
+    return setOpen(true)
+  }, [])
```
JFrankfurt, line 147: call the handler if it exists

if we get a promise from the handler, then use its resolved value to set the token selector open state

if the promise rejects, then don't open the token selector

otherwise set the open state to true

src/utils/animations.ts
```diff
@@ -1,7 +1,8 @@
 import { RefObject } from 'react'
 
 export function isAnimating(node: Animatable | Document) {
-  return (node.getAnimations().length ?? 0) > 0
+  // ?. used here to support tests that call this function
+  return (node?.getAnimations?.().length ?? 0) > 0
```
JFrankfurt, line 5: see comment

test/test-utils.tsx
```diff
@@ -0,0 +1,69 @@
+import { render, RenderOptions } from '@testing-library/react'
```
JFrankfurt, line 1: this is a utility for the e2e tests to render all our providers

zzmp (changes requested):

test/test-utils.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const customRender = (ui: ReactElement, options?: CustomOptions) => {
+  if (options?.atomProviderInitialValues) {
+    atomProviderInitialValues = options.atomProviderInitialValues
+  }
+  return render(ui, { wrapper: SwapWidgetProviders, ...options })
+}
+
+export * from '@testing-library/react'
```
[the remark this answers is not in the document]
JFrankfurt: This is actually a recommended pattern! I thought it was weird too. https://testing-library.com/docs/react-testing-library/setup/#custom-render

src/utils/animations.ts
zzmp: Which tests pass undefined, here?
This should be codified in the parameter, with the comment there:

    // Some tests will pass undefined for the Document.
    export function isAnimating(node?: Animatable | Document) {

src/components/TokenSelect/index.tsx
```diff
@@ -77,14 +77,14 @@ export function TokenSelectDialog({ value, onSelect, onClose }: TokenSelectDialo
 
   if (!listHasTokens && isLoaded) {
     return (
-      <Dialog color="module" onClose={onClose}>
+      <Dialog color="module" onClose={onClose} data-testid="token-select-dialog">
```
JFrankfurt, line 80: @zzmp I went with this pattern initially, since I'm more familiar with it, but the convention here is to query by user-accessible information first: https://testing-library.com/docs/queries/about#priority

src/components/TokenSelect/index.tsx
```diff
@@ -129,7 +131,20 @@ export default memo(function TokenSelect({ value, collapsed, disabled, onSelect
   usePrefetchBalances()
 
   const [open, setOpen] = useState(false)
-  const onOpen = useCallback(() => setOpen(true), [])
+  const onTokenSelectorClick = useAtomValue(onTokenSelectorClickAtom)
+  const onOpen = useCallback(() => {
+    const promise = onTokenSelectorClick?.()
+    if (promise) {
+      return promise
```
zzmp, line 138: nit: if you're one-lining L146, then you can one-line this, too:

    return promise.then((open) => setOpen(open)).catch(() => setOpen(false))

src/components/Widget.tsx
```diff
@@ -105,7 +106,15 @@ export type WidgetProps = {
...
-  const { children, theme, provider, dialog: userDialog, className, onError } = props
+  return <TestableWidget {...props} initialAtomValues={undefined} />
+}
+
+export interface TestableWidgetProps extends WidgetProps {
+  initialAtomValues?: Iterable<readonly [Atom<unknown>, unknown]>
+}
+
+export function TestableWidget(props: PropsWithChildren<TestableWidgetProps>) {
+  assert(props.initialAtomValues && process.env.NODE_ENV === 'test', 'initialAtomValues may only be used for testing')
```
zzmp, line 117: (This is my mistake, but I don't want to push too many commits to your branch in case you want to back it out.)

This fails if unset of if not in `'test'`. Instead, it should be:

    assert(
      props.initialAtomValues && process.env.NODE_ENV === 'test' ?? true,
      'initialAtomValues may only be used for testing'
    )

src/components/Swap/index.tsx
zzmp, line 61: I'd prefer if this followed the existing naming convention: `useSyncEventHandlers`.
For the most part, I don't think we should refer to the "integrator" in code.

src/hooks/useIntegratorEventHandlers.ts
```diff
@@ -0,0 +1,37 @@
...
+import { onConnectWalletClickAtom } from 'state/wallet'
+
+interface UseIntegratorEventHandlersArgs {
+  onConnectWalletClick?: () => void | Promise<boolean>
+  onReviewSwapClick?: () => void | Promise<boolean>
+  onTokenSelectorClick?: () => void | Promise<boolean>
+}
+
+export default function useIntegratorEventHandlers(handlers: UseIntegratorEventHandlersArgs): void {
+  const [onReviewSwapClick, setOnReviewSwapClick] = useAtom(onReviewSwapClickAtom)
```
zzmp, line 13: I think it's overkill to check against the old value of `onReviewSwapClick`: it adds a dependency, and if this behaves like useState, then setting the current value should be a nop anyways.

Can all of these handlers be simplified to

    const { onReviewSwapClick } = props
    const setOnReviewSwapClick = useSetAtom(onReviewSwapClickAtom)
    useEffect(() => setOnReviewSwapClick(onReviewSwapClick), [onReviewSwapClick])

zzmp (changes requested):

src/components/TokenSelect/index.tsx
zzmp: that makes sense, but it seems error-prone to query for the same string as the button text, and it makes it confusing to read. I'd prefer testid because of that, but could see the merits of either.

test/test-utils.tsx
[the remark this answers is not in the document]
zzmp: looks like I was already using this pattern, too, in src/tests. 🤷

src/hooks/useIntegratorEventHandlers.ts
JFrankfurt: @kristiehuang

kristiehuang (changes requested):
should pass in field input to handler so integrator knows whether or not input/output token selector was clicked

src/components/TokenSelect/index.test.tsx
```diff
@@ -35,11 +39,12 @@ describe('TokenSelect.tsx', () => {
         }
       )
 
-      await user.click(component.getByTestId('token-select'))
-      await expect(component.findByTestId('token-select-dialog')).resolves.toBeTruthy()
+      await user.click(component.getByRole('button'))
+      await waitFor(() => expect(component.getAllByText('Select a token').length).toBe(2))
```
zzmp, line 43: Can you find the header by role and then assert that the text content is 'Select a token', so it is less confusing (by just doubling the string)?

zzmp: This should not use `waitFor` - the change should be synchronous, and doing so prolongs the test duration.

src/components/TokenSelect/index.test.tsx
```diff
@@ -49,11 +54,12 @@ describe('TokenSelect.tsx', () => {
         }
       )
 
-      await user.click(component.getByTestId('token-select'))
-      await expect(component.findByTestId('token-select-dialog')).rejects.toThrow()
+      await user.click(component.getByRole('button'))
+      await waitFor(() => expect(component.getAllByText('Select a token').length).toBe(1))
```
zzmp, line 58: This should not use `waitFor` - the change should be synchronous.

src/components/TokenSelect/index.test.tsx
```diff
@@ -63,8 +69,8 @@ describe('TokenSelect.tsx', () => {
...
```
zzmp, line 73: This should not use `waitFor` - the change should be synchronous.

src/test/index.tsx
```diff
@@ -12,7 +12,6 @@ import { ThemeProvider } from 'theme'
 
 export type { RenderResult } from '@testing-library/react'
 export { act, waitFor } from '@testing-library/react'
-export { default as user } from '@testing-library/user-event'
```
zzmp: You could still `export { default as userEvent }` here, and then import from `.../test` and call `userEvent.setup()`. It may simplify imports in tests.

zzmp (changes requested):
Just some testing hygiene - the implementation code lgtm.

zzmp (approved):
lgtm - just pending @kristiehuang's requested changes

src/hooks/useIntegratorEventHandlers.ts
kristiehuang: when I tried this without the check I'd get 'Too many re-renderers' errors

kristiehuang: hm - I'd get 'too many re-renders' errors without the check but can't reproduce it now. might have fixed something in the meantime, seems fair to me to simplify if it works

kristiehuang (approved):

kristiehuang:
missing lint in TokenSelect/index.tsx but otherwise lgtm

PR #130, opened 2022-08-08
fix: make header text clickable

kristiehuang:
Both left arrow & header text are now clickable in a dialog popup (as per design request)

Screen Shot 2022-08-08 at 1 30 05 PM

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -67,11 +67,12 @@ interface HeaderProps {
 }
 
 export function Header({ title, children, ruled }: HeaderProps) {
+  const onClose = useContext(OnCloseContext)
   return (
     <>
       <Column>
         <HeaderRow iconSize={1.2}>
-          <TextButton color="primary" onClick={useContext(OnCloseContext)}>
+          <TextButton color="primary" onClick={onClose}>
             <Row justify="flex-start" gap={0.5}>
               <ChevronLeft />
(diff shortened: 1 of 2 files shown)


src/cosmos/Swap.fixture.tsx
```diff
@@ -91,9 +91,7 @@ function Fixture() {
           resolve(true) // to open our built-in wallet connect flow
         })
       }
-      onReviewSwapClick={() => {
-        return new Promise((resolve) => resolve(true))
-      }}
+      onReviewSwapClick={() => new Promise((resolve) => resolve(true))}
```
kristiehuang, line 94: unrelated but nit to remove extra brackets

src/components/Dialog.tsx
```diff
@@ -70,10 +71,12 @@ export function Header({ title, children, ruled }: HeaderProps) {
     <>
       <Column>
         <HeaderRow iconSize={1.2}>
-          <Row justify="flex-start" gap={0.5}>
-            <IconButton color="primary" onClick={useContext(OnCloseContext)} icon={ChevronLeft} />
-            <Row gap={0.5}>{title && <ThemedText.Subhead1>{title}</ThemedText.Subhead1>}</Row>
-          </Row>
+          <TextButton color="primary" onClick={useContext(OnCloseContext)}>
```
JFrankfurt, line 74: nit: (I know this isn't a part of this PR) - hooks shouldn't be used in the render function

JFrankfurt (approved):

PR #134, opened 2022-08-10
feat: compile ajv validators at build instead of runtime

### LINKED ISSUE #51: Unable to use Widget without unsafe-eval CSP
**Bug Description**
Uniswap's dependency AJV attempts to inject a script inline requiring `unsafe-eval`.

**Steps to Reproduce**
(Example, Next JS)

1. Disabling `unsafe-eval` or adding specific Content Security Policies for `script-src`
2. App will throw an error

image

**Expected Behavior**
The ability to use the widget in a more secure environment.

**Additional Context**
[AJV's documentation](https://ajv.js.org/security.html#content-security-policy) recommends compiling using CLI or including it in the build.

cmcewen:
fixes #51


scripts/compileAjvValidators.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable */
```
zzmp, line 1: To avoid a new pattern, can this be moved to a top-level file called `ajv.config.js`?

scripts/compileAjvValidators.js
```diff
@@ -0,0 +1,20 @@
...
+
+const fs = require('fs')
+const path = require('path')
+const Ajv = require('ajv')
+const standaloneCode = require('ajv/dist/standalone').default
+const addFormats = require('ajv-formats')
+const schema = require('@uniswap/token-lists/dist/tokenlist.schema.json')
+
+const ajv = new Ajv({ code: { source: true, esm: true } })
```
zzmp, line 10: Why not write both schemas to one file (see [docs](https://ajv.js.org/standalone.html#generating-functions-s-for-multiple-schemas-using-the-js-library-es6-and-esm-exports))

src/hooks/useTokenList/validateTokenList.ts
```diff
@@ -1,40 +1,34 @@
...
-    validate?.errors?.map((error) => [error.dataPath, error.message].filter(Boolean).join(' ')).join('; ') ??
+    validate?.errors?.map((error) => [error.instancePath, error.message].filter(Boolean).join(' ')).join('; ') ??
     'unknown error'
   )
 }
 
+async function loadValidator(schema: ValidationSchema): Promise<ValidateFunction> {
+  const [, validatorModule] = await Promise.all([
+    import('ajv'),
+    schema === ValidationSchema.LIST
```
zzmp, line 19: This should use a switch-case (in a new variable) to ensure completeness, should new schemas ever be added.

src/hooks/useTokenList/validateTokenList.ts
```diff
@@ -1,40 +1,34 @@
...
+  return (await validatorModule.default) as ValidateFunction
+}
+
 /**
  * Validates an array of tokens.
  * @param json the TokenInfo[] to validate
  */
 export async function validateTokens(json: TokenInfo[]): Promise<TokenInfo[]> {
-  const validate = (await validator).getSchema(ValidationSchema.TOKENS)
+  const validate = await loadValidator(ValidationSchema.TOKENS)
```
zzmp, line 31: Would you consider a single `validate(schema: ValidationSchema, data: unknown)` function?

.gitignore
```diff
@@ -34,4 +34,8 @@ yarn-error.log*
 notes.txt
 
 # tests
-/coverage
\ No newline at end of file
+/coverage
+
+# generated ajv schema
+/src/__generated__/*
```
zzmp, line 40: Unless you feel strongly about it, I'd rather this be put in something like `/src/ajv` or `/src/schema` or `/src/validators`, that better describes what it is. This is what is done with `/src/types/v3` and `/src/abis/types` (at the top of the file). I also don't want a separate pattern here for file generation.

zzmp (changes requested):

scripts/compileAjvValidators.js
cmcewen: i was fighting with it because the schemas have the same name, so it didn't like using the same instance. imo not worth the additional time to make it work

scripts/compileAjvValidators.js
cmcewen: so that eslint won't check it? idk i think that would be a little bit of misnaming because it's not a config file

src/hooks/useTokenList/validateTokenList.ts
cmcewen: i changed it, but i can't say i think it's much better - you lose type specificity and it doesn't save much in terms of LOC

.gitignore
cmcewen: i do feel like it is useful to know which code is generated vs authored. maybe a "lib" folder is better for that? the more practical reason is for lint/type ignores imo, but also i think you don't necessarily want to read generated code

cmcewen: i'd be willing to move all of that code to a __generated__ file path or lib if you agree

JFrankfurt (approved):

.gitignore
zzmp: Would you be willing to put this into `/src/validators`, and then follow-up by moving _all_ generated code to a `__generated` directory?

scripts/compileAjvValidators.js
zzmp: not for eslint, just to avoid more complexity in the directory tree by adding a new `scripts` folder. What about just `ajv.js`?

zzmp (approved):
Could be cleaned up by making a helper function instead of `ajv` and `ajv2`, and changing the file structure, but this is definitely good to merge and be cleaned up later.

PR #140, opened 2022-08-12
fix: downstream referential equality errors

NoahZinsmeister:


src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -70,30 +70,32 @@ interface ActiveWeb3ProviderProps {
   defaultChainId: SupportedChainId
 }
 
+const onError = console.error
```
NoahZinsmeister, line 73: if for whatever reason we're worried about `console.error` not being defined at file load time we could move this into a `useRef` in the function component body, but don't think that should be done unless we have an explicit reason to

zzmp: This prevents developers from monkey-patching console.error. I think it's preferable to invoke it lazily, and not cache it on file load.

What's wrong with `const onError = console.error` in the `ActiveWeb3Provider` fn body?

NoahZinsmeister: my worry was that console.error might be changing identity every render, which would have caused a lot of the memos to fire again without changing the key

your point about monkey-patching is good though, and also my second commit should fix the underlying issue even if console.error is changing all the time, so i'll revert this bit

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -77,25 +77,24 @@ export function ActiveWeb3Provider({
...
   useEffect(() => {
     if (propsDefaultChainId !== defaultChainId) setDefaultChainId(propsDefaultChainId)
   }, [propsDefaultChainId, defaultChainId, setDefaultChainId])
 
-  const integratorConnection = useMemo(() => getConnectionFromProvider(onError, provider), [onError, provider])
-  const metaMaskConnection = useMemo(
-    () => toWeb3Connection(initializeConnector<MetaMask>((actions) => new MetaMask({ actions, onError }))),
-    [onError]
+  // we're abusing useState a bit to ensure that `metaMaskConnection` is only ever computed once, i.e. is referentially static
+  const [metaMaskConnection] = useState(() =>
```
zzmp, line 87: This would more appropriately be `useMemo` - can we use `useMemo` and pass in `onError: console.error`?

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -104,12 +103,28 @@ export function ActiveWeb3Provider({
     [jsonRpcUrlMap, defaultChainId]
   )
 
+  const integratorConnection = useMemo(() => getConnectionFromProvider(onError, provider), [provider])
+
   connections = [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection]
```
zzmp, line 108: Instead of a `useRef`, can this be memo-ized? If we memoize this, then we can also have a counter for the key with `useEffect(() => key.current += 1, [connections])`.

zzmp: I feel like everything before this line can be reverted, and the key just regenerated in this way

    const connections = useMemo(() =>
      [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection], 
      [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection])
    
    // Re-render the Web3ReactProvider whenever connections change, so that the passed connections is always referentially stable. This is forced using a key.
    const key = useRef(0)
    useEffect(() => key.current += 1, [connections])
    
    return (
      <Web3ReactProvider connectors={connections} key={key.current}>
        {children}
      </Web3ReactProvider>
    )

zzmp, line 108: I feel like everything before this line can be reverted, and the key just regenerated in this way

    const connections = useMemo(() =>
      [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection], 
      [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection])
    
    // Re-render the Web3ReactProvider whenever connections change, so that the passed connections is always referentially stable. This is forced using a key.
    const key = useRef(0)
    useEffect(() => key.current += 1, [connections])
    
    return (
      <Web3ReactProvider connectors={connections} key={key.current}>
        {children}
      </Web3ReactProvider>
    )

zzmp (changes requested):

NoahZinsmeister:
closing in favor of https://github.com/Uniswap/widgets/pull/144

src/hooks/connectWeb3/useWeb3React.tsx
NoahZinsmeister: yeah i think you're right, see other PR

PR #144, opened 2022-08-12
fix: downstream referential equality errors

NoahZinsmeister:


src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -104,12 +104,22 @@ export function ActiveWeb3Provider({
     [jsonRpcUrlMap, defaultChainId]
   )
 
-  connections = [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection]
-  if (integratorConnection) connections = [integratorConnection, ...connections]
+  const key = useRef(0)
+  connections = useMemo(() => {
+    key.current += 1
```
NoahZinsmeister, line 109: this feels a little funny, i'd be happy to put this in a useEffect as you suggested @zzmp, but incrementing the key every time this memo function runs should accomplish what we're going for, and it felt more tightly couple to the memoization

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -104,12 +104,22 @@ export function ActiveWeb3Provider({
...
+    let connections = [metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection]
```
zzmp, line 110: I think this is more cleanly expressed as

    return [integratorConnection, metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection].filter(Boolean)

zzmp (approved):

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -7,7 +7,7 @@ import { Connector, Provider as Eip1193Provider, Web3ReactStore } from '@web3-re
 import { WalletConnect } from '@web3-react/walletconnect'
 import { SupportedChainId } from 'constants/chains'
 import { atom, useAtom } from 'jotai'
-import { PropsWithChildren, useEffect, useMemo } from 'react'
+import { PropsWithChildren, useEffect, useMemo, useRef } from 'react'
 import JsonRpcConnector from 'utils/JsonRpcConnector'
 
 export type Web3Connection = [Connector, Web3ReactHooks]
```
zzmp, line 13: `connections` is potentially invalid if it changes over time. Does this export need to be reconsidered?

src/hooks/connectWeb3/useWeb3React.tsx
zzmp: TBH I think this is most cleanly expressed by splitting the memo and the hook, because that's the most "react"-ey and best expresses the intentions:

    const connections = useMemo(() =>
      [integratorConnection, metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection].filter(Boolean),
      [integratorConnection, metaMaskConnection, walletConnectConnectionQR, walletConnectConnectionPopup, networkConnection])
    useEffect(() => key.current += 1, [connections])

zzmp (changes requested):

src/hooks/connectWeb3/useWeb3React.tsx
zzmp: also I'm very on the fence of splitting the useMemo into useMemo/useEffect, so I think you should probably ignore my last comment and leave them coupled. coupling here feels good.

NoahZinsmeister: going to leave in the useMemo for now, added an explanatory comment!

src/hooks/connectWeb3/useWeb3React.tsx
NoahZinsmeister: so, while i'm still not thrilled about this export, i checked and all the other components that use `connections` are children of `Web3ReactProvider`, and since the whole point of this pr is ensuring that this component re-renders whenever `connections` changes, there shouldn't be any immediate issues

my vote would be to merge this and consider a refactor in a later pr? can make an issue if there's agreement

JFrankfurt (approved):

src/hooks/connectWeb3/useWeb3React.tsx
```diff
@@ -104,12 +104,30 @@ export function ActiveWeb3Provider({
...
+    // so tightly coupling the key update with the memo update shouldn't cause any issues,
+    // and most clearly expresses the intent
+    key.current += 1
+    return [
+      integratorConnection,
+      metaMaskConnection,
+      walletConnectConnectionQR,
+      walletConnectConnectionPopup,
+      networkConnection,
+    ].filter((connection): connection is Web3Connection => Boolean(connection))
```
zzmp, line 120: Nit: I think you can add the type to the template parameters and just use Boolean, so that the actual code remains simpler.

src/hooks/connectWeb3/useWeb3React.tsx
zzmp: If this is the case, could there just be a hook to expose connections within the React lifecycle? I'm fine with doing this as a follow up, but it seems brittle and not future-proof, so I may just do it today.

zzmp (approved):

src/hooks/connectWeb3/useWeb3React.tsx
NoahZinsmeister: hm i tried this and i don't think it's possible - similar issue (without a fix) described here: https://github.com/microsoft/TypeScript/issues/16069

src/hooks/connectWeb3/useWeb3React.tsx
NoahZinsmeister: tracking in https://github.com/Uniswap/widgets/issues/145

PR #147, opened 2022-08-16
fix: guard against undefined window for SSR (eg nextjs)

### LINKED ISSUE #146: NextJS "Window Not Defined" Error
**Bug Description**
NextJS sometimes does server side rendering which does not have access to a 'window' object causing the app to crash before even loading. Seems to be stemming from [here](https://github.com/Uniswap/widgets/blob/ebe37e65ac98b50a04fcb1d9c9e8df820a851046/src/polyfills.ts)

**Steps to Reproduce**
Since it's hard to control when NextJS renders server side, this is somewhat intermittent but generally:
1. Create a new NextJS project
2. Import and insert the `<SwapWidget />` component
3. Reload the page
4. Note that the page does not even load and the app crashes with the error: 
```Server Error ReferenceError: window is not defined. This error happened while generating the page. Any console logs will be displayed in the terminal window.``` 

**Expected Behavior**
The page should load with the swap widget

**Additional Context**
https://github.com/Uniswap/widgets/blob/ebe37e65ac98b50a04fcb1d9c9e8df820a851046/src/polyfills.ts

akarys92:
**Problem:** Fixes #146 when NextJS does a server side render there is no `window` object so the server throws an error and crashes in the `polyfill.js` file. 

**Solution:** Since server side node shouldn't need the polyfill, just check if the window object exists. If it doesn't, we're server side and don't need to attempt to polyfill. 

NOTE: NextJS has explicit dead-code elimination for when guarding against an undefined window as `typeof window`, so this is the canonical way to check for SSR (see https://nextjs.org/blog/next-9#dead-code-elimination-for-typeof-window-branches).


zzmp (changes requested):
Please back out all changes to `yarn.lock`.

zzmp (approved):

PR #150, opened 2022-08-18
fix: respond to events

zzmp:
- Prevents re-keying web3-react when `Widget` props change
  Effectively, this was preventing state from updating in async handlers (such as event handlers).
- Simplifies event handler code throughout:
  - Integrates typings in SwapProps
  - Excludes values from hook deps (to reduce hook calls)
  - Simplifies cosmos `useEventHandler` to handle all arguments

### DIFF
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -160,7 +160,8 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
         : trade.state === TradeState.VALID
         ? {
             onClick: async () => {
-              setOpen((await onReviewSwapClick?.()?.catch(() => false)) ?? true)
+              const open = await onReviewSwapClick?.()?.catch(() => false)
+              setOpen(open ?? true)
             },
           }
         : { disabled: true }
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -160,7 +160,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
         : trade.state === TradeState.VALID
         ? {
             onClick: async () => {
-              setOpen((await onReviewSwapClick?.()) ?? true)
+              setOpen((await onReviewSwapClick?.()?.catch(() => true)) ?? true)
             },
           }
         : { disabled: true }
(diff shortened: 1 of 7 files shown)


src/components/Swap/SwapButton/index.tsx
```diff
@@ -159,9 +159,8 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
         ? { action: approvalAction }
         : trade.state === TradeState.VALID
         ? {
-            onClick: () => {
-              const promise = onReviewSwapClick && onReviewSwapClick()
-              return promise ? promise.then((open) => setOpen(open)) : setOpen(true)
+            onClick: async () => {
+              setOpen((await onReviewSwapClick?.()) ?? true)
```
JFrankfurt, line 163: what happens here if onReviewSwapClick throws?

src/components/Swap/index.tsx
```diff
@@ -45,20 +45,17 @@ function getTransactionFromMap(
...
 }
 
 export default function Swap(props: SwapProps) {
   useValidate(props)
-  useSyncConvenienceFee(props)
-  useSyncTokenDefaults(props)
-  useSyncEventHandlers(props)
+  useSyncEventHandlers(props as EventHandlers)
+  useSyncConvenienceFee(props as FeeOptions)
+  useSyncTokenDefaults(props as TokenDefaults)
```
JFrankfurt, line 58: 👍

src/components/TokenSelect/index.tsx
```diff
@@ -133,18 +133,8 @@ export default memo(function TokenSelect({ collapsed, disabled, field, onSelect,
...
-        .then((open) => {
-          setOpen(open)
-        })
-        .catch(() => {
-          setOpen(false)
-        })
-    }
-    return setOpen(true)
+  const onOpen = useCallback(async () => {
+    setOpen((await onTokenSelectorClick?.(field)) ?? true)
```
JFrankfurt, line 137: I'm pretty sure this drops the `setOpen(false)` on promise rejection functionality

src/components/Swap/SwapButton/index.tsx
zzmp: wow good call

src/components/TokenSelect/index.tsx
zzmp: you're right, I'll back these changes out

zzmp: this was caught by tests, too :)

JFrankfurt (approved):

PR #157, opened 2022-08-22
feat: always show price impact %

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/DE-117

Shows PI even when unable to fetch a stablecoin price impact value.

image
image


src/utils/prices.ts
```diff
@@ -15,6 +15,17 @@ import JSBI from 'jsbi'
 const THIRTY_BIPS_FEE = new Percent(JSBI.BigInt(30), JSBI.BigInt(10000))
 const INPUT_FRACTION_AFTER_FEE = ONE_HUNDRED_PERCENT.subtract(THIRTY_BIPS_FEE)
 
+export function largerPercentValue(a?: Percent, b?: Percent) {
+  if (a && b) {
+    return a.greaterThan(b) ? a : b
+  } else if (a) {
```
zzmp, line 21: L21 and below could be simplified to

    }
    return a || b

src/hooks/swap/useSwapInfo.tsx
```diff
@@ -64,10 +65,27 @@ function useComputeSwapInfo(routerUrl?: string): SwapInfo {
...
   )
 
-  // Compute slippage and impact off of the trade so that it refreshes with the trade.
-  // (Using amountIn/amountOut would show (incorrect) intermediate values.)
   const slippage = useSlippage(trade.trade)
-  const { inputUSDC, outputUSDC, impact } = useUSDCPriceImpact(trade.trade?.inputAmount, trade.trade?.outputAmount)
+  const {
+    inputUSDC,
+    outputUSDC,
+    impact: stablecoinPriceImpact,
```
zzmp, line 72: I think this would be cleaner if you instead computed usdc as part of useSwapInfo, and that passed that into an updated `usePriceImpact` hook (to replace the existing `useUSDCPriceImpact` hook). The only reason the usdc prices are currently part of the price impact hook is to avoid computing them redundantly (because it is amongst the most expensive things we do in terms of compute) - adding actual logic to `useUSDCPriceImpact` makes me thing it'd be better to separate out the usdc computations.

eg

    const inputUSDC = useUSDCValue(inputAmount) ?? undefined
    const outputUSDC = useUSDCValue(outputAmount) ?? undefined
    const impact = usePriceImpact(trade.trade,  {inputUSDC, outputUSDC })

zzmp (changes requested):

src/components/Swap/Output.tsx
```diff
@@ -85,7 +85,7 @@ export default function Output({ disabled, focused, children }: PropsWithChildre
           <ThemedText.Body2 color="secondary" userSelect>
             <Row>
               <USDC gap={0.5} isLoading={isRouteLoading}>
-                {outputUSDC ? `$${formatCurrencyAmount(outputUSDC, 6, 'en', 2)}` : '-'}{' '}
+                {outputUSDC && `$${formatCurrencyAmount(outputUSDC, 6, 'en', 2)} `}
```
zzmp, line 88: The space at the end of the string literal is intentional, right? I _think_ it's necessary, just double-checking.

src/hooks/swap/useSwapInfo.tsx
```diff
@@ -63,24 +64,25 @@ function useComputeSwapInfo(routerUrl?: string): SwapInfo {
     useMemo(() => [currencyIn, currencyOut], [currencyIn, currencyOut])
   )
 
-  // Compute slippage and impact off of the trade so that it refreshes with the trade.
```
zzmp, line 66: Retain the comments on L66-67 (or move them to `useSlippage`).

src/hooks/usePriceImpact.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+    const percent = largerPercentValue(marketPriceImpact, fiatPriceImpact)
+    return percent
+      ? {
+          percent,
+          warning: getPriceImpactWarning(percent),
+          toString: () => toHumanReadablePercent(percent),
+        }
+      : undefined
+  }, [trade])
```
zzmp, line 34: missing deps (usdc values)

zzmp (changes requested):

src/components/Swap/Output.tsx
JFrankfurt: yep! intentional

zzmp (approved):

PR #158, opened 2022-08-22
feat: allow integrators to disable uniswap branding

JFrankfurt:
image


src/components/BrandedFooter.tsx
```diff
@@ -27,6 +29,10 @@ const UniswapA = styled(ExternalLink)`
 `
 
 export default memo(function BrandedFooter() {
+  const disableBranding = useAtomValue(disableBrandingAtom)
+  if (disableBranding) {
```
zzmp, line 33: nit: As a guard I'd prefer this as a one-liner (no brackets).

src/hooks/useSyncBrandingSetting.ts
```diff
@@ -0,0 +1,12 @@
+import { useUpdateAtom } from 'jotai/utils'
```
zzmp, line 1: Instead of splitting the state and the hooks, if this is a one-off, how do you feel about simplifying it by just having a hook file that does not export an atom, eg

    const disableBrandingAtom = atom<boolean>(false)

    export function useDisableBranding(): [boolean, (value: boolean) => void] {
      return useAtom(disableBrandingAtom)
    }

src/components/Swap/index.tsx
```diff
@@ -62,6 +63,7 @@ export default function Swap(props: SwapProps) {
   useSyncSwapEventHandlers(props as SwapEventHandlers)
   useSyncTokenDefaults(props as TokenDefaults)
   useSyncWidgetEventHandlers(props as WidgetEventHandlers)
+  useSyncBrandingSetting(props as BrandingSetting)
```
zzmp, line 66: This should be sync'ed as part of the `Widget` component, not as part of the `Swap` component, as it is widget-scoped state.

zzmp (changes requested):

src/hooks/useSyncBrandingSetting.ts
JFrankfurt: I like this idea a lot :)

src/components/Swap/index.tsx
JFrankfurt: I had the same thought, but the AtomProvider is at the `Widget` component level, so only its children can share atoms

src/components/Swap/Output.tsx
```diff
@@ -98,7 +101,7 @@ export default function Output({ disabled, focused, children }: PropsWithChildre
         </TokenInput>
         <Rule />
         {children}
-        <BrandedFooter />
+        {!disableBranding && <BrandedFooter />}
```
zzmp, line 104: nit: because this is a double negative, it may be more readable as

    {disableBranding ? null : <BrandedFooter />}

src/hooks/useSyncBrandingSetting.ts
```diff
@@ -0,0 +1,14 @@
+import { atom } from 'jotai'
+import { useUpdateAtom } from 'jotai/utils'
+import { useEffect } from 'react'
+
+export const disableBrandingAtom = atom<boolean>(false)
```
zzmp, line 5: If you have the atom in the hook file, you should only expose it through hooks. Why not a wrapper hook?

    export function useBrandingSetting() {
      return useAtomValue(disableBrandingAtom)
    }

zzmp, line 5: If you expect more branding settings, you should shape the atom for them:

    const initialBrandingSettings: BrandingSettings = {}
    const brandingSettingsAtom = atom(initialBrandingSettings)

This will also let you simplify your sync hook, so that any future additions to the `BrandingSettings` type can be effected by just updating the type, and not touching the rest of the implementation.

    export default function useSyncBrandingSettings(settings: BrandingSettings): void {
      const updateSettings = useUpdateAtom(brandingSettingsAtom)
      useEffect(() => updateSettings(settings), [settings, updateSettings])
    }

src/components/Widget.tsx
```diff
@@ -97,6 +97,7 @@ export interface WidgetProps {
   tokenList?: string | TokenInfo[]
   width?: string | number
   dialog?: HTMLDivElement | null
+  disableBranding?: boolean
```
zzmp, line 100: As a sync'ed setting, instead of adding this to widget props, extend widget props from this: `WidgetProps extends BrandingSettings`. Should you add any settings to the `BrandingSettings` type, you won't need to touch the `Widget.tsx` file again.

src/components/Swap/index.tsx
zzmp: This will need to be moved eventually (with `useSyncWidgetEventHandlers`) but I think it's fine for now.

zzmp: On L53, `SwapProps` must extend `BrandingSettings` so that the type system will enforce that these props are available.

zzmp (changes requested):

src/components/Swap/Output.tsx
JFrankfurt: the negative in the prop is so gross to work with, but it really feels like the right name 😱

src/hooks/useSyncBrandingSetting.ts
JFrankfurt: I don't expect more. Do you?

src/components/Widget.tsx
```diff
@@ -97,6 +98,7 @@ export interface WidgetProps extends TransactionEventHandlers {
...
```
zzmp, line 101: Remove the explicit `disableBranding` prop, as it's already included in `WidgetProps` through `extends BrandingSettings`.

zzmp (approved):
lgtm, but remove the explicit prop from WidgetProps

PR #159, opened 2022-08-23
feat: include Trade tx info

zzmp:
Include more usable classes in the swap/wrap transaction info to facilitate more featureful integrations.

- Refactors wrap logic to re-use the existing `TransactionType`
- Includes fully instantiated `Trade` and `CurrencyAmount` in transaction info


src/components/Swap/SwapButton/index.tsx
```diff
@@ -173,11 +174,10 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
   }, [approvalAction, disableSwap, onReviewSwapClick, isPending, onWrap, trade.state, wrapType])
   const Label = useCallback(() => {
     switch (wrapType) {
```
vm, line 176: this variable should be `transactionType` now, plus wherever else in this file

vm: or leave as WrapType?

src/components/Swap/Status/StatusDialog.tsx
```diff
@@ -20,7 +26,7 @@ const TransactionRow = styled(Row)`
   flex-direction: row-reverse;
 `
 
-type PendingTransaction = Transaction<SwapTransactionInfo | WrapTransactionInfo>
+type PendingTransaction = Transaction<SwapTransactionInfo | WrapTransactionInfo | UnwrapTransactionInfo>
```
vm, line 29: why can't this be `Transaction<TransactionInfo>`?

vm (approved):
Some of the naming of `WrapType` vs `TransactionType` is confusing. I might consider leaving only `TransactionType` and then having a function called `isWrap` which returns true of Wrap | Unwrap.

src/components/Swap/Status/StatusDialog.tsx
zzmp: NB: That's equivalent to just `Transaction` (as `TransactionInfo` defaults).

src/hooks/swap/useWrapCallback.tsx
```diff
@@ -5,15 +5,13 @@ import { useWETHContract } from 'hooks/useContract'
...
 import tryParseCurrencyAmount from 'utils/tryParseCurrencyAmount'
 
 import useCurrencyBalance from '../useCurrencyBalance'
 
-export enum WrapType {
-  NONE,
-  WRAP,
-  UNWRAP,
-}
+export type WrapType = TransactionType.WRAP | TransactionType.UNWRAP | undefined
```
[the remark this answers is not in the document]
zzmp: only directly below it - I'll just inline the usage.

zzmp:
> Some of the naming of `WrapType` vs `TransactionType` is confusing. I might consider leaving only `TransactionType` and then having a function called `isWrap` which returns true of Wrap | Unwrap.

I added `isWrap` to the `useWrapCallback` return value, and made a new helper hook, `useIsWrap`, to cover this use case and clarify the code. Hope that resolves this 🤞

PR #160, opened 2022-08-23
fix: icon button height

zzmp:
Fixes alignment of the info icon.
Updates header and icon button heights to avoid misalignments.

See https://uniswaplabs.atlassian.net/browse/WEB-891.


cmcewen (approved):
screenshots are nice 😄

PR #162, opened 2022-08-23
fix: allow less than signed amount for swaps

zzmp:
Allows a user to trade _less than_ the signed amount in a swap.
Eg if a user allows 0.1 UNI to be swapped, this will allow 0.05 UNI to be swapped without requiring a new signature for the lesser amount.


cmcewen (approved):
ah makes sense

ianlapham (approved):
lgtm

PR #165, opened 2022-08-24
feat: reset inputs after submission

zzmp:
Zeroes out the inputs and sets Field.INPUT as the independent field (ie TradeType.EXACT_INPUT) once a swap or wrap is submitted. This is in line with github.com/Uniswap/interface, as per @infredible.

https://user-images.githubusercontent.com/5403956/186292639-13744922-1b1e-4fb4-b5ee-d5b477fd88c5.mov

### DIFF
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -1,7 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { CurrencyAmount } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
-import { useSwapInfo } from 'hooks/swap'
+import { useSwapAmount, useSwapInfo } from 'hooks/swap'
 import { useSwapApprovalOptimizedTrade } from 'hooks/swap/useSwapApproval'
 import { useSwapCallback } from 'hooks/swap/useSwapCallback'
 import useWrapCallback from 'hooks/swap/useWrapCallback'
@@ -69,12 +69,16 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
 
   const addTransactionInfo = useAddTransaction()
   const setDisplayTxHash = useUpdateAtom(displayTxHashAtom)
+  const [, setSwapAmount] = useSwapAmount(Field.INPUT)
+  const resetSwapAmount = useCallback(() => setSwapAmount(''), [setSwapAmount])
 
   const onSubmit = useCallback(
     async (submit: () => Promise<TransactionInfo | undefined>): Promise<void> => {
       try {
         const info = await submit()
         if (!info) return
+
+        resetSwapAmount()
         addTransactionInfo(info)
         setDisplayTxHash(info.response.hash)
 
@@ -93,7 +97,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
         console.error('Failed to submit', e)
       }
     },
-    [addTransactionInfo, setDisplayTxHash]
+    [addTransactionInfo, resetSwapAmount, setDisplayTxHash]
   )
 
   const [isPending, setIsPending] = useState(false)
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -111,7 +111,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
       const response = await wrapCallback()
       if (!response) return
 
-      invariant(wrapType)
+      invariant(wrapType !== undefined)
       const amount = CurrencyAmount.fromRawAmount(native, response.value?.toString() ?? '0')
       return { response, type: wrapType, amount }
     })
--- a/src/components/Swap/SwapButton/index.tsx
+++ b/src/components/Swap/SwapButton/index.tsx
@@ -5,7 +5,7 @@ import { useSwapAmount, useSwapInfo } from 'hooks/swap'
 import { useSwapApprovalOptimizedTrade } from 'hooks/swap/useSwapApproval'
 import { useSwapCallback } from 'hooks/swap/useSwapCallback'
 import useWrapCallback from 'hooks/swap/useWrapCallback'
-import { useAddTransaction } from 'hooks/transactions'
+import { useAddTransactionInfo } from 'hooks/transactions'
 import { useSetOldestValidBlock } from 'hooks/useIsValidBlock'
 import useNativeCurrency from 'hooks/useNativeCurrency'
 import useTransactionDeadline from 'hooks/useTransactionDeadline'
@@ -67,7 +67,7 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
   // Close the review modal on chain change.
   useEffect(() => setOpen(false), [chainId])
 
-  const addTransactionInfo = useAddTransaction()
+  const addTransactionInfo = useAddTransactionInfo()
   const setDisplayTxHash = useUpdateAtom(displayTxHashAtom)
   const [, setSwapAmount] = useSwapAmount(Field.INPUT)
   const resetSwapAmount = useCallback(() => setSwapAmount(''), [setSwapAmount])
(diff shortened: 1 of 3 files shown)


src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,73 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
-  const addTransaction = useAddTransaction()
+  const addTransactionInfo = useAddTransaction()
   const setDisplayTxHash = useUpdateAtom(displayTxHashAtom)
-  const setOldestValidBlock = useSetOldestValidBlock()
+  const [, setSwapAmount] = useSwapAmount(Field.INPUT)
+  const resetSwapAmount = useCallback(() => setSwapAmount(''), [setSwapAmount])
+
+  const onSubmit = useCallback(
+    async (submit: () => Promise<TransactionInfo | undefined>): Promise<void> => {
+      try {
```
vm, line 77: which part of this actually throws an error? would be better to wrap only that piece

src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,73 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
-      if (isAnimating(document)) {
-        document.addEventListener('animationend', postWrap)
-      } else {
-        setIsPending(false)
-      }
-    } catch (e) {
-      // TODO(zzmp): Surface errors from wrap.
-      console.log(e)
-    } finally {
-      setIsPending(false)
```
vm, line 103: is this not necessary anymore?

zzmp: `onSubmit` will always return, so a `finally` is no longer needed.

src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,73 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
-      }
-    } catch (e) {
-      // TODO(zzmp): Surface errors from swap.
-      console.log(e)
-    }
-  }, [addTransaction, setDisplayTxHash, setOldestValidBlock, swapCallback, trade.trade])
+      invariant(trade.trade)
+      return { type: TransactionType.SWAP, response, tradeType: trade.trade.tradeType, trade: trade.trade }
+    })
+    setOpen(false)
```
zzmp: This should only be called in the case of a successful submission.

src/components/Swap/SwapButton/index.tsx
zzmp: The submission. Done.

src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,82 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
-        response: transaction,
-        tradeType: trade.trade.tradeType,
-        trade: trade.trade,
-      })
-      setDisplayTxHash(transaction.hash)
+    const submitted = await onSubmit(async () => {
+      const response = await swapCallback?.()
+      if (!response) return
 
       // Set the block containing the response to the oldest valid block to ensure that the
```
JFrankfurt, line 131: shouldn't all these `oldest` references be `newest`?

src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,82 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
-      }
-      if (isAnimating(document)) {
-        document.addEventListener('animationend', postSwap)
-      } else {
-        setOpen(false)
-      }
-    } catch (e) {
-      // TODO(zzmp): Surface errors from swap.
-      console.log(e)
+      invariant(trade.trade)
```
JFrankfurt, line 137: is adding an `invariant` call here the correct move? iirc it throws if the argument is falsy, so this would be an uncaught error and trigger the error boundary.

src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,82 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
   // Close the review modal on chain change.
   useEffect(() => setOpen(false), [chainId])
 
-  const addTransaction = useAddTransaction()
+  const addTransactionInfo = useAddTransaction()
   const setDisplayTxHash = useUpdateAtom(displayTxHashAtom)
-  const setOldestValidBlock = useSetOldestValidBlock()
+  const [, setSwapAmount] = useSwapAmount(Field.INPUT)
+  const resetSwapAmount = useCallback(() => setSwapAmount(''), [setSwapAmount])
```
JFrankfurt, line 73: I was looking for *where* the the `Field.INPUT` setting was happening, then realized it's probably happening implicitly here. Just want to point out that this feels a little hard to parse. I don't know if someone reading this code w/o context would be able to recognize that is happening.

src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,82 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
   // Close the review modal on chain change.
   useEffect(() => setOpen(false), [chainId])
 
-  const addTransaction = useAddTransaction()
+  const addTransactionInfo = useAddTransaction()
```
JFrankfurt, line 70: should this hook name change?

src/components/Swap/SwapButton/index.tsx
```diff
@@ -67,80 +67,82 @@ export default memo(function SwapButton({ disabled }: SwapButtonProps) {
...
-      console.log(e)
-    } finally {
-      setIsPending(false)
-    }
-  }, [addTransaction, native, setDisplayTxHash, wrapCallback, wrapType])
+    await onSubmit(async () => {
+      const response = await wrapCallback()
+      if (!response) return
+
+      invariant(wrapType)
```
JFrankfurt, line 114: This is a subtly different assertion than the one being removed. `WrapType.NONE` (`0`) will throw here now. Is that intentional?

JFrankfurt (approved):

src/components/Swap/SwapButton/index.tsx
zzmp: This is adding a `TransactionInfo` type (`Transaction` type is different), so I'm going to propagate this change to all callsites.

src/components/Swap/SwapButton/index.tsx
zzmp: `WrapType.NONE` was removed in https://github.com/Uniswap/widgets/pull/159 - this is now a type alias for `TransactionType.WRAP | TransaactionType.UNWRAP | undefined`.

I can (and should) still make this more explicit, so I'll update to `wrapType !== undefined`.

src/components/Swap/SwapButton/index.tsx
zzmp: I thought that by semantically naming a derived setter I would make it easier to parse, so this is great feedback. I'll opt for the more explicit `setSwapAmount('')` inlined, and alias `setSwapAmount` to `setInputAmount` or similar.

src/components/Swap/SwapButton/index.tsx
zzmp: we want the `oldest` valid block because that implies that nothing older than that is still valid.
I'm happy to change this to something less ambiguous in a separate PR, but would prefer not to change the naming in this PR.

src/components/Swap/SwapButton/index.tsx
zzmp: `onWrap` will not be called if `wrapType === undefined`, but due to the complex type that's not evident to the type checker. We _could_ replace this with a check, but if `wrapCallback()` returns a valid response, then `wrapType` should also be valid. I can annotate this with a comment.

src/components/Swap/SwapButton/index.tsx
JFrankfurt: specifically I mean it's not immediately clear to me that these lines reset the widget to an `exactInput` swap if it was not immediately in that state. The derived setter makes sense for the `Amount` for sure, just not the `Field`/`tradeType`

PR #167, opened 2022-08-24
fix: slide modal out on other side

zzmp:
This PR changes the transition behavior for modals.

If there is another dialog taking the place of an active one, the old dialog will slide out to the left (instead of to the right). In code, this is called "PAGING". This avoids risking exposing the screen underneath modals when they are replacing each other, and makes the modals look more sequenced.

It also fixes some issues I discovered while implementing this:
- Displays unwrap transactions, which was broken by #159 
- Delays input reset until after animations, to avoid layout thrashing

Before

https://user-images.githubusercontent.com/5403956/186477938-81c2ded0-4364-4bf9-a25d-4bf8a823710f.mov

After

https://user-images.githubusercontent.com/5403956/186519113-772405ce-da6e-44b4-9f8e-e551904cb471.mov

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -129,6 +129,9 @@ export default function Dialog({ color, children, onClose }: DialogProps) {
 
   const modal = useRef<HTMLDivElement>(null)
   useUnmountingAnimation(modal, () => {
+    // Returns the context element's child count at the time of unmounting.
+    // This cannot be done through state because the count is updated outside of React's lifecycle -
+    // it *must* be checked at the time of unmounting in order to include the next page of Dialog.
     return (context.element?.childElementCount ?? 0) > 1 ? Animation.PAGING : Animation.CLOSING
   })
 
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -23,7 +23,12 @@ declare global {
 export enum Animation {
   /** Used when the Dialog is closing. */
   CLOSING = 'closing',
-  /** Used when the Dialog is paging to another Dialog screen. */
+  /**
+   * Used when the Dialog is paging to another Dialog screen.
+   * Paging occurs when multiple screens are sequenced in the Dialog, so that an action that closes
+   * one will simultaneously open the next. Special-casing paging animations can make the user feel
+   * like they are not leaving the Dialog, despite the initial screen closing.
+   */
   PAGING = 'paging',
 }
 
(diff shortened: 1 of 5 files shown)


JFrankfurt:
This results in an unexpected animation (imo) when clicking the back arrow on this screen: 
image

can you make this change conditional on transaction submission instead of universal?

JFrankfurt (changes requested):
This results in an unexpected animation (imo) when clicking the back arrow on this screen: 
image

can you make this change conditional on transaction submission instead of universal?

zzmp:
> This results in an unexpected animation (imo) when clicking the back arrow on this screen: image
> 
> can you make this change conditional on transaction submission instead of universal?

To clarify (from @JFrankfurt):

I don’t expect the back arrow pointing left < to send the screen that direction
I expect it to send the screen back the direction it came (to the right)
I think this change is a much improved transition to the next screen, but a worse experience when hitting “back” from the swap summary

infredible:
lgtm

src/components/Dialog.tsx
```diff
@@ -20,6 +20,13 @@ declare global {
   }
 }
 
+export enum Animation {
+  /** Used when the Dialog is closing. */
+  CLOSING = 'closing',
+  /** Used when the Dialog is paging to another Dialog screen. */
+  PAGING = 'paging',
```
vm, line 27: perhaps a good place to explain the concept of paging and why it's important

src/components/Widget.tsx
```diff
@@ -59,6 +58,11 @@ const slideIn = keyframes`
     transform: translateX(calc(100% - 0.25em));
   }
 `
+const slideBack = keyframes`
```
vm, line 61: would consider renaming `slideBack` or `slideOut` since they don't feel like opposites. maybe `slideIn` and `slideOut`. or `slideRight` and `slideLeft`?

src/utils/animations.ts
```diff
@@ -1,26 +1,28 @@
...
+  useEffect(() => {
+    const current = node.current
+    const parent = current?.parentElement
+    const removeChild = parent?.removeChild
+    if (!(parent && removeChild)) return
+
     parent.removeChild = function <T extends Node>(child: T) {
       if ((child as Node) === current) {
-        current.classList.add(UNMOUNTING)
+        const klass = typeof animatingClass === 'string' ? animatingClass : animatingClass()
```
vm, line 24: not sure i'm following here. why could animating class be a class or string?

src/components/Widget.tsx
zzmp: `slideRight` can mean sliding in or out, but `slideInLeft` conveys the entire action.
Note that I'm naming the action, and leaving it to the CSS (not the keyframes) to assign semantic meaning.

src/components/Dialog.tsx
zzmp: done

src/utils/animations.ts
zzmp: I've narrowed the type (to a callback) and tried to explain why this is a callback (and not a string) through comments.

vm (approved):

PR #168, opened 2022-08-24
fix: token select with less friction

zzmp:
Selects the top token from the token selector by default.
This allows a user to eg type `UNI<ENTER>` to use the UNI token, instead of needing to use arrows/mouse to select it from the list.

### DIFF
--- a/src/components/TokenSelect/TokenOptions.tsx
+++ b/src/components/TokenSelect/TokenOptions.tsx
@@ -12,6 +12,7 @@ import {
   memo,
   SyntheticEvent,
   useCallback,
+  useEffect,
   useImperativeHandle,
   useMemo,
   useRef,
@@ -135,7 +136,6 @@ areEqual)
 
 interface TokenOptionsHandle {
   onKeyDown: (e: KeyboardEvent) => void
-  blur: () => void
 }
 
 interface TokenOptionsProps {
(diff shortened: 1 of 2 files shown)


infredible:
lgtm

src/components/TokenSelect/TokenOptions.tsx
```diff
@@ -148,8 +148,11 @@ const TokenOptions = forwardRef<TokenOptionsHandle, TokenOptionsProps>(function
   ref
 ) {
   const [focused, setFocused] = useState(false)
-  const [selected, setSelected] = useState<Currency>()
-  const hover = useMemo(() => (selected ? tokens.indexOf(selected) : -1), [selected, tokens])
+  const [selected, setSelected] = useState<Currency>(tokens[0])
+  useEffect(() => {
+    setSelected((selected) => (tokens.includes(selected) ? selected : tokens[0]))
+  }, [tokens, setSelected])
```
vm, line 154: in practice tokens will never change between renders right? so this is only run on load

vm (approved):

src/components/TokenSelect/TokenOptions.tsx
zzmp: no, this will run when a user modifies the `tokens` by entering search terms.
eg when a user enters `UNI` `tokens` will be re-passed as `[<UNI>]` and the component will re-render.

Effectively this is a guard that says "if there is no selected token when the user searches, select the one at the top of the list".

vm: ah! i would definitely explain that in code somehow. not obvious. either a comment or name some variables

PR #170, opened 2022-08-25
fix: integration issues

zzmp:
- Fixes spurious rerenders
- Exports more used types (for integrators)
- Makes the jsonRpcUrlMap more permissive by allowing singletons _or_ arrays
- Treats max slippage as a string to avoid layout thrashing / eliding the decimal in integrations
- Fixes popper's max size treatment


src/components/Popover.tsx
```diff
@@ -113,9 +113,10 @@ export default function Popover({ content, show, children, placement, offset, co
           requires: ['maxSize'],
           fn({ state }) {
             const { width } = state.modifiersData.maxSize
+            const maxWidth = Math.max(width, (boundary?.clientWidth ?? 16) - 16)
             state.styles.popper = {
               ...state.styles.popper,
-              maxWidth: `${width}px`,
+              maxWidth: `${maxWidth}px`,
```
vm, line 119: why not do width and maxWidth separately?

src/components/Widget.tsx
```diff
@@ -143,7 +143,7 @@ export function TestableWidget(props: PropsWithChildren<TestableWidgetProps>) {
     }
     return props.defaultChainId
   }, [props.defaultChainId])
-  const jsonRpcUrlMap: string | JsonRpcProvider | { [chainId: number]: string[] } = useMemo(() => {
+  const jsonRpcUrlMap: { [chainId: number]: string[] } = useMemo(() => {
```
vm, line 146: would prob define this type for jsonRpcUrlMap above somewhere

src/components/Widget.tsx
```diff
@@ -154,7 +154,10 @@ export function TestableWidget(props: PropsWithChildren<TestableWidgetProps>) {
         props.jsonRpcUrlMap[supportedChainId as number] = fallbackRpc
       }
     }
-    return props.jsonRpcUrlMap
+    return Object.entries(props.jsonRpcUrlMap).reduce(
```
vm, line 157: can you make a function for this or name the result as a variable? will make it clearer what's going on

src/hooks/useSlippage.ts
```diff
@@ -5,9 +5,10 @@ import { useMemo } from 'react'
 import { InterfaceTrade } from 'state/routing/types'
 import { slippageAtom } from 'state/swap/settings'
 
-export function toPercent(maxSlippage: number | undefined): Percent | undefined {
+export function toPercent(maxSlippage: string | undefined): Percent | undefined {
```
vm, line 8: what necessitated this change? feels kinda unrelated to the rest of the pr

e2e/connect.test.tsx
```diff
@@ -45,6 +45,16 @@ describe('connect', () => {
       expect(toolbar.textContent).toBe('Connect wallet to swap')
     })
 
+    it('expects widget not to be disabled', async () => {
```
vm, line 48: should this be separated into separate tests? or it needs to happen in sync

src/components/Swap/Settings/MaxSlippageSelect.tsx
```diff
@@ -113,9 +113,8 @@ export default function MaxSlippageSelect() {
   }, [focus, slippage, setSlippage])
 
   const processInput = useCallback(
-    (input: string | undefined) => {
-      setMaxSlippageInput(input || '')
-      const max = input ? +input : undefined
+    (max: string | undefined) => {
+      setMaxSlippageInput(max || '')
```
JFrankfurt, line 117: this line has always bothered me a little, but are you sure you want to drop the type coercion here?

src/components/Swap/Settings/TransactionTtlInput.tsx
```diff
@@ -31,7 +31,7 @@ export default function TransactionTtlInput() {
           <IntegerInput
             placeholder={placeholder}
             value={ttl?.toString() ?? ''}
-            onChange={(value) => setTtl(value ? parseFloat(value) : 0)}
+            onChange={(value) => setTtl(value ? parseFloat(value) : undefined)}
```
JFrankfurt, line 34: 💯

src/components/Widget.tsx
```diff
@@ -154,7 +154,10 @@ export function TestableWidget(props: PropsWithChildren<TestableWidgetProps>) {
...
+      (urlMap, [id, urls]) => ({ ...urlMap, [id]: Array.isArray(urls) ? urls : [urls] }),
```
JFrankfurt, line 158: is the `.isArray` call necessary here with the given jsonRpcUrlMap type?

src/state/swap/settings.ts
```diff
@@ -5,7 +5,7 @@ import { pickAtom } from '../atoms'
 
 export interface Slippage {
   auto: boolean // if true, slippage will use the default calculation
-  max: number | undefined // expressed as a percent (eg 0.42 is 0.42%)
+  max: string | undefined // expressed as a percent (eg '0.42' is 0.42%)
```
JFrankfurt, line 8: fwiw this is out of sync w/ our fee parameter, which is integer basis points

JFrankfurt (approved):

src/components/Popover.tsx
zzmp: We don't want to set `width` in case the popover can be smaller than the total width - just the `maxWidth`.
I override it here because there were issues in integration where it was being set to 12px, and I couldn't figure out why, but the `boundary` being passed was correct so I just overwrote it with a valid value.

zzmp: Adding a comment inline to explain this.

src/components/Widget.tsx
zzmp: I'm going to omit this type because it's actually enforced when passed to `ActiveWeb3Provider`.

src/hooks/useSlippage.ts
zzmp: This is required for integrations to have smooth input. Otherwise, things like '' or '1.' will change to a number (like 0 or 1) and give a bad user experience. It's a follow-up for maxSlippage becoming a string, which is done because it's input, so it needs to be treated like a string and not a number for better UX.

e2e/connect.test.tsx
zzmp: I'll put the singleton test in its own describe block

src/components/Swap/Settings/MaxSlippageSelect.tsx
zzmp: type coercion is no longer necessary, as the coercion will now happen in `toPercent`.
The logic is necessary for smooth UX.

src/components/Widget.tsx
zzmp: This is converting the value from `string[] | string` to `string[]`, so `.isArray` is the easiest way to disambiguate here.

src/components/Widget.tsx
zzmp: I split this into so that Widget.tsx only deals with fallbacks, so that this callback only has a single function.
Converting from `string|string[]` to `string[]` is now done in the child component.

vm (approved):

PR #171, opened 2022-08-25
fix: clean up integration exports

zzmp:


src/state/swap/index.ts
```diff
@@ -59,23 +52,20 @@ export type OnTokenChange = (field: Field, token: Currency) => void
...
  */
-export type OnReviewSwapClick = () => void | Promise<boolean>
+export type OnReviewSwapClick = () => void | boolean | Promise<boolean>
 
 /**
  * An integration hook called when the user clicks the token selector.
  * If the hook resolve to false or rejects, the token selector will not open.
  */
-export type OnTokenSelectorClick = (field: Field) => void | Promise<boolean>
+export type OnTokenSelectorClick = (field: Field) => void | boolean | Promise<boolean>
```
vm, line 68: can you explain why this change is desirable?

zzmp: In the case that you always want to disable token selection, it's easier to write `() => false` than `() => Promise.resolve(false)`.

vm (approved):

PR #172, opened 2022-08-26
refactor: use TradeType instead of derived independent field

zzmp:
- Replaces usages of `independentField: Field` with `type: TradeType`, the primitive exposed through the SDK.
- Adds utilities (`invertTradeType` and `toTradeType`) to make it easier to deal with `TradeType`s.
- Replaces explicit checks of `tradeType === TradeType.EXACT_INPUT` with a boolean conversion `isExactInput` to standardize these checks.
- Standardizes `SwapController` to match internal `Swap` input (which makes internal conversion logic easier / unnecessary).

### DIFF
--- a/src/hooks/swap/useSwapInfo.tsx
+++ b/src/hooks/swap/useSwapInfo.tsx
@@ -9,6 +9,7 @@ import { useAtomValue } from 'jotai/utils'
 import { createContext, PropsWithChildren, useContext, useMemo } from 'react'
 import { InterfaceTrade, TradeState } from 'state/routing/types'
 import { Field, swapAtom } from 'state/swap'
+import { isExactInput } from 'utils/tradeType'
 import tryParseCurrencyAmount from 'utils/tryParseCurrencyAmount'
 
 import { useIsWrap } from './useWrapCallback'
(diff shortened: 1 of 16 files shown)


src/hooks/swap/useSwapInfo.tsx
```diff
@@ -34,28 +35,27 @@ interface SwapInfo {
...
+    [amount, type, currencyIn, currencyOut]
   )
   const hasAmounts = currencyIn && currencyOut && parsedAmount && !isWrap
   const trade = useRouterTrade(
-    isExactIn ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT,
+    type,
     routerUrl,
     hasAmounts ? parsedAmount : undefined,
-    hasAmounts ? (isExactIn ? currencyOut : currencyIn) : undefined
+    hasAmounts ? (isExactInput(type) ? currencyOut : currencyIn) : undefined
```
ianlapham, line 49: Nit: can `isExactInput(type)` be moved to a constant here and re-used?

ianlapham (approved):
Looks good small nit

src/hooks/swap/useSwapInfo.tsx
ianlapham: (as we use it throughout the file?)

zzmp: I'd like to keep it with the helper function because it's just a boolean check, and doing so keeps it consistent (so - hopefully - easier to read).

PR #176, opened 2022-08-29
feat: expose transaction types

JFrankfurt:


src/index.tsx
```diff
@@ -30,6 +30,7 @@ export type {
 } from 'state/swap'
 export { Field } from 'state/swap'
 export type { Slippage } from 'state/swap/settings'
+export type { Transaction, TransactionInfo, TransactionType } from 'state/transactions'
```
zzmp, line 33: Should this include Swap, ExactInput, ExactOutput, Wrap, and Unwrap transactions?

zzmp (approved):

src/index.tsx
JFrankfurt: I didn't need them, but I can't think of a reason not to.

PR #177, opened 2022-08-29
fix: clean SwapProps and standardize event handler logic

zzmp:


src/hooks/useConditionalHandler.ts
```diff
@@ -0,0 +1,20 @@
...
+    async (...params) => {
+      if (!handler) return true
+
+      try {
+        const result = await handler(...params)
+        if (result === false) return false
+      } catch {
+        return false
+      }
+      return true
```
JFrankfurt, line 16: can replace these returns with a single `finally` block

JFrankfurt (approved):
one question and a nit--lgtm

src/components/Swap/index.tsx
```diff
@@ -33,13 +33,7 @@ import useValidate from './useValidate'
 // since the wallet connection component exists within the Swap component.
 // This includes useSyncWidgetEventHandlers.
 // TODO(zzmp): refactor WalletConnection outside of Swap component
-export interface SwapProps
-  extends BrandingSettings,
-    FeeOptions,
-    SwapController,
```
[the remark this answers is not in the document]
zzmp: Yes, it should only need supplied through the 'value' prop now

src/hooks/useConditionalHandler.ts
zzmp: I think a `finally` block will always run, so it overrides the returned value from within the try-catch block.

PR #181, opened 2022-08-30
fix: add correct values to example

JFrankfurt:
ts errors here
- (can't have `undefined`) from the fixture controller
- must have `Field` keys


src/cosmos/ControlledSwap.fixture.tsx
```diff
@@ -42,10 +42,10 @@ function Fixture() {
     <Row align="start" justify="space-around">
       <SwapWidget
         value={{
-          type,
+          type: type || TradeType.EXACT_INPUT,
```
zzmp, line 45: This should be typed such that defaulting is not necessary

zzmp (changes requested):

PR #185, opened 2022-08-31
fix: fixing scroll, z-index and removing a dash

aballerr:
Tasks
https://uniswaplabs.atlassian.net/browse/WEB-985
https://uniswaplabs.atlassian.net/browse/WEB-953
z-index: I noticed this when using the site that there is overlap of the navbar and widget

Screen Shot 2022-08-31 at 12 42 43 PM
Screen Shot 2022-08-31 at 1 11 42 PM
Screen Shot 2022-08-31 at 1 11 49 PM

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -5,7 +5,7 @@ import { largeIconCss } from 'icons'
 import { createContext, ReactElement, ReactNode, useContext, useEffect, useRef, useState } from 'react'
 import { createPortal } from 'react-dom'
 import styled from 'styled-components/macro'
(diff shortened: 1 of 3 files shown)


src/components/Dialog.tsx
```diff
@@ -111,7 +111,7 @@ export const Modal = styled.div<{ color: Color }>`
   position: absolute;
   top: 0;
   width: 100%;
-  z-index: ${Layer.DIALOG};
+  z-index: 1;
```
aballerr, line 114: do we want to add a layer and if so what should we call it? If this is left as Dialog, it has overlap seen in the attached image

PR #186, opened 2022-08-31
fix: fixing scroll, z-index and removing a dash

aballerr:
Tasks
https://uniswaplabs.atlassian.net/browse/WEB-985
https://uniswaplabs.atlassian.net/browse/WEB-953
z-index: I noticed this when using the site that there is overlap of the navbar and widget
Edit: z-index addressed here https://github.com/Uniswap/interface/pull/4576

Screen Shot 2022-08-31 at 12 42 43 PM
Screen Shot 2022-08-31 at 1 11 42 PM
Screen Shot 2022-08-31 at 1 11 49 PM


src/components/Dialog.tsx
```diff
@@ -111,7 +111,7 @@ export const Modal = styled.div<{ color: Color }>`
   position: absolute;
   top: 0;
   width: 100%;
-  z-index: ${Layer.DIALOG};
+  z-index: 1;
```
aballerr, line 114: there isn't a layer for 1, I can add one if wanted

src/components/Swap/Settings/index.tsx
```diff
@@ -33,7 +31,7 @@ export function SettingsDialog() {
           </ThemedText.ButtonSmall>
         </TextButton>
       </Header>
-      <Column gap={1} style={{ paddingTop: '1em' }} ref={setBoundary} padded css={scrollbar}>
+      <Column gap={1} style={{ paddingTop: '1em' }} ref={setBoundary} padded>
```
aballerr, line 34: It's not obvious to me that we need this scrolling styling.  I believe we are okay to remove it

src/components/Dialog.tsx
zzmp: Why are you not using `Layer.DIALOG` for the `Dialog`?
If this is shared by another layer, then _that_ layer should be renamed.

src/components/Swap/Settings/index.tsx
zzmp: It used to be necessary (there were more settings), but with the pared down settings menu there should never be a scrollbar 👍

zzmp (changes requested):
This is blocked by `Layer`. If it's easy to make it three separate PRs it'll get the other two bits in faster.

src/components/Dialog.tsx
aballerr: if you look at the description you can see the modal is overlapping on the navbar.  Changing this to a z-index of 1 fixes that.

zzmp: z-index's should only be done through the `Layer` enum to keep them organized.
This should be updated in src/theme/layer.ts, and continue using `Layer.DIALOG`.

zzmp (changes requested):

src/components/Dialog.tsx
aballerr: lynn solved this issue here: https://github.com/Uniswap/interface/pull/4576.  I'm going to revert

zzmp (approved):

PR #188, opened 2022-09-01
refactor: split SwapActionButton into child components

zzmp:
Modularizes the logic in the SwapActionButton to make it easier to maintain.
This will also make it simpler to add a "Switch Network" button as a fast follow-up.


JFrankfurt (approved):
This is definitely an improvement, so approving. I really wish we could get a single `ActionButton` component permanently rendered here and just switch out props on it as needed for each of these buttons, though.

PR #189, opened 2022-09-01
feat: switch network when input currency is on another chain

zzmp:
Adds chain switching button to when the input currency is not on the active chain.
Sets jsonRpcUrlMap as state, and uses the fallback URLs in case switching to the user-specified chain fails.

To test:
- Use the ControlledSwap fixture
- Change the input to USDC on the desired chain

Based on (and deprecates) #70.


src/constants/chainInfo.ts
```diff
@@ -25,7 +25,7 @@ interface BaseChainInfo {
   readonly nativeCurrency: {
     name: string // e.g. 'Goerli ETH',
     symbol: string // e.g. 'gorETH',
-    decimals: number // e.g. 18,
+    decimals: 18 // e.g. 18,
```
JFrankfurt, line 28: I don't think this is a safe change--what's your reasoning for making it?

src/constants/errors.ts
```diff
@@ -0,0 +1,3 @@
+export enum ErrorCode {
+  USER_REJECTED = 4001,
```
JFrankfurt, line 2: can you add a comment documenting the source of this error code?

JFrankfurt:
I experienced a weird issue here... Not 100% sure about repro yet, but I think I set an input of mainnet dai, and an output of optimism usdc. I then connected my wallet and got this prompt, and the output token was suddenly on kovan? then I got a web3 react error about referentially static connectors (assuming your other pr solves that one)

image

src/constants/chainInfo.ts
zzmp: This is the definition of an ethereum-native currency according to web3-react: https://github.com/Uniswap/web3-react/blob/main/packages/types/src/index.ts#L62.
We need it to inter-operate. All of our defined native currencies use `18`, so it's internally safe - it's just a narrowing.

JFrankfurt: Not sure if we can rely on this for all chains we might deploy to.

zzmp:
> I experienced a weird issue here... Not 100% sure about repro yet, but I think I set an input of mainnet dai, and an output of optimism usdc. I then connected my wallet and got this prompt, and the output token was suddenly on kovan? then I got a web3 react error about referentially static connectors (assuming your other pr solves that one)

I can't reproduce this, but it should be resolved by the other PR.

src/components/Swap/SwapActionButton/ApproveButton.tsx
```diff
@@ -20,19 +20,21 @@ export function useIsPendingApproval(token?: Token, spender?: string): boolean {
...
-  onSubmit: (submit: () => Promise<ApprovalTransactionInfo | undefined>) => Promise<boolean>
+  color: keyof Colors
   trade: ReturnType<typeof useSwapApprovalOptimizedTrade>
   approvalState: ApproveOrPermitState
   handleApproveOrPermit: () => Promise<{
     response: TransactionResponse
     tokenAddress: string
     spenderAddress: string
   } | void>
+  onSubmit: (submit: () => Promise<ApprovalTransactionInfo | undefined>) => Promise<boolean>
```
JFrankfurt, line 37: I think at this point we should have a standalone interface for the props

JFrankfurt (approved):

PR #194, opened 2022-09-08
fix: connections

zzmp:
- Improves performance (and first-render flashing) around connections / jsonRpcUrlMap / defaultChainId
- Moves connectors into the React lifecycle
- Fixes e2e tests to actually test SwapWidget (and not a weirdly double-test-harnessed Swap in Widget in TestableWidget)
  - This should fix the flakiness around e2e tests, as the rendered components had been in an invalid state


src/hooks/web3/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  // Attempt to connect eagerly if there is no user-provided provider.
+  useEffect(() => {
+    // Ignore any errors during connection so they do not propagate to the widget.
+    if (connectors.user) {
+      connectors.user.activate().catch(() => undefined)
+      return
+    }
+    for (const connector of [connectors.metaMask, connectors.walletConnect]) {
+      connector.connectEagerly().catch(() => undefined)
+    }
```
zzmp, line 76: This must continue to activate the network connector.

e2e/connect.test.tsx
```diff
@@ -5,84 +5,78 @@
...
+  }
 
-  describe('with valid jsonRpcUrlMap', () => {
-    it('prompts for wallet connection in the Wallet and Toolbar', async () => {
-      const component = renderWidget(<SwapWidget tokenList={tokens} jsonRpcUrlMap={{ 1: [hardhat.url] }} />)
-      const connectWallet = await component.findByTestId('connect-wallet')
-      const toolbar = await component.findByTestId('toolbar')
-      expect(connectWallet.textContent).toBe('Connect wallet to swap')
-      expect(toolbar.textContent).toBe('Connect wallet to swap')
+  function itExpectsWidgetToBeDisabled(renderWidget: () => RenderResult, disabled = true) {
```
JFrankfurt, line 31: 🤘

src/hooks/web3/index.tsx
```diff
@@ -0,0 +1,139 @@
...
```
JFrankfurt, line 76: nit: prefer `.forEach` because `for ... of` is traumatically close to `for ... in`

src/hooks/web3/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const walletConnect = useMemo(
+    () =>
+      initializeWeb3ReactConnector(WalletConnect, { options: { rpc: jsonRpcUrlMap, qrcode: true }, defaultChainId }),
+    [defaultChainId, jsonRpcUrlMap]
+  ) // WC via built-in popup
+  const walletConnectQR = useMemo(
+    () =>
+      initializeWeb3ReactConnector(WalletConnect, { options: { rpc: jsonRpcUrlMap, qrcode: false }, defaultChainId }),
+    [jsonRpcUrlMap, defaultChainId]
+  ) // WC via tile QR code scan
```
JFrankfurt, line 124: maybe I'm reading this wrong, but it seems like the comments on these WalletConnect connector instances are reversed?

JFrankfurt (approved):

src/hooks/web3/index.tsx
zzmp: These are confusingly named, and the qrcode property just makes it more confusing.
I'm leaving for now, but in the future we can look into cleaning this up by only using a single connector.

PR #195, opened 2022-09-08
fix: switch chain behavior

zzmp:
- Switch chains without prompting on network provider
- Prompt chain switch if either token mismatches the current chain
- Continue to display fetched swaps when disabled


src/components/Swap/SwapActionButton/SwitchChainButton.tsx
```diff
@@ -1,10 +1,42 @@
...
+
+function useDesiredChainId({ inputChainId, outputChainId }: ChainIds): number | undefined {
+  const { chainId } = useWeb3React()
+  if (chainId === undefined) return
+
+  const desiredChainId = inputChainId ?? outputChainId
+  if (desiredChainId === chainId) return
+
+  return desiredChainId
+}
```
JFrankfurt, line 22: It's not clear from the implementation when I'd want to use this function, but it reads like a general purpose utility. Can you add a doc comment at the top explaining why it exists?

src/components/Swap/SwapActionButton/SwitchChainButton.tsx
```diff
@@ -1,10 +1,42 @@
...
+
+export function useSwitchChainId(chainIds: ChainIds): number | undefined {
```
JFrankfurt, line 24: I think this is could be named better

JFrankfurt (changes requested):
Fine w/ the functionality changes going in as they are, but I think this needs some work to make the code "self-documenting".

JFrankfurt (approved):

PR #200, opened 2022-09-12
feat: add swap widget skeleton for loading states

cartcrom:
Created a widget skeleton component, exported it, and added a cosmos fixture for it


src/cosmos/SwapSkeleton.fixture.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+function Fixture() {
+  const [width] = useValue('width', { defaultValue: 360 })
+
+  const [theme, setTheme] = useValue('theme', { defaultValue: defaultTheme })
+  const [darkMode] = useValue('darkMode', { defaultValue: false })
+  useEffect(() => setTheme((theme) => ({ ...theme, ...(darkMode ? darkTheme : lightTheme) })), [darkMode, setTheme])
+  const widget = <SwapWidgetSkeleton theme={theme} width={width} />
+
+  // If framed in a different origin, only display the SwapWidget, without any chrome.
```
zzmp, line 14: This is unnecessary - it's only used for the Widget so it can be iframed in the docs.

src/cosmos/SwapSkeleton.fixture.tsx
```diff
@@ -0,0 +1,25 @@
...
+  // This is done to faciliate iframing in the documentation (https://docs.uniswap.org).
+  if (!window.frameElement) return widget
+
+  return (
```
zzmp, line 18: I think you can just return the widget, without a row wrapping it, because there is no event feed.

zzmp (changes requested):
- The token selector button skeletons should align with the token selectors
- Should their be a skeleton for the reverse button?
- The skeleton does not respect the borderRadius property of the theme

@infredible should these be updated?

infredible:
Thanks for catching those things @zzmp 

I made some updates to the designs:
1. Added swap arrow button
2. Changed height and vertical positioning of some of the boxes
3. Added horizontal rule
4. Changed left padding of some boxes to align with horizontal rule
CleanShot 2022-09-13 at 18 55 32@2x

src/cosmos/SwapSkeleton.fixture.tsx
```diff
@@ -0,0 +1,15 @@
+import { darkTheme, defaultTheme, lightTheme, SwapWidgetSkeleton } from '@uniswap/widgets'
+import { useEffect } from 'react'
+import { useValue } from 'react-cosmos/fixture'
+
+function Fixture() {
+  const [width] = useValue('width', { defaultValue: 360 })
+  const [willDisableBranding] = useValue('willDisableBranding', { defaultValue: false })
```
zzmp, line 7: rm `willDisableBranding` - it is unused.

src/components/SwapWidgetSkeleton.tsx
```diff
@@ -0,0 +1,89 @@
...
+import ReverseButton from './Swap/ReverseButton'
+import { WidgetWrapper } from './Widget'
+
+const LoadingWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  justify-content: space-between;
+`
+const Blob = styled.div<{ height: string; width: string; radius?: number; darker?: boolean }>`
```
zzmp, line 16: `darker` should be renamed to `module` or `isModule`, as the theming is configurable.

src/components/SwapWidgetSkeleton.tsx
```diff
@@ -0,0 +1,89 @@
+import { StrictMode } from 'react'
```
zzmp, line 1: Can you move this to `src/components/Swap/Skeleton.tsx`?

src/components/SwapWidgetSkeleton.tsx
```diff
@@ -0,0 +1,89 @@
...
+`
+const WideColumn = styled(Column)`
+  width: 100%;
+`
+
+function FloatingDetails({ darker }: { darker?: boolean }) {
+  return (
+    <WideColumn gap={0.75}>
+      <Row>
+        <Blob height="16px" width="40px" darker={darker} />
```
zzmp, line 30: Widget expresses sizing in `em` - can you convert these all to `em` instead of `px` for consistency?

src/components/SwapWidgetSkeleton.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <Blob height="32px" width="117px" darker={darker} />
+      </Row>
+    </WideColumn>
+  )
+}
+
+function FloatingButton() {
+  return (
+    <WideColumn gap={0.875}>
+      <Blob height="1px" width="100%" darker />
```
zzmp, line 43: This should use `<Rule />`, like the actual SwapWidget.

src/components/SwapWidgetSkeleton.tsx
```diff
@@ -0,0 +1,89 @@
...
+export function SwapWidgetSkeleton({ theme, width }: SwapWidgetSkeletonProps) {
+  return (
+    <StrictMode>
+      <ThemeProvider theme={theme}>
+        <WidgetWrapper width={width}>
+          <LoadingWrapper>
+            <InputColumn>
+              <FloatingDetails />
+            </InputColumn>
+            <div>
```
zzmp, line 77: Does this div need to be here?

zzmp (changes requested):

src/components/SwapWidgetSkeleton.tsx
cartcrom: the ReverseButton seems to climb to the top of whatever container its in, so it needed to be a level lower than everything else here

zzmp (approved):

PR #202, opened 2022-09-13
feat: router perf

zzmp:
- Differentiates between TRADE and PRICE route queries
- Fixes SOR usage (to ensure it is lazy-loaded)
- Caches routers per provider

This mirrors many recent changes to the github.com/Uniswap/interface to improve routing performance and decrease JSON-RPC requests.


src/state/routing/slice.ts
```diff
@@ -91,7 +91,7 @@ export const routing = createApi({
               tokenOutAddress,
               tokenOutChainId,
               amount,
-              tradeType,
+              type: tradeType === TradeType.EXACT_INPUT ? 'exactIn' : 'exactOut',
```
vm, line 94: nit: should `exactIn` and `exactOut` be enum values?

vm (approved):

src/state/routing/slice.ts
zzmp: These are the values expected by the `quote` API.
The API should maybe allow our existing tradeType enum values, but it makes sense that these are strings.

PR #203, opened 2022-09-13
fix: action button border radius

zzmp:
- Subtracts the padding value from the borderRadius, so the ActionButton's inset appropriately nests.

image

### DIFF
--- a/src/components/ActionButton.tsx
+++ b/src/components/ActionButton.tsx
@@ -40,6 +40,7 @@ const actionCss = css`
   }
 
   ${StyledButton} {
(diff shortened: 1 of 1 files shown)


src/components/ActionButton.tsx
```diff
@@ -40,7 +40,7 @@ const actionCss = css`
   }
 
   ${StyledButton} {
-    border-radius: ${({ theme }) => theme.borderRadius}em;
+    border-radius: ${({ theme }) => theme.borderRadius - 0.25}em;
```
vm, line 43: nit: where is this .25 coming from?

vm (approved):

PR #206, opened 2022-09-14
fix: rm obsolete endpoint

zzmp:


aballerr (approved):

src/constants/jsonRpcEndpoints.ts
```diff
@@ -33,8 +33,6 @@ export const JSON_RPC_FALLBACK_ENDPOINTS: Record<SupportedChainId, string[]> = {
     'https://rpc.ankr.com/eth_goerli',
   ],
   [SupportedChainId.KOVAN]: [
-    // "Safe" URLs
-    'https://kovan.poa.network',
     // "Fallback" URLs
```
vm, line 36: so there are no safe urls anymore?

zzmp: there are no safe urls. someone should probably tell metamask.

zzmp: actually, the kovan infura endpoint is still considered a safe URL. there are no _public_ safe urls anymore.

PR #210, opened 2022-09-16
fix: rm v2/v3 swap routers

zzmp:


src/components/Swap/Price.tsx
```diff
@@ -1,16 +1,16 @@
...
+import { InterfaceTrade } from 'state/routing/types'
 import { ThemedText } from 'theme'
 import { formatCurrencyAmount, formatPrice } from 'utils/formatCurrencyAmount'
 import formatLocaleNumber from 'utils/formatLocaleNumber'
 
 import { TextButton } from '../Button'
 
 interface PriceProps {
-  trade: Trade<Currency, Currency, TradeType>
+  trade: InterfaceTrade
```
vm, line 13: need to use this in interface... i don't think i do

src/hooks/swap/useSendSwapTransaction.tsx
```diff
@@ -42,7 +34,7 @@ export default function useSendSwapTransaction(
   account: string | null | undefined,
   chainId: number | undefined,
   provider: JsonRpcProvider | undefined,
-  trade: AnyTrade | undefined, // trade to execute, required
+  trade: InterfaceTrade | undefined, // trade to execute, required
```
vm, line 37: if it's required why allow undefined?

zzmp: I think this is in contrast to `trade?: InterfaceTrade`, which would allow undefined or unspecified.
I can remove this comment - I think it's just confusing.

vm: yea kinda unnecessary

src/utils/prices.ts
```diff
@@ -34,49 +30,31 @@ export function getPriceImpactWarning(priceImpact?: Percent): 'warning' | 'error
...
-                ? // not currently possible given protocol check above, but not fatal
-                  FeeAmount.MEDIUM
-                : pool.fee
-            return currentFee.multiply(ONE_HUNDRED_PERCENT.subtract(new Fraction(fee, 1_000_000)))
-          }, ONE_HUNDRED_PERCENT)
-        )
+    const routeRealizedLPFeePercent = overallPercent.multiply(
+      ONE_HUNDRED_PERCENT.subtract(
+        swap.route.pools.reduce<Percent>((currentFee: Percent, pool): Percent => {
+          const fee = pool instanceof Pair ? /* V2 FeeAmount= */ FeeAmount.MEDIUM : pool.fee
```
vm, line 44: can you make ` pool instanceof Pair `a variable called `isV2`? then explain why that pool won't have a fee attribute

vm: ideally also name `FeeAmount` -> `V2FeeAmount`

src/utils/prices.ts
```diff
@@ -34,49 +30,31 @@ export function getPriceImpactWarning(priceImpact?: Percent): 'warning' | 'error
...
   }
 
   return new Percent(percent.numerator, percent.denominator)
 }
 
 // computes price breakdown for the trade
-export function computeRealizedLPFeeAmount(
-  trade?: Trade<Currency, Currency, TradeType> | null
-): CurrencyAmount<Currency> | undefined {
+export function computeRealizedLPFeeAmount(trade?: InterfaceTrade | null): CurrencyAmount<Currency> | undefined {
```
vm, line 57: the `| null` is a bit odd here, though if it's a pain to remove it's fine to ignore

src/components/Swap/RoutingDiagram/utils.ts
```diff
@@ -30,7 +27,7 @@ export function getTokenPath(trade: InterfaceTrade<Currency, Currency, TradeType
       const entry: RoutingDiagramEntry['path'][0] = [
         tokenIn,
         tokenOut,
-        nextPool instanceof Pair ? V2_DEFAULT_FEE_TIER : nextPool.fee,
+        nextPool instanceof Pool ? nextPool.fee : /* V2 FeeAmount= */ FeeAmount.MEDIUM,
```
vm, line 30: this logic is repeated in two diff places (as shown below) and requires some specific knowledge imo. maybe worth factoring out somehow

vm (approved):
with some comments

src/utils/prices.ts
zzmp: I'm not sure, because moving the check to a variable may lose the narrowing of the type to `Pair`. Made a new helper method `getFeeAmount` to get around this.

vm: works for me

PR #211, opened 2022-09-16
fix: return cached quotes immediately

zzmp:
- Returns cached quotes immediately by splitting the query calls into state and trigger.
- Also tightens up typings for the RTK integration.

This allows already-cached keys to be displayed in the UI immediately, while still debouncing any additional queries.

Note that using a single RTK hook (in `useRouterTrade`) does not allow returning the cached value immediately _and_ debouncing the request. The request _cannot_ be debounced _within_ RTK because aborting the request (within RTK) also evicts any existing cache entry.


src/hooks/routing/useRouterTrade.ts
```diff
@@ -94,40 +80,26 @@ export function useRouterTrade<TTradeType extends TradeType>(
...
 
+  // get USD gas cost of trade in active chains stablecoin amount
+  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(data?.gasUseEstimateUSD)
+
   return useMemo(() => {
     if (!currencyIn || !currencyOut) return INVALID_TRADE
 
     if (!trade && !isError) {
-      return { state: isDebouncing ? TradeState.SYNCING : TradeState.LOADING, trade: undefined }
+      return { state: TradeState.LOADING, trade: undefined }
```
justindomingue, line 90: should we extract this return value and the one for `TradeState.NO_ROUTE_FOUND` to the global scope to keep refs stable, similar to `INVALID_TRADE`?

justindomingue (dismissed):

src/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -44,7 +44,7 @@ async function getQuote(
   },
   router: AlphaRouter,
   routerConfig: Partial<AlphaRouterConfig>
-): Promise<{ data: GetQuoteResult; error?: unknown }> {
+): Promise<{ data: GetQuoteResult }> {
```
zzmp, line 47: This fn _throws_ errors, rather than including them in the return type.

src/hooks/routing/useRouterTrade.ts
```diff
@@ -40,94 +39,60 @@ export function useRouterTrade<TTradeType extends TradeType>(
...
-    // Price-fetching is informational and costly, so it's done less frequently.
+  // Get the cached state *immediately* to update the UI without sending a request - using useGetQuoteQueryState -
+  // but debounce the actual request - using useLazyGetQuoteQuery - to avoid flooding the router / JSON-RPC endpoints.
+  const { isError, data, currentData } = useGetQuoteQueryState(queryArgs)
+  const isValidBlock = useIsValidBlock(Number(data?.blockNumber))
+  const isSyncing = currentData !== data || !isValidBlock
+
+  const [trigger] = useLazyGetQuoteQuery({
+    // PRICE fetching is informational and costly, so it's done less frequently.
     pollingInterval: routerPreference === RouterPreference.PRICE ? ms`2m` : ms`15s`,
```
justindomingue, line 62: how does polling work with lazy queries? Does it start polling on the first `trigger` call?

src/hooks/routing/useRouterTrade.ts
```diff
@@ -40,94 +39,60 @@ export function useRouterTrade<TTradeType extends TradeType>(
...
   })
-
-  const quoteResult: GetQuoteResult | undefined = useIsValidBlock(Number(data?.blockNumber) || 0) ? data : undefined
+  useEffect(() => {
```
justindomingue, line 64: maybe would help readability/reusability to move this behavior to a `useTimeout` hook similar to https://github.com/Uniswap/interface/blob/f9fc506db4552af87f6f619fa0c1ec7b3dd08619/src/lib/hooks/useInterval.ts

src/hooks/routing/useRouterTrade.ts
justindomingue: > Accepts polling/re-fetching options to trigger automatic re-fetches when the corresponding criteria is met and the fetch has been manually called at least once

https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#uselazyquery

justindomingue: 👍 bit of a confusing behavior but it was my intuition

justindomingue (approved):

PR #212, opened 2022-09-20
fix: refetch stale data

zzmp:
Fixes triggering logic around the RTK query to refetch stale data, _without_ waiting an additional polling interval.

### DIFF
--- a/src/hooks/routing/useRouterTrade.ts
+++ b/src/hooks/routing/useRouterTrade.ts
@@ -54,19 +54,24 @@ export function useRouterTrade(
 
   // Get the cached state *immediately* to update the UI without sending a request - using useGetQuoteQueryState -
   // but debounce the actual request - using useLazyGetQuoteQuery - to avoid flooding the router / JSON-RPC endpoints.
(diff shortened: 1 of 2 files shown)


src/hooks/routing/useRouterTrade.ts
```diff
@@ -54,19 +54,24 @@ export function useRouterTrade(
 
   // Get the cached state *immediately* to update the UI without sending a request - using useGetQuoteQueryState -
   // but debounce the actual request - using useLazyGetQuoteQuery - to avoid flooding the router / JSON-RPC endpoints.
-  const { isError, data, currentData } = useGetQuoteQueryState(queryArgs)
+  const { isError, data, currentData, fulfilledTimeStamp } = useGetQuoteQueryState(queryArgs)
   const isValidBlock = useIsValidBlock(Number(data?.blockNumber))
   const isSyncing = currentData !== data || !isValidBlock
 
-  const [trigger] = useLazyGetQuoteQuery({
+  const pollingInterval = useMemo(
```
justindomingue, line 61: do you need the memo here? at build time it replaces `ms` with a number

justindomingue (approved):

PR #214, opened 2022-09-20
feat: propagate more info in swap info (wrap, error)

zzmp:
- includes wrap info as part of swap info
- includes error state as part of swap info (instead of recomputing it for the toolbar and the action button)
- conditionally includes trade, so that swap info includes whether a swap is a wrap or a trade


src/state/routing/types.ts
```diff
@@ -2,11 +2,20 @@ import { Trade } from '@uniswap/router-sdk'
...
   INVALID,
+  LOADING,
   NO_ROUTE_FOUND,
+  /**
+   * Only valid states should be listed after VALID.
+   * This simplifies validity checks to state >= TradeState.VALID
+   */
   VALID,
   SYNCING,
+  WRAP,
```
JFrankfurt, line 14: not sure if this is unhelpful, but `WRAP` doesn't seem like a trade **state** in the same way that these others are.

JFrankfurt (approved):
This works, but I feel a littttle weird making the `trade` entity possibly just `{ state: TradeState.WRAP }` when that object isn't actually a trade at all. It feels like we should have some other control interface for wrap transactions, since they're not trades at all.

src/state/routing/types.ts
zzmp: These are strongly coupled (ie a wrap is always an invalid trade), so I'm struggling with how to cleanly express this.
Because of the way the hooks are set up, we'll always execute the trade logic even if it is a wrap.

src/hooks/swap/useSwapInfo.tsx
```diff
@@ -35,39 +35,37 @@ interface SwapInfo {
...
   )
-  const hasAmounts = currencyIn && currencyOut && parsedAmount && !isWrap
   const trade = useRouterTrade(
     type,
-    hasAmounts ? parsedAmount : undefined,
-    hasAmounts ? (isExactInput(type) ? currencyOut : currencyIn) : undefined,
-    RouterPreference.TRADE,
+    parsedAmount,
+    isExactInput(type) ? outputCurrency : inputCurrency,
+    useIsWrap() ? RouterPreference.SKIP : RouterPreference.TRADE,
```
JFrankfurt, line 48: nit: hook usage in nested fn calls feels wrong to me. It hides the usage and is slightly more likely to cause future issues with conditional or duplicate calls.

src/hooks/swap/useSwapInfo.tsx
```diff
@@ -35,39 +35,37 @@ interface SwapInfo {
...
-    () => (isWrap || isExactInput(type) ? parsedAmount : trade.trade?.inputAmount),
-    [isWrap, parsedAmount, trade.trade?.inputAmount, type]
+  const inputAmount = useMemo(
+    () => (isExactInput(type) ? parsedAmount : trade.trade?.inputAmount),
+    [parsedAmount, trade.trade?.inputAmount, type]
   )
-  const amountOut = useMemo(
-    () => (isWrap || !isExactInput(type) ? parsedAmount : trade.trade?.outputAmount),
-    [isWrap, parsedAmount, trade.trade?.outputAmount, type]
+  const outputAmount = useMemo(
```
JFrankfurt, line 56: nit: amountIn/amountOut have protocol meanings that are lost by this change. The memo calls are specifically outputting the protocol's (exact)amount(in/out), so I think these names should stay the same.

src/hooks/swap/useWrap.ts
```diff
@@ -10,11 +10,6 @@ import tryParseCurrencyAmount from 'utils/tryParseCurrencyAmount'
```
JFrankfurt, line 10: any idea why there's no syntax highlighting here on github?

src/state/routing/types.ts
```diff
@@ -2,13 +2,21 @@ import { Trade } from '@uniswap/router-sdk'
 import { Currency, Token, TradeType } from '@uniswap/sdk-core'
 
 export enum TradeState {
-  LOADING,
   INVALID,
+  LOADING,
```
JFrankfurt, line 6: is loading def an invalid trade state? feels like this could cause bugs down the line

JFrankfurt (approved):
some notes re: future maintainability--up to you if you agree and want to address

PR #215, opened 2022-09-20
feat: desaturate unapproved input token

zzmp:
- Includes approval info as part of swap info
- Desaturates the input token if unapproved

DE-76


src/hooks/useAllowance.ts
```diff
@@ -0,0 +1,95 @@
...
+import { Currency, CurrencyAmount, Token } from '@uniswap/sdk-core'
+import { useWeb3React } from '@web3-react/core'
+import { useTokenContract } from 'hooks/useContract'
+import { useTokenAllowance } from 'hooks/useTokenAllowance'
+import { useCallback, useMemo } from 'react'
+import { calculateGasMargin } from 'utils/calculateGasMargin'
+
+export enum AllowanceState {
+  UNKNOWN,
+  NOT_ALLOWED,
```
JFrankfurt, line 12: allowances aren't binary, so I think this code might be a bit easier to follow if you use SUFFICIENT/INSUFFICIENT here instead of (NOT_)ALLOWED

src/hooks/swap/useApproval.ts
```diff
@@ -0,0 +1,103 @@
...
+      return ApprovalState.PENDING_ALLOWANCE
+    } else if (permitState === PermitState.LOADING) {
+      return ApprovalState.PENDING_SIGNATURE
+    } else if (allowance !== AllowanceState.NOT_ALLOWED || permitState === PermitState.SIGNED) {
+      return ApprovalState.APPROVED
+    } else if (gatherPermitSignature) {
+      return ApprovalState.REQUIRES_SIGNATURE
+    } else {
+      return ApprovalState.REQUIRES_ALLOWANCE
+    }
```
JFrankfurt, line 70: nit: (might not even be worth addressing) it's possible that a user has given sufficient approval to a token that implements permit

JFrankfurt: (so requesting a permit signature wouldn't be necessary)

zzmp: this should be considered APPROVED using the current PR logic

zzmp:
Replaced by #289.

PR #222, opened 2022-09-30
fix: attempt activate after switch

zzmp:
In the current state of Web3React, activating another chain will temporarily disable the connector, which may lead to a different connector appearing active. For this reason, switching through EIPs 3085/3326 is always attempted first, with a fallback to activating another chain.

### DIFF
--- a/src/hooks/useSwitchChain.ts
+++ b/src/hooks/useSwitchChain.ts
@@ -51,8 +51,12 @@ export default function useSwitchChain(): (chainId: SupportedChainId) => Promise
     async (chainId: SupportedChainId) => {
       try {
         try {
+          // A custom Connector may use a customProvider, in which case it should handle its own chain switching.
           if (!provider) throw new Error()
+
           await Promise.all([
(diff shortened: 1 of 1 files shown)


src/hooks/useSwitchChain.ts
```diff
@@ -56,13 +53,13 @@ export default function useSwitchChain(): (chainId: SupportedChainId) => Promise
...
   return useCallback(
-    (chainId: SupportedChainId) => {
-      if (connector instanceof WalletConnect || connector instanceof Network) {
-        return connector.activate(chainId)
+    async (chainId: SupportedChainId) => {
+      try {
+        if (!provider) throw new Error()
+        await switchChain(provider, chainId, urlMap[chainId])
+      } catch {
+        await connector.activate(chainId)
```
vm, line 61: this could also fail though?

zzmp: If this fails, I'd expect it to be an error.

vm (approved):

vm (approved):

PR #227, opened 2022-10-04
feat: replace no token logo image

cmcewen:
Screen Shot 2022-10-04 at 1 29 06 PM
Screen Shot 2022-10-04 at 1 50 01 PM


src/cosmos/Swap.fixture.tsx
```diff
@@ -19,6 +19,21 @@ import EventFeed, { Event } from './EventFeed'
...
+  decimals: 18,
+  symbol: 'HDRN',
+  name: 'Hedron',
+  address: '0x3819f64f282bf135d62168C1e513280dAF905e06',
+}
+
+const newTokenList = (tokens as TokenInfo[]).concat([TOKEN_WITH_NO_LOGO])
+const mainnetTokens = tokens.filter((token) => token.chainId === SupportedChainId.MAINNET)
+const tokenLists: Record<string, TokenInfo[]> = {
+  Default: newTokenList,
```
zzmp, line 33: The Default list should remain all `tokens`.
Can you make a new list, `Hedron`, or `Logoless`, and put this there?

src/components/TokenImg.tsx
```diff
@@ -39,7 +45,13 @@ function TokenImg({ token, ...rest }: TokenImgProps) {
     [src]
   )
 
-  if (!src) return <MissingTokenImg src={missingTokenSrc} alt={alt} color="secondary" {...rest} />
+  if (!src) {
+    return (
+      <MissingTokenImg>
```
zzmp, line 50: This should continue to pass `{...rest}`, as these are the DOM attributes being passed down.

src/components/TokenImg.tsx
```diff
@@ -1,15 +1,21 @@
...
 const badSrcs = new Set<string>()
 
-const MissingTokenImg = styled.img`
-  height: 1em;
-  width: 1em;
+const MissingTokenImg = styled.div<{ fontSize?: number; circleSize?: number }>`
+  background-color: ${({ theme }) => theme.container};
+  border-radius: 100%;
+  color: ${({ theme }) => theme.primary};
+  font-size: ${({ fontSize }) => fontSize || 9}px;
```
zzmp, line 13: The font specification should be done by nesting a proper type from `src/theme/type.tsx` under the `MissingTokenImg`.
Also, this is no longer an img, it is a div, so it can just be named `MissingToken` (or be made a proper functional element).

src/components/TokenImg.tsx
```diff
@@ -1,15 +1,21 @@
...
+  font-weight: 500;
+  height: ${({ circleSize }) => circleSize || 24}px;
```
zzmp, line 15: For consistency, widgets uses `em` instead of `px`.

zzmp (changes requested):

src/components/TokenImg.tsx
cmcewen: i'm not sure that specifying a new font makes sense in this case since that font is imo very unlikely to be used outside of the context of this specific image? and semantically this element is playing the role of an image - it's an implementation detail to me that it's a div. what do you mean proper functional element?

zzmp (approved):

PR #228, opened 2022-10-04
fix: smart-order-router integrations

zzmp:
- [X] Use a static V2SubgraphProvider for non-mainnet chains to prevent querying non-existent subgraph data
- [x] Use a custom V3QuoteProvider (or OnChainQuoteProvider) for Polygon to prevent hitting gas limits
- [x] Cache 404 NO_ROUTE_FOUND responses as valid responses to prevent retries
- [x] Update configuration to prevent infinite retries and (at some point) display an error in the toolbar


zzmp:
The content of this PR was split into smaller PRs, and has been merged.

PR #233, opened 2022-10-05
fix: update router parameters for esoteric chains

zzmp:
- Uses a static v2SubgraphProvider for non-mainnet chains, which do not have ipns metadata for v2
- Tunes the onChainQuoteProvider for polygon, which was running out of gas


src/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -26,7 +33,47 @@ function getRouter(chainId: ChainId, provider: BaseProvider): AlphaRouter {
   const cached = routers[chainId]
   if (cached) return cached
 
-  const router = new AlphaRouter({ chainId, provider })
+  // V2 is unsupported for chains other than mainnet.
+  // TODO(zzmp): Upstream to @uniswap/smart-order-router.
```
cmcewen, line 37: in the upstream, i'd vote to have an enum list of supported v2/v3 deployments on different chains to make it more clear

cmcewen (approved):

PR #234, opened 2022-10-05
fix: router trade performance

zzmp:
- Caches NO_ROUTEs as a valid response (to avoid retries)
- Prevents refetching if a fetch is already in progress


src/hooks/routing/useRouterTrade.ts
```diff
@@ -41,50 +42,49 @@ export function useRouterTrade(
...
-  const pollingInterval = routerPreference === RouterPreference.PRICE ? ms`2m` : ms`15s`
+  // An already-fetched value should be refetched if it is older than the pollingInterval.
+  // Without explicit refetch, it would not be refetched until another pollingInterval has elapsed.
   const [trigger] = useLazyGetQuoteQuery({ pollingInterval })
   const request = useCallback(() => {
     const { refetch } = trigger(queryArgs, /*preferCacheValue=*/ true)
-    // An already-fetched value should be refetched if it is older than the pollingInterval.
-    // Without explicit refetch, it would not be refetched until another pollingInterval has elapsed.
-    if (fulfilledTimeStamp && Date.now() - fulfilledTimeStamp > pollingInterval) {
+    if (!isFetching && fulfilledTimeStamp && Date.now() - fulfilledTimeStamp > pollingInterval) {
```
zzmp, line 57: Without `!isFetching`, this was refetching multiple times every polling period.

src/hooks/routing/useRouterTrade.ts
```diff
@@ -41,50 +42,49 @@ export function useRouterTrade(
...
+  }, [amountSpecified?.currency, otherCurrency, result, tradeType])
+  const isValidBlock = useIsValidBlock(Number(result?.blockNumber))
+  const isSyncing = currentData !== data || !isValidBlock
+  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(result?.gasUseEstimateUSD)
 
   return useMemo(() => {
-    if (!currencyIn || !currencyOut) return TRADE_INVALID
-    if (!trade && !isError) return TRADE_LOADING
+    if (queryArgs === skipToken) return TRADE_INVALID
+    if (data === NO_ROUTE) return TRADE_NOT_FOUND
```
zzmp, line 83: This would otherwise be treated as `TRADE_LOADING`, so I created a new return value (`NO_ROUTE`) to signify this state. `NO_ROUTE` is a string so that redux can serialize it - redux cautions against using a Symbol.

src/hooks/routing/useRouterTrade.ts
```diff
@@ -41,50 +42,49 @@ export function useRouterTrade(
...
       refetch()
     }
-  }, [fulfilledTimeStamp, pollingInterval, queryArgs, trigger])
+  }, [fulfilledTimeStamp, isFetching, pollingInterval, queryArgs, trigger])
   useTimeout(request, 200)
 
-  const [currencyIn, currencyOut] = isExactInput(tradeType)
-    ? [amountSpecified?.currency, otherCurrency]
-    : [otherCurrency, amountSpecified?.currency]
+  const result = useMemo(() => (typeof data === 'object' ? data : undefined), [data])
```
cmcewen, line 63: is this check expensive enough that it needs to be memoized? don't quite understand this

src/hooks/routing/useRouterTrade.ts
```diff
@@ -41,50 +42,49 @@ export function useRouterTrade(
...
     try {
       return transformRoutesToTrade(routes, tradeType)
     } catch (e: unknown) {
       console.debug('transformRoutesToTrade failed: ', e)
       return
     }
-  }, [currencyIn, currencyOut, data, tradeType])
-  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(data?.gasUseEstimateUSD)
+  }, [amountSpecified?.currency, otherCurrency, result, tradeType])
+  const isValidBlock = useIsValidBlock(Number(result?.blockNumber))
```
cmcewen, line 77: will this cast hurt memoizing? i think it'll do number compares by value instead of reference but just checking

src/state/routing/slice.ts
```diff
@@ -43,22 +43,32 @@ export const routing = createApi({
...
+              // NO_ROUTE should be treated as a valid response to prevent retries.
+              if (typeof data === 'object' && data.errorCode === 'NO_ROUTE') {
+                return { data: NO_ROUTE as GetQuoteResult }
+              }
+
+              throw data
             }
 
             const data: GetQuoteResult = await response.json()
             return { data }
```
cmcewen, line 60: i don't love that this is just called "data"

cmcewen (approved):
if you're okay with the memoization i'm okay with merging

src/hooks/routing/useRouterTrade.ts
zzmp: probably not, I'll remove it.
I consider any computation as needing memoization, to reduce computation time; or any logic which would return a referentially unstable value, to prevent unnecessary rerenders.

src/hooks/routing/useRouterTrade.ts
zzmp: yes, primitive comparisons (number, string, boolean) should always be by value.

src/state/routing/slice.ts
zzmp: very fair - I've renamed it to `quote` in both cases where it is returned as valid data.
the name `data` is required by the typing of redux.

PR #235, opened 2022-10-05
fix: remove For in Swap component

vm:
Screen Shot 2022-10-05 at 6 32 23 PM
Screen Shot 2022-10-05 at 6 32 35 PM

### DIFF
--- a/src/components/Swap/Output.tsx
+++ b/src/components/Swap/Output.tsx
@@ -1,4 +1,3 @@
-import { Trans } from '@lingui/macro'
 import { useLingui } from '@lingui/react'
 import BrandedFooter from 'components/BrandedFooter'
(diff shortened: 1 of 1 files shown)


cmcewen (approved):
looks like a bit too much padding up top in the lower box now but if design is happy i'm happy

vm:
@zzmp when you see this, please lmk if i need to change anything about the widget's fixed height stuff based on this change

zzmp:
> @zzmp when you see this, please lmk if i need to change anything about the widget's fixed height stuff based on this change

You should test and fix any expando texts (error and details), as these have height hard coded. You'll probably want to do it in tandem with updating height when hiding branding. I think the height should be static, even if it's variable based on inputs (ie it can change but it's static given certain props).

PR #240, opened 2022-10-06
feat: move max ui

zzmp:
Moves max UI from the token button to the balance row.
Updating UX (ie setting focus after updating amount) will be done in a follow-up PR.

WEB-790

### DIFF
--- a/src/components/Swap/Input.tsx
+++ b/src/components/Swap/Input.tsx
@@ -128,7 +128,7 @@ export default function Input({ disabled, focused }: InputProps) {
             {balance && (
               <Row gap={0.5}>
                 <Balance color={insufficientBalance ? 'error' : focused ? 'secondary' : 'hint'}>
-                  Balance: <span>{formatCurrencyAmount(balance, 4, i18n.locale)}</span>
+                  <Trans>Balance:</Trans> <span>{formatCurrencyAmount(balance, 4, i18n.locale)}</span>
                 </Balance>
                 {max && (
                   <TextButton onClick={onClickMax}>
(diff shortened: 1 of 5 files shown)


src/components/Swap/Input.tsx
```diff
@@ -80,26 +83,33 @@ export default function Input({ disabled, focused }: InputProps) {
...
+  const max = useMemo(() => {
+    // account for gas needed if using max on native token
+    const max = maxAmountSpend(balance)
+    if (!max || !balance) return
+    if (max.equalTo(0) || balance.lessThan(max)) return
+    if (inputCurrencyAmount && max.equalTo(inputCurrencyAmount)) return
+    return max.toExact()
+  }, [balance, inputCurrencyAmount])
+  const onClickMax = useCallback(() => {
+    invariant(max)
```
cmcewen, line 110: idk if we use this elsewhere in the codebase but i had to look up what this does because it wasn't immediately obvious to me this was just a check for a truthy value

src/components/Swap/Input.tsx
```diff
@@ -117,9 +126,18 @@ export default function Input({ disabled, focused }: InputProps) {
           <Row>
             <USDC isLoading={isRouteLoading}>{usdc ? `$${formatCurrencyAmount(usdc, 6, 'en', 2)}` : ''}</USDC>
             {balance && (
-              <Balance color={insufficientBalance ? 'error' : focused ? 'secondary' : 'hint'}>
-                Balance: <span>{formatCurrencyAmount(balance, 4, i18n.locale)}</span>
-              </Balance>
+              <Row gap={0.5}>
+                <Balance color={insufficientBalance ? 'error' : focused ? 'secondary' : 'hint'}>
+                  Balance: <span>{formatCurrencyAmount(balance, 4, i18n.locale)}</span>
```
cmcewen, line 131: does this need to be translated?

cmcewen (approved):

PR #241, opened 2022-10-06
feat: focus after setting max

zzmp:
Focuses the input amount after setting it to max, and scrolls to the beginning so the set value is apparent to the user.


src/components/Swap/Input.tsx
```diff
@@ -75,6 +75,7 @@ export default function Input({ disabled, focused }: InputProps) {
   const [inputAmount, updateInputAmount] = useSwapAmount(Field.INPUT)
   const [inputCurrency, updateInputCurrency] = useSwapCurrency(Field.INPUT)
   const inputCurrencyAmount = useSwapCurrencyAmount(Field.INPUT)
+  const [input, setInput] = useState<TokenInputHandle | null>(null)
```
cmcewen, line 78: you'd think react would have a nicer way to do this

cmcewen (approved):

src/components/Swap/Input.tsx
zzmp: It should be useRef but there are typing issues with the imperative handle afaict. Maybe it's fixed in v18 🤷 but the widget is v17-compatible.

PR #245, opened 2022-10-10
feat: batch phase0 style updates

JFrankfurt:
https://uniswaplabs.atlassian.net/browse/WEB-1200

- allow dynamic height (wallet connection UI was conflicting with this, so updated some `height` values to make it more flexible
image

- remove double info icon in swap summary `Expando`
image

- move output wrapper color container to no longer include the footer or actionbutton
image

- updates some theme color values to match phase 0 branding


src/components/ActionButton.tsx
```diff
@@ -50,6 +50,7 @@ const actionCss = css`
 export const Overlay = styled(Row)<{ hasAction: boolean }>`
   border-radius: ${({ theme }) => theme.borderRadius}em;
   flex-direction: row-reverse;
+  margin-top: 12px;
```
zzmp, line 53: The button itself should not have a margin-top. It should be a part of the containing element (eg as a `gap`).

src/components/BrandedFooter.tsx
```diff
@@ -10,6 +10,7 @@ import ExternalLink from './ExternalLink'
 const UniswapA = styled(ExternalLink)`
   color: ${({ theme }) => theme.secondary};
   cursor: pointer;
+  margin-top: 12px;
```
zzmp, line 13: The anchor itself should not have a margin-top. It should be a part of the containing element (eg as a gap), either of the Row of the BrandedFooter, or of the Footer's containing element.

src/components/ConnectWallet/ConnectWalletDialog.tsx
```diff
@@ -15,8 +15,10 @@ import { WalletConnectQR } from 'utils/WalletConnect'
 const NO_WALLET_HELP_CENTER_URL = 'https://help.uniswap.org/en/articles/5391585-how-to-get-a-wallet'
 
 const Body = styled(Column)`
+  display: grid;
+  gap: 12px;
```
zzmp, line 19: This gap looks much larger than the Row's horizontal gap. Can they be made equal?

src/components/Expando.tsx
```diff
@@ -66,7 +75,7 @@ export default function Expando({ title, open, onExpand, height, children, ...re
         <Rule />
         <TitleRow>
           <TitleHeader>
-            <InfoIcon style={{ marginRight: '5px' }} color="secondary" />
+            {hideIcon === true ? null : <InfoIcon style={{ marginRight: '5px' }} color="secondary" />}
```
zzmp, line 78: I think this icon should only ever be passed as a part of the title.
Can `hideIcon` and `InfoIcon` be removed, and `InfoIcon` just added explicitly to those use cases?

src/components/Swap/index.tsx
```diff
@@ -73,10 +74,9 @@ export default function Swap(props: SwapProps) {
           <SwapInfoProvider routerUrl={props.routerUrl}>
             <Input disabled={isDisabled} focused={focused} />
             <ReverseButton disabled={isDisabled} />
-            <Output disabled={isDisabled} focused={focused}>
-              <Toolbar />
-              <SwapActionButton disabled={isDisabled} />
-            </Output>
+            <Output disabled={isDisabled} focused={focused} />
```
zzmp, line 77: What happened to the Toolbar?

src/components/Swap/Toolbar/index.tsx
```diff
@@ -13,7 +13,8 @@ import Row from '../../Row'
 import * as Caption from './Caption'
 
 const ToolbarRow = styled(Row)`
-  padding: 0.5em 0;
+  border-top: 1px solid ${({ theme }) => theme.container};
+  padding: 14px 16px 0.5em 16px;
```
zzmp, line 17: Why are you mixing px and em?

src/components/Swap/TokenInput.tsx
```diff
@@ -71,7 +71,7 @@ export const TokenInput = forwardRef<TokenInputHandle, PropsWithChildren<TokenIn
   useImperativeHandle(ref, () => ({ focus }), [focus])
 
   return (
-    <Column gap={0.25}>
+    <Column gap={0.25} {...rest}>
```
zzmp, line 74: I love that you propagate all the DOM props like this ❤️

src/components/Swap/Summary/index.tsx
```diff
@@ -43,7 +43,7 @@ const Body = styled(Column)`
 function Subhead({ impact, slippage }: { impact?: PriceImpact; slippage: Slippage }) {
   return (
     <Row gap={0.5}>
-      {impact?.warning || slippage.warning ? (
+      {Boolean(impact?.warning || slippage.warning) ? (
```
zzmp, line 46: This would be more readable as a separate variable:

    const hasWarning = Boolean(impact?.warning || slippage.warning)

src/components/Swap/Output.tsx
```diff
@@ -74,37 +67,37 @@ export default function Output({ disabled, focused, children }: PropsWithChildre
...
+                {balance && (
+                  <Balance color={focused ? 'secondary' : 'hint'}>
+                    Balance: <span>{formatCurrencyAmount(balance, 4, i18n.locale)}</span>
+                  </Balance>
+                )}
+              </Row>
+            </ThemedText.Body2>
+          </TokenInput>
+        </MarginWrapper>
+        <Toolbar />
```
zzmp, line 100: I prefer to keep this as a separate element in Swap/index.tsx, so that Input.tsx can more closely mirror Output.tsx. Is there a reason to prefer it as part of Output.tsx?

zzmp (changes requested):

src/components/Swap/Toolbar/index.tsx
JFrankfurt: mistake

src/components/Swap/Output.tsx
JFrankfurt: Added Toolbar here to make it easy to accomplish the background color/border-radius/border-color arrangement necessary here:
image

It's not technically required, but the diff is much smaller if we do it this way.

src/components/Swap/index.tsx
JFrankfurt: image

src/components/ActionButton.tsx
JFrankfurt: image
Where would you put the 12px here? (This currently puts in the `Row` wrapper for `SwapActionButton`)

src/components/Swap/Output.tsx
JFrankfurt: (No longer doing it this way. Bit the bullet and put it up a level at `Swap`.)

src/components/BrandedFooter.tsx
```diff
@@ -25,10 +25,13 @@ const UniswapA = styled(ExternalLink)`
     transform: rotate(-5deg);
   }
 `
+const Wrapper = styled(Row)`
+  margin-top: 12px;
+`
 
 export default memo(function BrandedFooter() {
```
zzmp, line 32: You can remove `memo` - it does nothing here afaik.

src/components/Swap/index.tsx
```diff
@@ -61,6 +62,7 @@ export default function Swap(props: SwapProps) {
   const isDisabled = !(isActive && onSupportedNetwork)
 
   const focused = useHasFocus(wrapper)
+  const disableBranding = useBrandingSetting()
```
zzmp, line 65: nit: `useBrandingSetting` should be renamed so it's output is more obvious. I'd be in favor of just returning `BrandingSettings`, and then destructuring.

zzmp (approved):
two nits but lgtm

PR #246, opened 2022-10-10
feat: replace LP fee with Network fee

zzmp:
Replaces LP fee with Network fee.

image

WEB-1469


src/components/Swap/Summary/Details.tsx
```diff
@@ -68,9 +68,8 @@ export default function Details({ trade, slippage, impact }: DetailsProps) {
       rows.push([t`Price impact`, impact.toString(), impact.warning])
     }
 
-    if (lpFeeAmount) {
-      const parsedLpFee = formatCurrencyAmount(lpFeeAmount, 6, i18n.locale)
-      rows.push([t`Liquidity provider fee`, `${parsedLpFee} ${inputCurrency.symbol || currencyId(inputCurrency)}`])
+    if (gasUseEstimateUSD) {
+      rows.push([t`Network fee`, `~$${gasUseEstimateUSD.toFixed(2)}`])
```
cmcewen, line 72: this would be a candidate for a format function

src/components/Swap/Summary/Details.tsx
```diff
@@ -86,14 +85,15 @@ export default function Details({ trade, slippage, impact }: DetailsProps) {
...
     i18n.locale,
     impact,
-    inputCurrency,
+    inputCurrency.symbol,
     integrator,
-    lpFeeAmount,
     outputAmount,
     outputCurrency,
-    slippage,
+    slippage.allowed,
```
cmcewen, line 95: is this actually desirable to have two comparisons vs one? not to be pedantic but i had to read a bit to figure out these changes.

cmcewen (approved):

src/components/Swap/Summary/Details.tsx
zzmp: This is done through linting, so it is desirable in that sense.

If you can guarantee that a change to `slippage` will always occur if `slippage.allowed || slippage.warning` changes, then it is fine to just depend on `slippage`, but nothing in code (aside from our own conventions) actually guarantees this, so depending on them separately is technically correct and safer.

src/components/Swap/Summary/Details.tsx
zzmp: Completely agree - AFAIK this is a follow-up for widget, generally. Right now https://github.com/uniswap/interface has a `formatDollar` function, but https://github.com/uniswap/widgets does not yet.

PR #247, opened 2022-10-10
chore: change font style of error

cmcewen:
WEB-999
before:
Screen Shot 2022-10-10 at 3 36 16 PM

after:
Screen Shot 2022-10-10 at 3 39 04 PM


lynnshaoyu (approved):

zzmp:
Where did you pull the font choice from?
The element that you changed will nearly always be code (ie a stack trace, or an error generated directly from code), so I'm afraid it may be more appropriate to leave that as code.

PR #249, opened 2022-10-10
feat: chain errors

zzmp:
- Propagates ChainError's through SwapInfo, so that checking for chain errors (which is error-prone) is not duplicated throughout the Widget.
- Refactors Toolbar and SwapActionButton to use ChainError to display errors:
  - Resolves WEB-1446: when "Switch network" is visible, disable all other widget click targets
  - Resolves WEB-1445: Clear toolbar when "Switch network" CTA is in Action Button

### DIFF
--- a/src/components/Swap/Output.tsx
+++ b/src/components/Swap/Output.tsx
@@ -7,7 +7,7 @@ import useCurrencyColor from 'hooks/useCurrencyColor'
 import { useBrandingSetting } from 'hooks/useSyncBrandingSetting'
 import { atom } from 'jotai'
 import { useAtomValue } from 'jotai/utils'
-import { PropsWithChildren } from 'react'
+import { ReactNode } from 'react'
 import { TradeState } from 'state/routing/types'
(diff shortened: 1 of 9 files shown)


src/components/Swap/Output.tsx
```diff
@@ -41,21 +41,23 @@ const Footer = styled(Row)`
...
     trade: { state: tradeState },
     impact,
   } = useSwapInfo()
 
   const [swapOutputAmount, updateSwapOutputAmount] = useSwapAmount(Field.OUTPUT)
   const [swapOutputCurrency, updateSwapOutputCurrency] = useSwapCurrency(Field.OUTPUT)
 
-  const isRouteLoading = disabled || tradeState === TradeState.LOADING
+  const isDisabled = error !== undefined
+  const isRouteLoading = isDisabled || tradeState === TradeState.LOADING
```
cmcewen, line 60: these few lines are duplicated between input + output above with just the Field being different

cmcewen (approved):

PR #250, opened 2022-10-10
feat: Token select chevron, i/o spacing, input background

JFrankfurt:
depends on #245 

image

image


src/components/Swap/Input.tsx
```diff
@@ -142,7 +153,6 @@ export default function Input() {
           </Row>
         </ThemedText.Body2>
       </TokenInput>
-      <Row />
```
JFrankfurt, line 145: @zzmp a little nervous about removing this because I can't figure out why it was here in the first place

JFrankfurt: (except if as a shim to make `gap` on the parent do something?)

src/components/Swap/Input.tsx
```diff
@@ -33,10 +33,21 @@ export const Balance = styled(ThemedText.Body2)`
...
-const InputColumn = styled(Column)<{ approved?: boolean }>`
-  margin: 0.75em;
+export const InputColumn = styled(Column)<{ approved?: boolean; hasColor?: boolean | null }>`
+  background-color: ${({ theme }) => theme.module};
+  border-radius: ${({ theme }) => theme.borderRadius - 0.25}em;
+  padding: 20px 0;
+  margin-bottom: 4px;
   position: relative;
 
+  // Set transitions to reduce color flashes when switching color/token.
```
cmcewen, line 43: i'm assuming this comment will get removed right? idk how the styled components pipeline works

src/components/Swap/Input.tsx
```diff
@@ -33,10 +33,21 @@ export const Balance = styled(ThemedText.Body2)`
...
+  // When color loads, transition the background so that it transitions from the empty or last state, but not _to_ the empty state.
+  transition: ${({ hasColor }) => (hasColor ? 'background-color 0.25s ease-out' : undefined)};
```
cmcewen, line 45: this looks specific to this component but always nice to parameterize these animations

src/components/Swap/Input.tsx
cmcewen: how could you confirm this either way?

src/components/Swap/Input.tsx
cmcewen: i see that this is a copy paste from below so nvm

src/components/Swap/ReverseButton.tsx
```diff
@@ -10,7 +10,7 @@ import Row from '../Row'
 const ReverseRow = styled(Row)`
   left: 50%;
   position: absolute;
-  transform: translate(-50%, -50%);
+  transform: translate(-50%, -60%);
```
zzmp, line 13: Why is this necessary? Will this not off center it?

cmcewen, line 13: negative numbers in CSS are always sus - why 60% now?

cmcewen (approved):
i am okay to merge this if you can reduce your nervousness about removing the row

src/components/Swap/Input.tsx
zzmp: It's definitely a spacing shim

src/components/Swap/ReverseButton.tsx
zzmp: This is necessary because of the new paddings/margins for the Input/Output.
However, in the initial screen, the Token Select touches the ReverseButton and it doesn't look great. Could we consult with design on that and see if it's intended?

JFrankfurt: with the modified asymmetric padding values we're using now this is necessary to keep the button centered

JFrankfurt: - `absolute` positions the element on the left-hand side of the screen w/ it's top left (`0`) pixel at the element's natural start point
- `left: 50%` puts it halfway across the screen
- `translate(-50%, __)` moves the element half its own width to the left of where it would fall given the prior `left: 50%` rule
- `translate(__, -60%)` moves the element up 60% of its own height to center it. (This accounts for the slight asymetry introduced by the new bottom margin btwn the in/output elements that allows the `container` color value to separate them.)

here's the now off-center look of `translate(-50%, -50%)`:
image

cc @zzmp

JFrankfurt:
updating w/ input from @infredible 
image

PR #253, opened 2022-10-11
feat: add hover states

cmcewen:
😬 needs #250 to land


zzmp (approved):

src/components/Swap/Input.tsx
```diff
@@ -52,6 +52,30 @@ export const InputColumn = styled(Column)<{ approved?: boolean; hasColor?: boole
...
+
+    position: absolute;
+    top: 0;
+    left: 0;
+
+    width: 100%;
+    height: 100%;
+    pointer-events: none;
+    content: '';
+    border: 1px solid ${({ theme }) => theme.module};
```
JFrankfurt, line 69: add `transition: 125ms ease border-color`

src/components/Swap/Input.tsx
```diff
@@ -52,6 +52,30 @@ export const InputColumn = styled(Column)<{ approved?: boolean; hasColor?: boole
...
+  }
+
+  &:hover:before {
+    border-color: ${({ theme }) => theme.interactive};
+  }
+
+  &:focus-within:before {
+    border-color: ${({ theme }) => theme.outline};
```
JFrankfurt, line 77: thoughts on using `outline` here instead of border to better align w/ standard `focus` behavior?

zzmp: similarly, why does this need to be on a pseudo-element?

src/components/Swap/Input.tsx
zzmp, line 69: Why can't you just add `border` to the `InputColumn`?

zzmp (changes requested):

src/components/Swap/Input.tsx
cmcewen: adding a border changes the size of the element so the UI jumps a bit, whereas the pseudo overlays. i had the same thought 😄

src/components/Swap/Input.tsx
cmcewen: i'm guessing due to this? `Before Chrome 94, outline does not follow the shape of border-radius` https://caniuse.com/?search=outline. idk what our browser support targets are

zzmp:
The on hover state should not occur if the InputColumn is disabled.

src/components/Swap/Input.tsx
```diff
@@ -52,6 +52,35 @@ export const InputColumn = styled(Column)<{ approved?: boolean; hasColor?: boole
...
+
+    width: 100%;
+    height: 100%;
+    pointer-events: none;
+    content: '';
+    border: 1px solid ${({ theme }) => theme.module};
+    transition: 125ms ease border-color;
+  }
+
+  ${({ theme, disableHover }) =>
```
zzmp, line 73: nit: these two could be put into one conditional.
another nit: I find this easier to read if the conditional is within border-color, and not over a larger css block.

zzmp (approved):
nits, but lgtm

src/components/Swap/Input.tsx
```diff
@@ -122,7 +151,7 @@ export default function Input() {
   }, [input, max, updateInputAmount])
 
   return (
-    <InputColumn gap={0.5} approved={mockApproved}>
+    <InputColumn gap={0.5} approved={mockApproved} disableHover={isDisabled}>
```
zzmp, line 154: This will need to be replicated in Output.tsx.
Also, the logic in both cases should be `disableHover={isDisabled || !swapXputCurrency}` - in other words, if no currency is selected, they should not be selectable.

zzmp (changes requested):

PR #258, opened 2022-10-11
feat: update skeleton

cmcewen:
Screen Shot 2022-10-11 at 10 51 09 AM
Screen Shot 2022-10-11 at 10 51 40 AM


zzmp (changes requested):
- The heights do not match up with the loaded version
- The horizontal rule between Output/Toolbar should be a different color
- There should be space for the BrandedFooter

PR #259, opened 2022-10-11
fix: design nits

JFrankfurt:
widget:
image

web:
image


zzmp (changes requested):
These should not be using custom font-weights. Is the design system out of date?

JFrankfurt:
@zzmp updating `BaseButton` would increase font-weight in a lot of places that I'm not sure about changing (like the <address> field in the navbar)

JFrankfurt:
image

new reverse button

yarn.lock
```diff
@@ -5178,9 +5178,9 @@ camelize@^1.0.0:
   integrity sha1-FkpUg+Yw+kMh5a8HAg5TGDGyYJs=
```
zzmp, line 5178: This should be updated in a separate PR.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -28,7 +28,7 @@ function Caption({ icon: Icon = AlertTriangle, caption }: CaptionProps) {
   return (
     <>
       <Icon color="secondary" />
-      <ThemedText.Caption>{caption}</ThemedText.Caption>
+      <ThemedText.Caption color="secondary">{caption}</ThemedText.Caption>
```
zzmp, line 31: Just to double-check: is a valid trade's toolbar rendered in the correct colors?

src/components/Input.tsx
```diff
@@ -43,8 +44,11 @@ const Input = styled.input`
 
   :disabled {
     // Overrides WebKit's override of input:disabled color.
-    -webkit-text-fill-color: ${({ theme }) => transparentize(1 - loadingOpacity, theme.primary)};
-    color: ${({ theme }) => transparentize(1 - loadingOpacity, theme.primary)};
+    -webkit-text-fill-color: ${({ theme }) => inputValueColorTransparentize(theme.primary)};
```
zzmp, line 47: Can you use opacify directly in these instances, instead of creating a helper function in L6? I think it is small enough that inlining is more readable.

Also, prefer opacify if it exists over transparentize, to avoid needing add'l math.

zzmp: I'd prefer the changes to this file be backed out - I think it makes this harder to read.

src/components/Swap/ReverseButton.tsx
```diff
@@ -1,72 +1,40 @@
...
-  height: 2.5em;
-  position: relative;
-  width: 2.5em;
-
-  div {
-    transform: rotate(${({ turns }) => turns / 2}turn);
-    transition: transform 0.25s ease-in-out;
-    will-change: transform;
+  :hover {
+    background-color: ${({ theme }) => onHoverColorShift(theme.module)};
```
zzmp, line 26: inline this. splitting into a helper function splits the logic and makes it harder to follow.

zzmp (changes requested):
See comments

src/components/Input.tsx
JFrankfurt: it exists, but it doesn't work quite the way you're thinking. 

transparentize is the tool for reduction, opacify for increases.

yarn.lock
JFrankfurt: I don't think caniuse is ever really important enough to merit its own pr -- too much process for its value

JFrankfurt: prefer people bump it whenever they notice the message

zzmp (approved):

PR #264, opened 2022-10-11
fix: center button and toolbar text with relevant icons

JFrankfurt:
## before
image

## after
image

### DIFF
--- a/src/components/Swap/Toolbar/Caption.tsx
+++ b/src/components/Swap/Toolbar/Caption.tsx
@@ -27,7 +27,7 @@ interface CaptionProps {
 function Caption({ icon: Icon = AlertTriangle, caption }: CaptionProps) {
   return (
     <>
-      <Icon color="secondary" size={16.67 / 14} />
+      <Icon color="secondary" />
       <ThemedText.Caption lineHeight="14px">{caption}</ThemedText.Caption>
     </>
   )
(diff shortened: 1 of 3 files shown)


src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -27,8 +27,8 @@ interface CaptionProps {
 function Caption({ icon: Icon = AlertTriangle, caption }: CaptionProps) {
   return (
     <>
-      <Icon color="secondary" />
-      <ThemedText.Caption>{caption}</ThemedText.Caption>
+      <Icon color="secondary" size={16.67 / 14} />
```
zzmp, line 30: Is this sizing necessary?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -27,8 +27,8 @@ interface CaptionProps {
...
+      <ThemedText.Caption lineHeight="14px">{caption}</ThemedText.Caption>
```
zzmp, line 31: I thought we were leaving this until post-phase0.

JFrankfurt: no, just the font-weight

alignment is going in

src/components/Swap/Toolbar/Caption.tsx
JFrankfurt: it is... but it looks like it's not actually being applied

JFrankfurt: was working with the old `largeiconcss`, but that's not here now

cmcewen (approved):

PR #266, opened 2022-10-11
feat: add initial analytics handlers

zzmp:


src/components/Swap/Summary/index.tsx
```diff
@@ -95,17 +97,20 @@ function ConfirmButton({
   const [ackPriceImpact, setAckPriceImpact] = useState(false)
 
   const [ackTrade, setAckTrade] = useState(trade)
+  const { onSwapPriceUpdateAck } = useAtomValue(swapEventHandlersAtom)
```
vm, line 100: think you can unswap both of these swapeventhandleratom attributes in one spot?

src/cosmos/ControlledSwap.fixture.tsx
```diff
@@ -46,6 +46,9 @@ function Fixture() {
 
   const connector = useProvider(SupportedChainId.MAINNET)
 
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
vm, line 49: mehh i feel like this makes it harder to read tbh

src/hooks/swap/index.ts
```diff
@@ -68,17 +68,17 @@ export function useIsAmountPopulated() {
...
+export function useSwapAmount(field: Field): [string | undefined, (amount: string, origin?: 'max') => void] {
   const value = useAtomValue(amountAtom)
   const isFieldIndependent = useIsSwapFieldIndependent(field)
   const amount = isFieldIndependent ? value : undefined
 
   const { onAmountChange } = useAtomValue(swapEventHandlersAtom)
   const setSwap = useUpdateAtom(swapAtom)
   const updateAmount = useCallback(
-    (update: string) => {
+    (update: string, origin?: 'max') => {
```
vm, line 79: why the default here?

vm (approved):

src/hooks/swap/index.ts
zzmp: not a default, a typing to only allow 'max'. I did not want a boolean so that it would be extensible in the future.

src/cosmos/ControlledSwap.fixture.tsx
zzmp: this guarantees that all of the events are there, whereas otherwise there's a lot of manual checking required.
it's just meant to make it easier to test, and it's a fixture, so I'm less worried about breaking the rules (of hooks).

PR #268, opened 2022-10-11
fix: do not eagerly connect on null provider

zzmp:


src/hooks/web3/index.tsx
```diff
@@ -72,18 +72,26 @@ export function Provider({
...
     connectors.network.activate().catch(() => undefined)
-  }, [connectors.metaMask, connectors.network, connectors.user, connectors.walletConnect])
+  }, [
+    connectors.metaMask,
+    connectors.network,
+    connectors.user,
+    connectors.walletConnect,
+    provider,
+    shouldEagerlyConnect,
+    web3ReactConnectors.user,
```
cmcewen, line 93: why should this trigger the effect?

src/hooks/web3/index.tsx
```diff
@@ -72,18 +72,26 @@ export function Provider({
...
     }
     connectors.network.activate().catch(() => undefined)
-  }, [connectors.metaMask, connectors.network, connectors.user, connectors.walletConnect])
+  }, [
+    connectors.metaMask,
+    connectors.network,
+    connectors.user,
+    connectors.walletConnect,
+    provider,
+    shouldEagerlyConnect,
```
cmcewen, line 92: this will always change when provider changes right?

zzmp: yes. provider should be omitted.

src/hooks/web3/index.tsx
zzmp: this should be omitted.

cmcewen (approved):

cmcewen (approved):

PR #274, opened 2022-10-12
feat: add an onSwitchChain handler

zzmp:
onSwitchChain allows an integrator to switch chains using their own implementation (eg if their provider does not implement `wallet_switchEthereumChain`) - this is necessary for the network connection in interface.

Includes some type cleanup to ensure the handlers are uniform.


src/components/Widget.tsx
```diff
@@ -94,17 +88,14 @@ export const DialogWrapper = styled.div`
   }
 `
 
-export interface WidgetProps extends TransactionEventHandlers {
+export interface WidgetProps extends TransactionEventHandlers, Web3Props, WidgetEventHandlers {
   theme?: Theme
   locale?: SupportedLocale
-  provider?: Eip1193Provider | JsonRpcProvider | null
```
zzmp, line 100: These are more complex types, so they are used by extending `Web3Props`, which is less error-prone because it is less duplication.

JFrankfurt (approved):

Issue #252, opened 2022-10-11
Bug: defaultChainId not being used


akarys92:
**Bug Description**
When the widget loads with `defaultChainId` that is not equal to the chain ID that the users wallet loaded with, the widget simply switches to what users wallet is set to. This makes the `defaultChainId` prop useless. 

**Steps to Reproduce**

1. Instantiate the swap widget with `defaultChainId` set to 1 (Mainnet)
```
<SwapWidget defaultChainId={1} />
```
2. Open the app, and connect a wallet. Then switch the wallet to a supported, non-Mainnet network like Polygon
3. Refresh the app. Notice that the widget still loads with Polygon as the network, not respecting the `defaultChainId` property

**Expected Behavior**
In this case the widget should display the chain switch behavior and prompt the user to switch to the network defined by `defaultChainId`
<img width="404" alt="image" src="https://user-images.githubusercontent.com/8218221/194981538-db8693cc-13f9-4800-a75f-0eaf3876dacd.png">

PR #281, opened 2022-10-17
fix: the default chainId is picked from`defaultChainId` 

dannydaniil:
**Bug Description**
Attempts to solve: https://github.com/Uniswap/widgets/issues/252

The identified bug was that the `chainId` would initially default to `defaultChainId` as expected, but `chainId` would then be overwritten by a different connector (like MetaMask) which does not accept `defaultChainId` as an argument. The approach is specified below.

**Steps to Reproduce**
1. Run the widget
2. Connect to MetaMask with Polygon
3. Refresh the page
4. Widget is on Polygon

**Expected Behavior**
The requirements after discussions with @akarys92 are:
1. on first load, show default chain and Switch behavior if MetaMask/WalletConnect is on another chain
2. on new connect to MetaMask/WalletConnect on another chain, if there _is_ a default chain, show default chain and Switch behavior; if there is not a default chain, switch to MetaMask/WalletConnect chain
3. on disconnect from MetaMask/WalletConnect, switch to default chain --> This is a softer requirement which is not currently part of the implementation 

**Additional Context**
This change attempts to overcome this issue by updating the `tokensIn` with the corrected `chainId`.
It's correct value depends on what priority the current connector is under, and wether the top priority connector has already been activated.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -78,5 +78,13 @@
         ]
       }
     ]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["*test.ts"], 
+      "rules": {
+        "@typescript-eslint/no-var-requires": "off"
+      }
+    }
+  ]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -78,13 +78,5 @@
         ]
       }
     ]
-  },
-  "overrides": [
-    {
-      "files": ["*test.ts"], 
-      "rules": {
-        "@typescript-eslint/no-var-requires": "off"
-      }
-    }
-  ]
+  }
 }
--- a/src/components/Swap/index.tsx
+++ b/src/components/Swap/index.tsx
@@ -31,7 +31,7 @@ import useValidate from './useValidate'
 export interface SwapProps extends BrandingSettings, FeeOptions, SwapController, SwapEventHandlers, TokenDefaults {
   hideConnectionUI?: boolean
   routerUrl?: string
-  defaultChainId: SupportedChainId
+  defaultChainId?: SupportedChainId
 }
 
 export default function Swap(props: SwapProps) {
--- a/src/components/Swap/index.tsx
+++ b/src/components/Swap/index.tsx
@@ -1,6 +1,7 @@
 import { Trans } from '@lingui/macro'
 import BrandedFooter from 'components/BrandedFooter'
 import Wallet from 'components/ConnectWallet'
+import { SupportedChainId } from 'constants/chains'
 import { SwapInfoProvider } from 'hooks/swap/useSwapInfo'
 import useSyncController, { SwapController } from 'hooks/swap/useSyncController'
 import useSyncConvenienceFee, { FeeOptions } from 'hooks/swap/useSyncConvenienceFee'
@@ -30,6 +31,7 @@ import useValidate from './useValidate'
 export interface SwapProps extends BrandingSettings, FeeOptions, SwapController, SwapEventHandlers, TokenDefaults {
   hideConnectionUI?: boolean
   routerUrl?: string
+  defaultChainId: SupportedChainId
 }
 
 export default function Swap(props: SwapProps) {
@@ -37,7 +39,7 @@ export default function Swap(props: SwapProps) {
   useSyncController(props as SwapController)
   useSyncConvenienceFee(props as FeeOptions)
   useSyncSwapEventHandlers(props as SwapEventHandlers)
-  useSyncTokenDefaults(props as TokenDefaults)
+  useSyncTokenDefaults(props as TokenDefaults, props.defaultChainId)
   useSyncBrandingSetting(props as BrandingSettings)
 
   const [wrapper, setWrapper] = useState<HTMLDivElement | null>(null)
--- a/src/components/Swap/index.tsx
+++ b/src/components/Swap/index.tsx
@@ -1,7 +1,6 @@
 import { Trans } from '@lingui/macro'
 import BrandedFooter from 'components/BrandedFooter'
 import Wallet from 'components/ConnectWallet'
-import { SupportedChainId } from 'constants/chains'
 import { SwapInfoProvider } from 'hooks/swap/useSwapInfo'
 import useSyncController, { SwapController } from 'hooks/swap/useSyncController'
 import useSyncConvenienceFee, { FeeOptions } from 'hooks/swap/useSyncConvenienceFee'
@@ -31,7 +30,6 @@ import useValidate from './useValidate'
 export interface SwapProps extends BrandingSettings, FeeOptions, SwapController, SwapEventHandlers, TokenDefaults {
   hideConnectionUI?: boolean
   routerUrl?: string
-  defaultChainId?: SupportedChainId
 }
 
 export default function Swap(props: SwapProps) {
@@ -39,7 +37,7 @@ export default function Swap(props: SwapProps) {
   useSyncController(props as SwapController)
   useSyncConvenienceFee(props as FeeOptions)
   useSyncSwapEventHandlers(props as SwapEventHandlers)
-  useSyncTokenDefaults(props as TokenDefaults, props.defaultChainId)
+  useSyncTokenDefaults(props as TokenDefaults)
   useSyncBrandingSetting(props as BrandingSettings)
 
   const [wrapper, setWrapper] = useState<HTMLDivElement | null>(null)
--- a/src/hooks/swap/useSyncTokenDefaults.test.ts
+++ b/src/hooks/swap/useSyncTokenDefaults.test.ts
@@ -1,0 +1,76 @@
+import { TradeType } from '@uniswap/sdk-core'
+import { SupportedChainId } from 'constants/chains'
+import { nativeOnChain } from 'constants/tokens'
+import { useAtomValue } from 'jotai/utils'
+import { Field, stateAtom, Swap, swapAtom } from 'state/swap'
+import { renderHook } from 'test'
+
+import { DAI, USDC_MAINNET } from '../../constants/tokens'
+import useSyncTokenDefaults, { TokenDefaults } from './useSyncTokenDefaults'
+
+const MOCK_DAI_MAINNET = DAI
+const MOCK_USDC_MAINNET = USDC_MAINNET
+
+const INITIAL_SWAP: Swap = {
+  type: TradeType.EXACT_INPUT,
+  amount: '10',
+  [Field.INPUT]: MOCK_USDC_MAINNET,
+  [Field.OUTPUT]: MOCK_USDC_MAINNET,
+}
+
+const TOKEN_DEFAULTS: TokenDefaults = {
+  defaultInputAmount: 10,
+  defaultInputTokenAddress: 'NATIVE',
+  defaultOutputAmount: 10,
+  defaultOutputTokenAddress: DAI.address,
+}
+
+jest.mock('@web3-react/core', () => {
+  const original = jest.requireActual('@web3-react/core')
+  return {
+    ...original,
+    useWeb3React: () => ({
+      chainId: 1,
+      connector: {},
+    }),
+  }
+})
+
+jest.mock('../useTokenList', () => {
+  const original = jest.requireActual('../useTokenList')
+  return {
+    ...original,
+    useIsTokenListLoaded: () => true,
+  }
+})
+
+jest.mock('hooks/useCurrency', () => {
+  const original = jest.requireActual('hooks/useCurrency')
+  return {
+    ...original,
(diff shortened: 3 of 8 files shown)


src/components/Swap/index.tsx
```diff
@@ -30,14 +31,15 @@ import useValidate from './useValidate'
...
+  defaultChainId?: SupportedChainId
 }
 
 export default function Swap(props: SwapProps) {
   useValidate(props)
   useSyncController(props as SwapController)
   useSyncConvenienceFee(props as FeeOptions)
   useSyncSwapEventHandlers(props as SwapEventHandlers)
-  useSyncTokenDefaults(props as TokenDefaults)
+  useSyncTokenDefaults(props as TokenDefaults, props.defaultChainId)
```
zzmp, line 42: If `defaultChainId` is part of `useSyncTokenDefaults`, you should add it to `TokenDefaults` instead of passing it here separately.

This effectively means you can back out all changes to this file, and just make the change at the definition of the `TokenDefaults` interface.

src/hooks/swap/useSyncTokenDefaults.ts
```diff
@@ -40,45 +42,73 @@ function useDefaultToken(
...
-export default function useSyncTokenDefaults({
-  defaultInputTokenAddress,
-  defaultInputAmount,
-  defaultOutputTokenAddress,
-  defaultOutputAmount,
-}: TokenDefaults) {
-  const updateSwap = useUpdateAtom(swapAtom)
-  const { chainId } = useWeb3React()
-  const onSupportedNetwork = useOnSupportedNetwork()
+function useDefaultInputToken(
```
zzmp, line 45: Why is this necessary? Unless there is a default set, the default input token should be undefined.

src/hooks/swap/useSyncTokenDefaults.ts
```diff
@@ -40,45 +42,73 @@ function useDefaultToken(
...
+export default function useSyncTokenDefaults(
+  { defaultInputTokenAddress, defaultInputAmount, defaultOutputTokenAddress, defaultOutputAmount }: TokenDefaults,
+  defaultChainId?: SupportedChainId
+) {
   const lastChainId = useRef<number | undefined>(undefined)
+  const lastConnector = useRef<Connector | undefined>(undefined)
+  const updateSwap = useUpdateAtom(swapAtom)
+  const { chainId, connector } = useWeb3React()
+
+  const defaultOutputToken = useDefaultToken(defaultOutputTokenAddress, chainId)
```
zzmp, line 69: I don't think that `chainId` is used to select the appropriate token. See L34.

zzmp (changes requested):

src/hooks/swap/useSyncTokenDefaults.ts
dannydaniil: Great question! So `useDefaultToken` was there from before. My addition was pulling out `useDefaultInputToken`, which used to be some logic defined as a variable which could not be reused. I pulled it out as a hook as I needed it to reuse it in order to get the defaults for the `defaultChainId`

src/hooks/swap/useSyncTokenDefaults.ts
dannydaniil: Good call out! This was actually there from before, it's not an addition. I believed it used [here](https://github.com/Uniswap/widgets/blob/9ebc5881b3968c6ca2307f1eb3c6c952293569a6/src/hooks/swap/useSyncTokenDefaults.ts#L37) to check if we can grab the Native token of the chain.

src/components/Swap/index.tsx
dannydaniil: Will do! I also thought about it tbh when I will create the new param 😄

.eslintrc.json
```diff
@@ -78,5 +78,13 @@
         ]
       }
     ]
-  }
+  },
```
zzmp, line 81: revert this file

src/hooks/swap/useSyncTokenDefaults.test.ts
```diff
@@ -0,0 +1,93 @@
...
+import { SupportedChainId } from 'constants/chains'
+import { nativeOnChain } from 'constants/tokens'
+import { useAtomValue } from 'jotai/utils'
+import { Field, stateAtom, Swap, swapAtom } from 'state/swap'
+import { renderHook } from 'test'
+
+import { DAI, USDC_MAINNET } from '../../constants/tokens'
+import useSyncTokenDefaults, { TokenDefaults } from './useSyncTokenDefaults'
+
+const web3React = require('@web3-react/core')
```
zzmp, line 11: use an import:

    import * as web3React from '@web3-react/core'

The jest.mock should still work, and you can avoid overriding the lint rule.

zzmp (changes requested):

src/hooks/swap/useSyncTokenDefaults.ts
```diff
@@ -40,45 +43,78 @@ function useDefaultToken(
...
   useEffect(() => {
-    const shouldSync = isTokenListLoaded && chainId && chainId !== lastChainId.current
+    const isSwitchChain = chainId && chainId !== lastChainId.current
+    const isSwitchConnector = connector && connector !== lastConnector.current
+    const shouldSync = isTokenListLoaded && (isSwitchChain || isSwitchConnector)
+    const shouldSetToDefaultChainId = isSwitchConnector && defaultChainId
+
     if (shouldSync) {
-      setToDefaults()
+      setToDefaults(shouldSetToDefaultChainId)
```
zzmp, line 114: This should match the parameter name: `shouldUseDefaultChainId`.

src/hooks/swap/useSyncTokenDefaults.ts
```diff
@@ -17,6 +19,7 @@ export interface TokenDefaults {
   defaultInputAmount?: number | string
   defaultOutputTokenAddress?: DefaultAddress
   defaultOutputAmount?: number | string
+  defaultChainId?: SupportedChainId
 }
 
 function useDefaultToken(
```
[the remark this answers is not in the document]
zzmp: I think this can be a fast follow, but there should be a TODO in this code to do it, linked to a jira ticket, ie

    TODO(WEB-1999): useToken will only get a token from the active chain, but should use the passed chainId.

zzmp (changes requested):

zzmp:
I would not consider 252 fixed until useToken properly gets a token from the desired chain (vs the active chain).

src/hooks/useTokenList/index.tsx
```diff
@@ -41,10 +42,15 @@ export default function useTokenList(): WrappedTokenInfo[] {
 
 export type TokenMap = { [address: string]: Token }
 
-export function useTokenMap(): TokenMap {
-  const { chainId } = useWeb3React()
+export function useTokenMap(chainId?: SupportedChainId): TokenMap {
+  let { chainId: _chainId } = useWeb3React()
```
zzmp, line 46: Call this `activeChainId` so there's some semantic differentiation.

src/hooks/useTokenList/index.tsx
```diff
@@ -41,10 +42,15 @@ export default function useTokenList(): WrappedTokenInfo[] {
...
+
+  if (chainId) {
```
zzmp, line 48: This can be simpler as a one-line defaulting statement: `chainId = chainId || activeChainId`

src/hooks/useOnSupportedNetwork.ts
```diff
@@ -1,10 +1,14 @@
 import { useWeb3React } from '@web3-react/core'
-import { ALL_SUPPORTED_CHAIN_IDS } from 'constants/chains'
+import { ALL_SUPPORTED_CHAIN_IDS, SupportedChainId } from 'constants/chains'
 import { useMemo } from 'react'
 
-function useOnSupportedNetwork() {
-  const { chainId } = useWeb3React()
-  return useMemo(() => Boolean(chainId && ALL_SUPPORTED_CHAIN_IDS.includes(chainId)), [chainId])
+function useOnSupportedNetwork(chainId?: SupportedChainId) {
+  let { chainId: _chainId } = useWeb3React()
```
zzmp, line 6: See comments for `useTokenList/index.tsx`

src/hooks/useCurrency.ts
```diff
@@ -90,8 +91,8 @@ export function useTokenFromMapOrNetwork(tokens: TokenMap, tokenAddress?: string
  * Returns null if token is loading or null was passed.
  * Returns undefined if tokenAddress is invalid or token does not exist.
  */
-export function useToken(tokenAddress?: string | null): Token | null | undefined {
-  const tokens = useTokenMap()
+export function useToken(tokenAddress?: string | null, chainId?: SupportedChainId): Token | null | undefined {
+  const tokens = useTokenMap(chainId)
   return useTokenFromMapOrNetwork(tokens, tokenAddress)
```
zzmp, line 96: If there is a `chainId` passed, and `chainId !== activeChainId`, the network should not be queried.
`useTokenFromMapOrNetwork` does not use chainId, so I would pass this as an extra boolean parameter: `return useTokenFromMapOrNetwork(tokens, tokenAddress, /*skipNetwork=*/ true)`.

    export function useTokenFromMapOrNetwork(tokens: TokenMap, tokenAddress?: string | null, skipNetwork = false): ... {

src/hooks/swap/useSyncTokenDefaults.ts
```diff
@@ -17,68 +18,92 @@ export interface TokenDefaults {
...
   defaultAddress: DefaultAddress | undefined,
-  chainId: number | undefined
+  chainId: number | undefined,
+  defaultToNative: boolean
 ): Currency | undefined {
-  let address = undefined
+  let address: undefined | string | Token = undefined
```
zzmp, line 29: This will never be `Token`.

zzmp (changes requested):

src/hooks/swap/useSyncTokenDefaults.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  [Field.OUTPUT]: MOCK_DAI_MAINNET,
+}
+
+const TOKEN_DEFAULTS: TokenDefaults = {
+  defaultInputAmount: 10,
+  defaultInputTokenAddress: 'NATIVE',
+  defaultOutputTokenAddress: 'NATIVE',
+}
+
+jest.mock('../useTokenList', () => {
```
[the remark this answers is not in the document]
dannydaniil: I was wondering what import you are referring to here. Is it the import of `@web3-react/core` ? The comment is on a mock line

src/hooks/useCurrency.ts
dannydaniil: Good call I changed this. Unfortunately, I had to use `useWeb3React` inside `useToken` to get`activeChainId`

src/hooks/useCurrency.ts
```diff
@@ -76,23 +77,29 @@ export function useTokenFromNetwork(tokenAddress: string | null | undefined): To
...
+  tokenAddress?: string | null,
+  skipNetwork = false
+): Token | null | undefined {
   const address = isAddress(tokenAddress)
   const token: Token | undefined = address ? tokens[address] : undefined
 
   const tokenFromNetwork = useTokenFromNetwork(token ? undefined : address ? address : undefined)
 
-  return tokenFromNetwork ?? token
+  return skipNetwork || !tokenFromNetwork ? token : tokenFromNetwork
```
zzmp, line 90: If `skipNetwork` you should return `undefined`:

    return skipNetwork ? undefined : tokenFromNetwork || token

src/hooks/useCurrency.ts
```diff
@@ -76,23 +77,29 @@ export function useTokenFromNetwork(tokenAddress: string | null | undefined): To
...
  * Returns undefined if tokenAddress is invalid or token does not exist.
  */
-export function useToken(tokenAddress?: string | null): Token | null | undefined {
-  const tokens = useTokenMap()
-  return useTokenFromMapOrNetwork(tokens, tokenAddress)
+export function useToken(tokenAddress?: string | null, chainId?: SupportedChainId): Token | null | undefined {
+  const { chainId: activeChainId } = useWeb3React()
+
+  const tokens = useTokenMap(chainId)
+  return useTokenFromMapOrNetwork(tokens, tokenAddress, chainId && chainId !== activeChainId)
```
zzmp, line 102: const skipNetwork = chainId && chainId !== activeChainId
    return useTokenFromMapOrNetwork(tokens, tokenAddress, skipNetwork)

zzmp (changes requested):

zzmp (approved):

PR #285, opened 2022-10-18
refactor: name change and removing token lists

Jesse-Sawa:
* INFURA -> RPC, future chains will likely use other providers 
* tokenLists are only required for testing, [it's imported already](https://github.com/Uniswap/widgets/blob/main/src/cosmos/Swap.fixture.tsx#L1)


zzmp (changes requested):
It looks like the src/constants changes are just removing dead code :+1:.

The src/cosmos changes should all be rolled back, unless you want to actually update the urls in src/cosmos/useProvider.ts to be publicly available providers. These are used for testing only, and not exported, so infura is still appropriate.

Jesse-Sawa:
> 

Ah okay, the name refactor was in result of adding non-infura (public facing) urls here for Celo support. 

But now that you mention it, the token list used for testing doesn't support Celo + likely any future chain additions. I think this change should include support for additional lists as well, which would make this name refactor relevant. Wdyt?

The alternative is to leave this as is and and testing for future chain support would just be done more manually

Jesse-Sawa:
> It looks like the src/constants changes are just removing dead code 👍.
> 
> The src/cosmos changes should all be rolled back, unless you want to actually update the urls in src/cosmos/useProvider.ts to be publicly available providers. These are used for testing only, and not exported, so infura is still appropriate.

Ah okay, the name refactor was in result of adding non-infura (public facing) urls here for Celo support. 

But now that you mention it, the token list used for testing doesn't support Celo + likely any future chain additions. I think this change should include support for additional lists as well, which would make this name refactor relevant. Wdyt?

The alternative is to leave this as is and and testing for future chain support would just be done more manually

PR #289, opened 2022-10-19
feat: desaturate unapproved input token

zzmp:
- Simplifies approval codepaths, and includes approval in swap info state
  - I wanted to clean these codepaths, but don't think it is worth the effort to consolidate them further.
- Renders an unapproved input token in grayscale

See DE-76. Replaces #215.


src/components/TokenSelect/TokenButton.tsx
```diff
@@ -13,13 +13,17 @@ const transitionCss = css`
...
   border-radius: ${({ theme }) => theme.borderRadius}em;
   padding: 0.25em;
 
   :enabled {
     ${({ transition }) => transition && transitionCss};
   }
+
+  ${TokenImg} {
+    filter: ${({ approved }) => approved === false && 'grayscale(1)'};
+  }
```
JFrankfurt, line 26: I prefer to have some kind of explicit unset value here instead of creating a floating invalid `filter` rule: `filter: `

JFrankfurt (approved):
lgtm--This would have been much easier to review as two PRs though. The actual desaturation code is very small, and I had to kinda figure out what all the rest of the code was doing on my own.

PR #293, opened 2022-10-25
fix: ignore integrator onSwitchChain errors

zzmp:


src/hooks/useSwitchChain.ts
```diff
@@ -77,8 +77,12 @@ export default function useSwitchChain(): (chainId: SupportedChainId) => Promise
 
         // The user connector may require custom logic: try onSwitchChain if it is available.
         if (connector === connectors.user) {
-          const switching = onSwitchChain?.(addChainParameter)
-          if (switching) return switching
+          try {
+            const switching = onSwitchChain?.(addChainParameter)
```
vm, line 81: don't you need to await on this line instead of the line below?

zzmp: It's poorly documented, but I only want to `await` and `return` if an action was taken by the integrator, which is signaled by the function returning a `Promise`. So `if (!switching)` then I do not want to `await`.

If I were to `await` on this line, then `switching === undefined` and there would be nothing to check.

vm (approved):

PR #295, opened 2022-10-27
fix: add space between usd, impact

zzmp:
- Adds space between usd and impact using a gap, as it was not being respected as html
- Makes impact non-selectable (to further distinguish it from price)


lynnshaoyu:
could we add screenshots for visual changes to PR descriptions in the future?🙏

lynnshaoyu (approved):

PR #299, opened 2022-11-08
feat: preload client side routes

zzmp:
Prices (and routes, if no routerUrl is specified) are loaded using the client-side router, which must initialize a route before it can fetch prices for it. Currently, this initialization is only done _after_ a swap is fetched (ie a waterfall pattern).

Updates the routing code to initialize a route as soon as its Token is selected. This has two important consequences: client-fetched swaps are significantly faster, since the delay is mostly initialization; prices show up (nearly) at the same time as the swap, since (again) the delay is mostly initialization.

Compare these two experiences for loading the DAI price (note the latter is instant):

https://user-images.githubusercontent.com/5403956/200449717-15747b84-2d1f-4c1c-b547-81388f968ff1.mov

_Before_

https://user-images.githubusercontent.com/5403956/200449744-0d016842-242c-4389-a200-2c8c911b7aba.mov

_After_

Compare these two experiences for loading the ETH/DAI swap using a client-side router (note the latter is faster):

https://user-images.githubusercontent.com/5403956/200450271-fcf2f3ab-2305-447a-b89b-8ade71e6eeaf.mov

_Before_

https://user-images.githubusercontent.com/5403956/200450148-4232be41-3c32-4e47-b2ff-6a9ebd6c5d37.mov

_After_

### DIFF
--- a/src/cosmos/Swap.fixture.tsx
+++ b/src/cosmos/Swap.fixture.tsx
@@ -59,7 +59,7 @@ function Fixture() {
     USDC: USDC_MAINNET.address,
   }
   const defaultInputToken = useOption('defaultInputToken', { options: currencies, defaultValue: 'Native' })
-  const [defaultInputAmount] = useValue('defaultInputAmount', { defaultValue: 1 })
+  const [defaultInputAmount] = useValue('defaultInputAmount', { defaultValue: 0 })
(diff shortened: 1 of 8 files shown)


cartcrom (approved):
Looks great

PR #301, opened 2022-11-16
fix: retain cursor when updating controlled amount

zzmp:
Fixes a reset of the `<input>` state by ensuring the update can be processed synchronously.

Previously, controlled state had a delay because of the use of `useEffect` to detect changes. This was unnecessary, as the setter should be a nop if an update referentially equivalent to the current state is passed to it.


yyip-dev (commented):
Lgtm! Quick question on the preview - is it supposed to work? I see you updated the controlled swap fixture but I still can't enter anything in it

zzmp:
@yyip-dev you'll need to use the cosmos selectors to choose a token - not the widget itself.
Once a token is selected, you can enter the amount in the widget.

yyip-dev (approved):
Looking good! Ship it 🚢

PR #303, opened 2022-11-18
fix: listen for walletconnect tx success

zzmp:
Upgrades walletconnect deps to a version that will recognize successful tx.

See https://github.com/Uniswap/web3-react/pull/702#issuecomment-1310195922


yyip-dev (approved):
Versions look correct. I checked out `@web3-react/walletconnect` latest tag as well as context on https://github.com/Uniswap/web3-react/pull/702#issuecomment-1310195922

PR #308, opened 2022-11-22
refactor: mv branding to flags

zzmp:
- Creates a new interface, `Flags`, for settings which cannot be changed without remounting
- Moves `brandedFooter` to new `Flags` type.

This is a precursor to adding Permit2, which will be passed as a flag.


cartcrom (approved):
Looks great

PR #311, opened 2022-11-29
feat: UI design tweaks

cartcrom:
Updated font sizes, weights, & made general spacing changes. Also updated caption messages & fixed disabled button state to match design.


src/components/Swap/ReverseButton.tsx
```diff
@@ -17,7 +17,7 @@ const StyledReverseButton = styled(Button)`
   justify-content: center;
   left: 50%;
   position: absolute;
-  transform: translate(-50%, -60%);
+  transform: translate(-50%, calc(-50% - 2px));
```
zzmp, line 20: Comment on why the `-2px` are necessary, so that they're not inadvertently removed in the future.

zzmp (changes requested):

src/components/ActionButton.tsx
```diff
@@ -11,6 +11,11 @@ const StyledButton = styled(Button)`
   flex-grow: 1;
   max-height: 56px;
   transition: background-color 0.25s ease-out, border-radius 0.25s ease-out, flex-grow 0.25s ease-out;
+  ${({ theme, disabled }) =>
```
zzmp, line 14: you're already writing css. Could you simplify this, or do you need to wrap it to avoid overwriting the background-color. eg

    background-color: ${({ disabled, theme }) => disabled && theme.interactive };

cartcrom: Yes, was avoiding an overwrite here

src/components/Swap/ReverseButton.tsx
cartcrom: done

zzmp (approved):

PR #319, opened 2022-12-05
refactor: pass approval amount from info

zzmp:
Prepares the way for permit2, which will need to check approval for the same amount through a novel codepath.


yyip-dev (approved):
Lgtm!

PR #321, opened 2022-12-05
feat: permit2 and universal-router integration

zzmp:
Pending a few other things:
- https://github.com/Uniswap/permit2-sdk/pull/18 (bug preventing recommended integration)
- https://github.com/Uniswap/permit2-sdk/pull/19 (simpler ethers nested dependency)
- https://github.com/Uniswap/widgets/pull/318 (errors as re-throwable strings)
- https://github.com/Uniswap/widgets/pull/319 (re-usable approval amount)
- https://github.com/Uniswap/widgets/pull/320 (permit2 feature flag)


src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export function useUniversalRouterSwapCallback(trade: InterfaceTrade | undefined, options: SwapOptions) {
+  const { account, chainId, provider } = useWeb3React()
+
+  return useCallback(async (): Promise<TransactionResponse> => {
+    try {
+      if (!account || !chainId || !provider || !trade) throw new Error('missing input')
+
+      const calldata = SwapRouter.swapERC20CallParameters(trade, {
```
JFrankfurt, line 29:
```suggestion
      const { calldata, value } = SwapRouter.swapERC20CallParameters(trade, {
```

```suggestion
      const methodParameters = SwapRouter.swapERC20CallParameters(trade, {
```

JFrankfurt: either of these seem like better options, since the variable isn't actually calldata itself, and these allow you to avoid `calldata.calldata` below

src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,87 @@
...
+  permit?: PermitSignature
+  feeOptions?: FeeOptions
+}
+
+export function useUniversalRouterSwapCallback(trade: InterfaceTrade | undefined, options: SwapOptions) {
+  const { account, chainId, provider } = useWeb3React()
+
+  return useCallback(async (): Promise<TransactionResponse> => {
+    try {
+      if (!account || !chainId || !provider || !trade) throw new Error('missing input')
```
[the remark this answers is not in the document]
JFrankfurt: basically, some hint about what's going on so a user can report it or we can debug it more easily

src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,87 @@
...
+        slippageTolerance: options.slippageTolerance,
+        deadlineOrPreviousBlockhash: options.deadline?.toString(),
+        inputTokenPermit: options.permit,
+        fee: options.feeOptions,
+      })
+      const tx =
+        calldata.value && !isZero(calldata.value)
+          ? {
+              from: account,
+              to: UNIVERSAL_ROUTER_ADDRESS(chainId),
```
JFrankfurt, line 39: this will throw on unsupported chains--are you expecting that/handling it the way you'd like?

src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,87 @@
...
+            })
+        })
+        .then((gasEstimate) =>
+          provider.getSigner().sendTransaction({
+            ...tx,
+            gasLimit: calculateGasMargin(gasEstimate),
+          })
+        )
+        .catch((error) => {
+          throw new Error(swapErrorToUserReadableMessage(error))
```
JFrankfurt, line 71: does our swapError handler actually handle all relevant errors from the new router?

src/hooks/usePermitAllowance.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      const permit: Permit = {
+        details: {
+          token: token.address,
+          amount: MaxAllowanceTransferAmount,
+          expiration: toDeadline(PERMIT_EXPIRATION),
+          nonce,
+        },
+        spender,
+        sigDeadline: toDeadline(PERMIT_SIG_EXPIRATION),
```
JFrankfurt, line 76: for my own edification, what's the diff here btwn the two expiration constants and why?

src/hooks/usePermitAllowance.ts
```diff
@@ -0,0 +1,88 @@
...
+const PERMIT_EXPIRATION = ms`30d`
+const PERMIT_SIG_EXPIRATION = ms`30m`
+
+function toDeadline(expiration: number): number {
+  return Math.floor((Date.now() + expiration) / 1000)
+}
+
+export function usePermitAllowance(token?: Token, spender?: string) {
+  const { account, provider } = useWeb3React()
+  const allowanceProvider = useMemo(() => provider && new AllowanceProvider(provider, PERMIT2_ADDRESS), [provider])
```
JFrankfurt, line 23: Why is there a new provider used here? I expected the PERMIT2_ADDRESS to be used with a contract constructor instead.

src/hooks/useTokenAllowance.ts
```diff
@@ -15,3 +18,32 @@ export function useTokenAllowance(token?: Token, owner?: string, spender?: strin
     [token, allowance]
   )
 }
+
+export function useTokenAllowanceCallback(amount?: CurrencyAmount<Token>, spender?: string) {
```
JFrankfurt, line 22:
```suggestion
export function useSetTokenAllowanceCallback(amount?: CurrencyAmount<Token>, spender?: string) {
```

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,75 @@
...
+  useEffect(() => setPermitAllowance(allowanceData?.amount), [allowanceData?.amount])
+
+  const [signature, setSignature] = useState<PermitSignature>()
+  const permitToken = usePermitAllowanceCallback(amount?.currency, spender, allowanceData?.nonce, setSignature)
+
+  const approveAndPermitToken = useCallback(async () => {
+    // Queue both transactions. Delay the permit to ensure the approval is prompted for first.
+    const info = approveToken()
+    await new Promise((resolve) => setTimeout(resolve, 500))
+    await permitToken()
```
JFrankfurt, line 41: it looks like `approveToken` is a promise. Could you `.then` chain or `await` the call to trigger `permitToken` instead of using the timeout?

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,75 @@
...
+    return await info
```
JFrankfurt, line 42: iirc an `await` on a return line doesn't do anything--the downstream consumer will be working with a promise either way. So I thiiink you can drop the `await`.

src/components/Swap/Summary/index.tsx
```diff
@@ -146,7 +147,7 @@ function ConfirmButton({
         },
       }}
     >
-      <Trans>Confirm swap</Trans>
+      {isPending ? <Trans>Confirm</Trans> : <Trans>Confirm swap</Trans>}
```
JFrankfurt, line 150: this seems like an odd copy distinction to me 🙃

src/components/Swap/SwapActionButton/useOnSubmit.ts
```diff
@@ -13,14 +13,15 @@ export default function useOnSubmit() {
...
       try {
-        info = await submit()
+        info = await submit?.()
       } catch (e) {
-        console.error('Failed to submit', e)
+        console.error(e)
+        return false
       }
-      if (!info) return false
+      if (!info) return true
```
JFrankfurt, line 24: this change seems significant, but it's hard to me to fully understand what's going on here. (Except that I know I'd do something like invert a bool for testing and maybe forget to switch it back.) Confirming this is intentional.

src/hooks/swap/useSyncController.ts
```diff
@@ -22,10 +22,10 @@ export default function useSyncController({ value, settings }: SwapController):
   }, [settings, value])
 
   const setSwap = useUpdateAtom(swapAtom)
-  setSwap(() => value)
+  if (value) setSwap(() => value)
 
   const setSettings = useUpdateAtom(settingsAtom)
-  setSettings(() => settings)
+  if (settings) setSettings(() => settings)
```
JFrankfurt, line 28: these read like it might be impossible to clear state

src/components/Swap/SwapActionButton/Permit2Button.tsx
```diff
@@ -0,0 +1,91 @@
...
+    setIsPending(false)
+    setIsFailed(false)
+  }, [currency])
+
+  const onClick = useCallback(async () => {
+    setIsPending(true)
+    const submitted = await onSubmit(callback)
+    setIsFailed(!submitted)
+    setIsPending(false)
+  }, [callback, onSubmit])
```
JFrankfurt, line 45: nit: would prefer this be `try/catch/finally`

JFrankfurt (changes requested):
Very cool! Kind of a big change, so lots of questions/confirmations, but overall I like it a lot.

src/hooks/usePermitAllowance.ts
marktoda: SigDeadline describes when the signature is no longer valid. Expiration describes when the _approval_ is no longer valid. So if we submit our permit within 30 minutes then we have unlimited approval for 30 days after that

src/hooks/usePermitAllowance.ts
marktoda: This is just a helper provider from permit2-sdk that wraps the contract object with nicer helpers

src/hooks/usePermit2.ts
zzmp: Done.
NB: The exception here is if you are in a try-catch - then `return await x` will throw any rejection within the try-block.

src/hooks/useUniversalRouter.ts
zzmp: I've added a check to useSyncFlags.ts#usePermit2 so that permit2 will only be enabled on chains on which the Universal Router is supported.

src/components/Swap/SwapActionButton/ApproveButton.tsx
```diff
@@ -50,7 +52,9 @@ export default function ApproveButton({
   // Reset the pending state if currency changes.
   useEffect(() => setIsPending(false), [currency])
 
-  const pendingApprovalHash = usePendingApproval(currency?.isToken ? currency : undefined)
+  const { chainId } = useWeb3React()
+  const spender = chainId ? SWAP_ROUTER_ADDRESSES[chainId] : undefined
```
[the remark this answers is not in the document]
zzmp: I've added a check to useSyncFlags.ts#usePermit2 so that permit2 will only be enabled on chains on which the Universal Router is supported.

src/hooks/useUniversalRouter.ts
zzmp: No, this will need to be revisited in the future. For now, I'm just treating it as a generic error parser.

src/hooks/usePermitAllowance.ts
```diff
@@ -0,0 +1,88 @@
...
+
+interface Permit extends PermitSingle {
+  sigDeadline: number
+}
+
+export interface PermitSignature extends Permit {
+  signature: string
+}
+
+export function usePermitAllowanceCallback(
```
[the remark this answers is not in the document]
zzmp: I don't want to change this. It matches with `useTokenAllowanceCallback`, and adding `Callback` at the end is a signifier to me that it sets the `PermitAllowance`.

src/components/Swap/SwapActionButton/useOnSubmit.ts
zzmp: This was a bug that is now fixed - this now returns false if the submission fails, instead of throwing.

src/hooks/swap/useSyncController.ts
zzmp: you're 100% correct - I've reverted this file in this PR.

src/hooks/usePermitAllowance.ts
```diff
@@ -0,0 +1,88 @@
...
+  const { account, chainId, provider } = useWeb3React()
+
+  return useCallback(async () => {
+    try {
+      if (!chainId || !provider || !token || !spender || nonce === undefined) throw new Error('missing input')
+
+      const permit: Permit = {
+        details: {
+          token: token.address,
+          amount: MaxAllowanceTransferAmount,
```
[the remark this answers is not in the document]
JFrankfurt: sweet! didn't understand that expiration portion

src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,76 @@
...
+      let gasEstimate: BigNumber
+      try {
+        gasEstimate = await provider.estimateGas(tx)
+      } catch (gasError) {
+        await provider.call(tx) // this should throw the actual error
+        throw new Error('unexpected issue with gas estimation; please try again')
+      }
+      const gasLimit = calculateGasMargin(gasEstimate)
+      const response = await provider.getSigner().sendTransaction({ ...tx, gasLimit })
+      return response
```
JFrankfurt, line 61: nit:
```suggestion
      return await provider.getSigner().sendTransaction({ ...tx, gasLimit })
```

zzmp: This is intentional because I find the `return await` construct confusing, and am tempted to remove the `await` (see https://github.com/Uniswap/widgets/pull/321#discussion_r1041172682 lol). In this case the `await` is important so that any rejection occurs within the try-block.

src/components/ActionButton.tsx
```diff
@@ -65,6 +66,7 @@ export const Overlay = styled(Row)<{ hasAction: boolean }>`
 export interface Action {
   message: ReactNode
   icon?: Icon
+  tooltip?: ReactNode
```
vm, line 69: this should be called tooltipContent imo

src/components/Swap/Summary/index.tsx
vm: this seems weird to me too... can we get design to check this

src/hooks/usePermitAllowance.ts
```diff
@@ -0,0 +1,92 @@
...
+interface Permit extends PermitSingle {
+  sigDeadline: number
+}
+
+export interface PermitSignature extends Permit {
+  signature: string
+}
+
+export function useUpdatePermitAllowance(
+  token?: Token,
```
vm, line 57: can't we just require these values vs making them optional and throwing?

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -91,8 +106,14 @@ export default function SwapButton({
     setOpen(await onReviewSwapClick())
   }, [onReviewSwapClick])
 
-  if (approval.state !== SwapApprovalState.APPROVED && !disabled) {
-    return <ApproveButton color={color} onSubmit={onSubmit} trade={trade} {...approval} />
+  if (usePermit2()) {
```
vm, line 109: you can combine these nested ifs fwiw

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -29,28 +33,37 @@ export default function SwapButton({
...
+  const permit2 = usePermit2()
+  const { callback: swapRouterCallback } = useSwapCallback({
+    trade: permit2 ? undefined : trade,
     allowedSlippage: slippage.allowed,
     recipientAddressOrName: account ?? null,
     signatureData: approval?.signatureData,
     deadline,
     feeOptions,
   })
+  const universalRouterCallback = useUniversalRouterSwapCallback(permit2 ? trade : undefined, {
```
vm, line 60: can you call this `universalRouterSwapCallback` or rename the hook?

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -29,28 +33,37 @@ export default function SwapButton({
...
     impact,
-    approval,
   } = useSwapInfo()
   const deadline = useTransactionDeadline()
   const feeOptions = useAtomValue(feeOptionsAtom)
 
-  const { callback: swapCallback } = useSwapCallback({
-    trade,
+  const permit2 = usePermit2()
+  const { callback: swapRouterCallback } = useSwapCallback({
```
vm, line 52: we may wanna rename this hook now to `useAutoRouterCallback`?

src/hooks/swap/useSwapInfo.tsx
```diff
@@ -97,11 +101,16 @@ function useComputeSwapInfo(routerUrl?: string): SwapInfo {
   const slippage = useSlippage(trade)
   const impact = usePriceImpact(trade.trade)
 
+  const permit2 = usePermit2()
   const maximumAmountIn = useMemo(() => {
     const maximumAmountIn = trade.trade?.maximumAmountIn(slippage.allowed)
     return maximumAmountIn?.currency.isToken ? (maximumAmountIn as CurrencyAmount<Token>) : undefined
   }, [slippage.allowed, trade.trade])
-  const approval = useSwapApproval(maximumAmountIn)
+  const approval = useSwapApproval(permit2 ? undefined : maximumAmountIn)
```
vm, line 109: a bit confused what's happening here

src/components/Swap/SwapActionButton/Permit2Button.tsx
```diff
@@ -0,0 +1,97 @@
...
+  color,
+  trade,
+  state,
+  callback,
+  onSubmit,
+}: {
+  color: keyof Colors
+  trade?: InterfaceTrade
+  state: PermitState
+  callback?: () => Promise<ApprovalTransactionInfo | void>
```
vm, line 28: can't this just be inside the onSubmit? why a callback here

src/components/Swap/SwapActionButton/Permit2Button.tsx
```diff
@@ -0,0 +1,97 @@
...
+      setIsFailed(true)
+    } finally {
+      setIsPending(false)
+    }
+  }, [callback, onSubmit])
+
+  const pendingApproval = usePendingApproval(currency?.isToken ? currency : undefined, PERMIT2_ADDRESS)
+
+  const actionProps = useMemo(() => {
+    switch (state) {
```
vm, line 56: is this to force that all states are checked?

src/components/Swap/SwapActionButton/Permit2Button.tsx
```diff
@@ -0,0 +1,97 @@
...
+      }
+    } else if (isFailed) {
+      return {
+        message: t`Approval failed`,
+        onClick,
+      }
+    } else {
+      return {
+        tooltip: t`Permission is required for Uniswap to swap each token. This will expire after one month for your security.`,
+        message: `Approve use of ${currency?.symbol ?? 'token'}`,
```
vm, line 86: no translate?

src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,76 @@
...
+  permit?: PermitSignature
+  feeOptions?: FeeOptions
+}
+
+export function useUniversalRouterSwapCallback(trade: InterfaceTrade | undefined, options: SwapOptions) {
+  const { account, chainId, provider } = useWeb3React()
+
+  return useCallback(async (): Promise<TransactionResponse> => {
+    try {
+      if (!account) throw new Error('missing account')
```
vm, line 27: should these be required instead of erroring here?

src/hooks/useTokenAllowance.ts
```diff
@@ -15,3 +18,34 @@ export function useTokenAllowance(token?: Token, owner?: string, spender?: strin
     [token, allowance]
   )
 }
+
+export function useUpdateTokenAllowance(amount?: CurrencyAmount<Token>, spender?: string) {
```
vm, line 22: filename and function name should be consistent

src/hooks/useTokenAllowance.ts
```diff
@@ -15,3 +18,34 @@ export function useTokenAllowance(token?: Token, owner?: string, spender?: strin
...
+  const contract = useTokenContract(amount?.currency.address)
+
+  return useCallback(async (): Promise<ApprovalTransactionInfo> => {
+    try {
+      if (!amount) throw new Error('missing amount')
+      if (!contract) throw new Error('missing contract')
+      if (!spender) throw new Error('missing spender')
+
+      let allowance: BigNumberish = MaxUint256.toString()
```
vm, line 31: why does this need to be in string form?

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,85 @@
...
+  signature?: PermitSignature
+  callback?: () => Promise<ApprovalTransactionInfo | void>
+}
+
+export default function usePermit(amount?: CurrencyAmount<Token>, spender?: string): Permit {
+  const { account } = useWeb3React()
+  const tokenAllowance = useTokenAllowance(amount?.currency, account, PERMIT2_ADDRESS)
+  const updateTokenAllowance = useUpdateTokenAllowance(amount, PERMIT2_ADDRESS)
+
+  const allowanceData = usePermitAllowance(amount?.currency, spender)
```
vm, line 30: permitAllowanceData

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,85 @@
...
+  const [permitAllowance, setPermitAllowance] = useState(allowanceData?.amount)
```
vm, line 31: permitAllowanceAmount?

vm, line 31: i think this should be called permitAllowanceAmount

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,85 @@
...
+    // but do not wait until token allowance resolves, to avoid some wallets closing their modal in between prompts (eg MetaMask).
+    await new Promise((resolve) => setTimeout(resolve, 500))
+    await updatePermitAllowance()
+    return info
+  }, [updatePermitAllowance, updateTokenAllowance])
+
+  // Trigger a re-render if either tokenAllowance or signature expire.
+  useInterval(
+    () => {
+      const now = Date.now() / 1000 - 12 // ensure it can still go into this block (assuming 12s block time)
```
vm, line 50: could use the 12 constant here potentially

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (signature && signature.sigDeadline < now) {
```
vm, line 51: comment here plz

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,85 @@
...
+        setSignature(undefined)
+      }
+      if (allowanceData && allowanceData.expiration < now) {
```
vm, line 54: comment here plz

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const allowanceData = usePermitAllowance(amount?.currency, spender)
+  const [permitAllowance, setPermitAllowance] = useState(allowanceData?.amount)
+  useEffect(() => setPermitAllowance(allowanceData?.amount), [allowanceData?.amount])
+
+  const [signature, setSignature] = useState<PermitSignature>()
+  const updatePermitAllowance = useUpdatePermitAllowance(amount?.currency, spender, allowanceData?.nonce, setSignature)
+
+  const updateTokenAndPermitAllowance = useCallback(async () => {
+    // Queue both transactions.
```
vm, line 38: this comment is confusing because it implies that the next line queues both transactions

src/hooks/usePermitAllowance.ts
zzmp: this is easier because of hooks

src/components/Swap/SwapActionButton/SwapButton.tsx
zzmp: i'm not sure i can

src/components/Swap/SwapActionButton/SwapButton.tsx
zzmp: I don't want to touch any of the old code. I'm hoping to just gut it in the very near future.

src/hooks/swap/useSwapInfo.tsx
zzmp: passing undefined if it is not the code path we want, to prevent nested hooks from executing and fetching on-chain or off-chain data

src/components/Swap/SwapActionButton/Permit2Button.tsx
zzmp: yes

zzmp: removed

src/hooks/useUniversalRouter.ts
```diff
@@ -0,0 +1,76 @@
...
+import { PermitSignature } from './usePermitAllowance'
+
+interface SwapOptions {
+  slippageTolerance: Percent
+  deadline?: BigNumber
+  permit?: PermitSignature
+  feeOptions?: FeeOptions
+}
+
+export function useUniversalRouterSwapCallback(trade: InterfaceTrade | undefined, options: SwapOptions) {
```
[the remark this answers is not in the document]
zzmp: Universal Router is capable of other types of trades, so I wanted the function name to be more specific.

src/hooks/useTokenAllowance.ts
zzmp: this is a setter/getter file, so I make an exception

src/hooks/useTokenAllowance.ts
zzmp: conversion between large number representations

src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,85 @@
...
+import { useWeb3React } from '@web3-react/core'
+import ms from 'ms.macro'
+import { useCallback, useEffect, useMemo, useState } from 'react'
+
+import { ApprovalTransactionInfo } from '..'
+import useInterval from './useInterval'
+import { PermitSignature, usePermitAllowance, useUpdatePermitAllowance } from './usePermitAllowance'
+import { useTokenAllowance, useUpdateTokenAllowance } from './useTokenAllowance'
+
+export enum PermitState {
```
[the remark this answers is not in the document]
zzmp: I think we made a mistake not doing so with initial widget atoms implementation

PR #332, opened 2022-12-09
fix: hexlify tx value from SwapRouter

zzmp:
Hexlifies the value returned from universal-router-sdk, which is required for ethers to accept it as input.
Fixes an error thrown by ethers (RE: un-hexlified data).
See https://github.com/Uniswap/universal-router-sdk/issues/92 (fixed in https://github.com/Uniswap/universal-router-sdk/pull/93 but not yet released).


vm (approved):
see comment in interface for a diff syntax here

PR #335, opened 2022-12-09
fix: disable ActionButton whilst awaiting wallet action

zzmp:
Prevents a user from double-sending transactions to their wallet (or the UX from implying that this is desirable given our UX flow).


JFrankfurt (approved):
this is actually possible! You can queue multiple actions in a wallet

zzmp:
> this is actually possible! You can queue multiple actions in a wallet

Totally possible, but doesn't match the UX of other actions. Maybe we should bring it up with design?

PR #336, opened 2022-12-09
fix: observe submitted allowance immediately

zzmp:
Fixes a bug where a user could submit a transaction, but would be required to re-sign a permit for a subsequent transaction, because the prior transaction's modification of allowance was not detected.

### DIFF
--- a/src/hooks/usePermitAllowance.ts
+++ b/src/hooks/usePermitAllowance.ts
@@ -11,6 +11,8 @@ import { useWeb3React } from '@web3-react/core'
 import ms from 'ms.macro'
 import { useCallback, useEffect, useMemo, useState } from 'react'
 
+import useBlockNumber from './useBlockNumber'
+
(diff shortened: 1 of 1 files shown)


src/hooks/usePermitAllowance.ts
```diff
@@ -23,6 +25,10 @@ export function usePermitAllowance(token?: Token, spender?: string) {
   const allowanceProvider = useMemo(() => provider && new AllowanceProvider(provider, PERMIT2_ADDRESS), [provider])
   const [allowanceData, setAllowanceData] = useState<AllowanceData>()
 
+  // If there is no allowanceData, recheck every block so a submitted allowance is immediately observed.
+  const blockNumber = useBlockNumber()
+  const shouldUpdate = allowanceData ? false : blockNumber
```
vm, line 30: this is a bit clever but i don't have a simple alternative option

vm (approved):

PR #340, opened 2022-12-15
fix: mark permit not syncing if not permitted

zzmp:


vm (approved):
i wish we had tests for this

PR #341, opened 2022-12-16
feat: fix permit2 approval syncing, etc

zzmp:
- Fixes block fast-forwarding. This is required for Permit2 approval syncing, as explained in a comment in `usePermit2Allowance`.
- Renames Permit2 usages to focus on `Allowance`: Permit2 abstracts over `TokenAllowance` and `PermitAllowance`, and grants one general `Allowance`. This rename also allows simplifying states, eg adding an `isApprovalLoading` boolean to the `Allowance` interface.
- Retypes Permit2 states to include an `AllowanceRequired`, which improves typing for the `AllowanceButton`.
- Parallelizes the code between `useTokenAllowance` and `usePermitAllowance`. Moves away from the permit2-sdk in favor of using the Permit2 ABI directly, so that calls can continue using redux-multicall (and not become outdated).

Some of these changes can be pulled into a separate PR, but most are highly coupled, so it is left as a larger PR.


src/hooks/transactions/updater.tsx
```diff
@@ -79,6 +79,7 @@ export default function Updater({ pendingTransactions, onCheck, onReceipt }: Upd
         promise
           .then((receipt) => {
             if (receipt) {
+              fastForwardBlockNumber(receipt.blockNumber)
```
vm, line 82: not sure why this is necessary

zzmp: This guarantees that the token allowance (`useTokenAllowance`) is synced to at least the block where it was confirmed. Without this, its possible that `useBlockNumber` returns a block prior to the approval's mining, in which case the UI will re-prompt for approval.

It's a bug in our block number / transaction context, which is fixed by fast forwarding in the case that a confirmation is observed in a block after the current block number.

src/abis/permit2.json
```diff
@@ -0,0 +1,41 @@
+[
```
vm, line 1: can't this be exported by the permit2 sdk?

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -60,7 +60,7 @@ export default function SwapButton({
   const universalRouterSwapCallback = useUniversalRouterSwapCallback(permit2Enabled ? trade : undefined, {
     slippageTolerance: slippage.allowed,
     deadline,
-    permit: permit.signature,
+    permit: allowance.state === AllowanceState.ALLOWED ? allowance.signature : undefined,
```
vm, line 63: should this be called `permitSignature`? or allowanceSignature?

src/hooks/useTokenAllowance.ts
```diff
@@ -16,14 +16,24 @@ export function useTokenAllowance(
...
   const inputs = useMemo(() => [owner, spender], [owner, spender])
-  const { result, syncing: isSyncing } = useSingleCallResult(contract, 'allowance', inputs)
 
-  return useMemo(() => {
-    const tokenAllowance = token && result && CurrencyAmount.fromRawAmount(token, result.toString())
-    return { tokenAllowance, isSyncing }
-  }, [isSyncing, result, token])
+  // If there is no allowance, recheck next observed block.
+  // This guarantees that the tokenAllowance is marked isSyncing upon approval and updated upon being synced.
+  const [blocksPerFetch, setBlocksPerFetch] = useState<1>()
```
vm, line 23: doesn't this need to be `1 | undefined`?

src/hooks/useTokenAllowance.ts
```diff
@@ -16,14 +16,24 @@ export function useTokenAllowance(
...
   const contract = useTokenContract(token?.address, false)
-
   const inputs = useMemo(() => [owner, spender], [owner, spender])
-  const { result, syncing: isSyncing } = useSingleCallResult(contract, 'allowance', inputs)
 
-  return useMemo(() => {
-    const tokenAllowance = token && result && CurrencyAmount.fromRawAmount(token, result.toString())
-    return { tokenAllowance, isSyncing }
-  }, [isSyncing, result, token])
+  // If there is no allowance, recheck next observed block.
```
vm, line 21: `re=check on next observed block.`

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -124,7 +121,7 @@ export default function SwapButton({
       <ActionButton
         color={color}
         onClick={onClick}
-        disabled={disabled || (permit2Enabled && permit.state === PermitState.LOADING)}
+        disabled={disabled || (permit2Enabled && allowance.state === AllowanceState.UNKNOWN)}
```
vm, line 124: how can the allowance state get to unknown?

src/hooks/transactions/updater.tsx
vm: the `fastForward` existed before though. Where was it being called and why do we need to also add it here? Can we consolidate?

src/hooks/usePermit2Allowance.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }, [isApprovalPending, isApprovalSyncing])
+
+  // Signature and PermitAllowance will expire, so they should be rechecked at an interval.
+  const [now, setNow] = useState(Date.now())
+  // Calculate now such that the signature will still be valid for the submitting block.
+  useInterval(() => setNow((Date.now() - STANDARD_L1_BLOCK_TIME) / 1000), STANDARD_L1_BLOCK_TIME, true)
+
+  const [signature, setSignature] = useState<PermitSignature>()
+  const isSigned = useMemo(() => {
```
vm, line 77: this does seem worth testing as a unit test fwiw, it's pretty critical

src/hooks/usePermit2Allowance.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (!(isPermitted || isSigned)) {
+      await updatePermitAllowance()
+    }
+    return info
+  }, [isApprovalLoading, isApproved, isPermitted, isSigned, updatePermitAllowance, updateTokenAllowance])
+
+  return useMemo(() => {
+    if (token) {
+      if (!tokenAllowance || !permitAllowance) {
```
vm, line 102: personally think its better to throw here than to keep the unknown state, as an invariant

src/hooks/transactions/updater.tsx
zzmp: It _wasn't_ being called, which was a benign bug which became malignant with this change.

src/abis/permit2.json
zzmp: it can, but we'd need to modify our build process to generate types for it. I think it's simpler to add a pared down version here, in `src/abis`, which are already being pre-processed to generate types. See the `contracts:compile:abi` script.

src/hooks/useTokenAllowance.ts
zzmp: It's inferred by the type checker to be `1 | undefined` by the initial value that is (not) set.

src/components/Swap/SwapActionButton/SwapButton.tsx
zzmp: This will occur if tokenAllowance or permitAllowance have not yet been read off of the chain, or if the JSON-RPC's doing so fail. It is essentially a "loading" state.

I am fine re-naming this `LOADING` - please advise.

src/hooks/usePermit2Allowance.ts
zzmp: yes. this is flag-guarded, and tests will be coming in subsequent PRs.

src/hooks/usePermit2Allowance.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const [signature, setSignature] = useState<PermitSignature>()
+  const isSigned = useMemo(() => {
+    if (!amount || !signature) return false
+    return signature.details.token === token?.address && signature.spender === spender && signature.sigDeadline >= now
+  }, [amount, now, signature, spender, token?.address])
+
+  const { permitAllowance, expiration: permitExpiration, nonce } = usePermitAllowance(token, account, spender)
+  const updatePermitAllowance = useUpdatePermitAllowance(token, spender, nonce, setSignature)
+  const isPermitted = useMemo(() => {
```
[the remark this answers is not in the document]
zzmp: yes. this is flag-guarded, and tests will be coming in subsequent PRs.

src/hooks/usePermit2Allowance.ts
zzmp: unknown state can and will occur while the data is loading, and throwing would crash the application. this is a valid (but unknown) state, so it should not throw.

src/components/Swap/SwapActionButton/SwapButton.tsx
zzmp: I've renamed this to `LOADING` to clarify.

src/hooks/usePermit2Allowance.ts
zzmp: Renamed this to `AllowanceState.LOADING` to make this more clear (and not interpretable as an unhandled error state).

vm (approved):

src/hooks/transactions/updater.tsx
vm: oh wow lol

PR #343, opened 2022-12-28
fix: show error message when metamask disconnects

just-toby:
https://uniswaplabs.atlassian.net/browse/WEB-1771?atlOrigin=eyJpIjoiZTJmM2NiZGYwODY0NDcxZWJjNTMzOWY3MTVmYjJhYjQiLCJwIjoiaiJ9

note: depends on #345 

similar to [this PR](https://github.com/Uniswap/interface/pull/5722) in the web app, we want to show an error dialog when metamask disconnects because of an internal error. This PR uses the existing Error Dialog system to display the message, and the logic lives at the `ErrorBoundary` level - note that this error doesn't happen during render, so we can't just `throw MetamaskConnectionError()`

[WEB-1771]: https://uniswaplabs.atlassian.net/browse/WEB-1771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/hooks/web3/index.tsx
```diff
@@ -107,11 +109,12 @@ export function Provider({
...
 }
 
-const onError = (error: Error) => console.error(error)
+const onErrorFallback = (error: Error) => console.error(error)
 
 function initializeWeb3ReactConnector<T extends Connector, P extends object>(
   Constructor: { new (options: P): T },
-  options?: Omit<P, 'actions'>
+  options?: Omit<P, 'actions'>,
+  onError: (error: Error) => void = onErrorFallback
```
zzmp, line 117: This does not need a fallback, and it is already included in the options pack, so it is confusing to additionally include it separately. You can remove `onError` and `onErrorFallback` entirely.

Instead, explicitly pass `onError: console.error` in the options pack in all invocations of `initializeWeb3ReactConnector`. If you want to keep it as a fallback, then add something like this to `initializeWeb3ReactConnector:

      options: Omit<P, 'actions'>, // makes options required (this should probably be done anyway)
    ...
      options.onError = options.onError ?? console.error

src/hooks/web3/index.tsx
```diff
@@ -107,11 +109,12 @@ export function Provider({
   )
 }
 
-const onError = (error: Error) => console.error(error)
+const onErrorFallback = (error: Error) => console.error(error)
```
zzmp, line 112: This can be removed - you can log explicitly inline as you are doing.

src/hooks/web3/index.tsx
```diff
@@ -107,11 +109,12 @@ export function Provider({
...
 
 function initializeWeb3ReactConnector<T extends Connector, P extends object>(
   Constructor: { new (options: P): T },
-  options?: Omit<P, 'actions'>
+  options?: Omit<P, 'actions'>,
```
zzmp, line 116: This can be made required, as in practice it is always passed.

src/state/error.ts
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'jotai'
+
+export enum ConnectionErrorType {
+  MetamaskDisconnected = 'MetamaskDisconnected',
+}
+
+// Used to set the error type when the error occurs outside of render time and cannot be "thrown" and caught by the ErrorBoundary.
```
zzmp, line 7: I like this idea, and think it deserves its own PR.

src/state/error.ts
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'jotai'
+
+export enum ConnectionErrorType {
```
zzmp, line 3: Errors may have metadata, so I think it is best to subclass `Error` explicitly:

    class ConnectionError extends Error {}
    class MetaMaskConnectionError extends ConnectionError {}

src/state/error.ts
```diff
@@ -0,0 +1,8 @@
...
+  MetamaskDisconnected = 'MetamaskDisconnected',
+}
+
+// Used to set the error type when the error occurs outside of render time and cannot be "thrown" and caught by the ErrorBoundary.
+export const connectionErrorAtom = atom<ConnectionErrorType | undefined>(undefined)
```
zzmp, line 8: I think this should be typed as `atom<Error | undefined>()`

src/components/Widget.tsx
```diff
@@ -131,9 +131,9 @@ export function TestableWidget(props: PropsWithChildren<TestableWidgetProps>) {
           <I18nProvider locale={locale}>
             <DialogWrapper ref={setDialog} />
             <DialogProvider value={props.dialog || dialog}>
-              <ErrorBoundary onError={props.onError}>
-                <ReduxProvider store={store}>
-                  <AtomProvider initialValues={props.initialAtomValues}>
+              <ReduxProvider store={store}>
+                <AtomProvider initialValues={props.initialAtomValues}>
+                  <ErrorBoundary onError={props.onError}>
```
zzmp, line 136: This may be dangerous - if there is a problem with the atoms, it may be out of scope of the ErrorBoundary.

It is safer to implement the error state (`src/state/error.ts`) using `Context`, so that it can be independent and the `ErrorBoundary` can still be at a higher level than `Redux` and `Atom` providers.

src/components/Error/ErrorBoundary.tsx
```diff
@@ -44,3 +46,27 @@ export default class ErrorBoundary extends Component<PropsWithChildren<ErrorBoun
...
+
+/**
+ * This is a wrapper around the ErrorBoundary class component above.
+ * The class component handles regular Errors "thrown" during render.
+ *
+ * For other errors, which rely on reading from state, we need this functional component.
+ */
+export default function ErrorBoundary({ children, onError }: PropsWithChildren<ErrorBoundaryProps>) {
+  const connectionError = useAtomValue(connectionErrorAtom)
+  if (connectionError === ConnectionErrorType.MetamaskDisconnected) {
```
zzmp, line 58: I think this can be more generic so that it is more maintainable down the line - something like setting `error`/`header`/`action`/`onClick` based on the type of the error (using `instanceof`, see comments in `src/state/error.ts` about subclassing) instead of retyping the JSX for every case.

zzmp (dismissed):

src/state/error.ts
zzmp: Some prior work:

- https://github.com/Uniswap/widgets/blob/main/src/errors.ts (this can probably be removed from the codebase)
- error codes: https://github.com/Uniswap/widgets/blob/main/src/constants/eip1193.ts
- less relevant: translates known errors to readable strings: https://github.com/Uniswap/widgets/blob/main/src/utils/swapErrorToUserReadableMessage.tsx

src/hooks/web3/index.tsx
```diff
@@ -107,13 +109,11 @@ export function Provider({
...
 
-const onError = (error: Error) => console.error(error)
-
 function initializeWeb3ReactConnector<T extends Connector, P extends object>(
   Constructor: { new (options: P): T },
-  options?: Omit<P, 'actions'>
+  options: Omit<P, 'actions'>
 ): Web3ReactConnector<T> {
-  const [connector, hooks] = initializeConnector((actions) => new Constructor({ actions, onError, ...options } as P))
+  const [connector, hooks] = initializeConnector((actions) => new Constructor({ actions, ...options } as P))
```
zzmp, line 116: You should continue to explicitly pass `onError` to the other connectors, eg, on L135:

    return initializeWeb3ReactConnector(EIP1193, { provider, onError: console.error })

zzmp (changes requested):

src/hooks/web3/index.tsx
just-toby: good catch thanks

src/errors.ts
```diff
@@ -41,5 +41,6 @@ export class MetaMaskConnectionError extends ConnectionError {
       action: t`Reload`,
       message: t`'A Metamask error caused your wallet to disconnect. Reload the page to reconnect.'`,
     })
+    this.name = 'MetaMask Connection Error'
```
zzmp, line 44: I think this should reuse the name of `ConnectionError` - the details about MM are included in the `message`.
I think this should be backed out.

zzmp (approved):
Just back out the name on the metamask error subclass and then lgtm

PR #345, opened 2023-01-03
feat: create new ErrorContext for errors that are not thrown

just-toby:
based on #343  - adds a new Context to the app to track app-wide errors that are not "thrown". for example, error thrown outside of render time cannot be caught by the existing ErrorBoundary, so they should use this context instead.

the MetamaskConnectionError will be used in a new PR once this infra is approved.


src/errors.ts
```diff
@@ -4,3 +4,12 @@ export class IntegrationError extends Error {
     this.name = 'Integration Error'
   }
 }
+
+class ConnectionError extends Error {}
+
+export class MetaMaskConnectionError extends ConnectionError {
+  constructor() {
+    super('A Metamask error caused your wallet to disconnect. Reload the page to reconnect.')
```
zzmp, line 12: These errors should be translated. Will the `t` macro work here?

src/state/error.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const ErrorContext = createContext<{
+  error: Error | undefined
+  setError: (e: Error) => void
+}>({
+  error: undefined,
+  setError: () => null,
+})
+
+export const ErrorContextProvider = (props: PropsWithChildren<Record<string, unknown>>) => {
```
zzmp, line 11: Just call this `Provider`, and then alias it when importing (`import { Provider as ErrorProvider } from 'state/error'`).

zzmp, line 11: The only prop is children, so just use `(props: PropsWithChildren) => {` (with no template arg).

src/state/error.tsx
```diff
@@ -0,0 +1,23 @@
+import { createContext, PropsWithChildren, useState } from 'react'
```
zzmp, line 1: I think this logic should be rolled into the `ErrorBoundary`, as they work in tandem. I don't think you need a whole new file to accomplish it.

src/state/error.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const ErrorContext = createContext<{
```
zzmp, line 3: Instead of exporting the `ErrorContext`, export a getter/setter:

    export function useError() {
      return useContext(ErrorContext).error
    }
    
    export function useSetError() {
      return useContext(ErrorContext).setError
    }

src/components/Error/ErrorBoundary.tsx
```diff
@@ -28,19 +32,52 @@ export default class ErrorBoundary extends Component<PropsWithChildren<ErrorBoun
     this.props.onError?.(error, errorInfo)
   }
 
+  getErrorDialogHeader(error: Error): string {
+    if (error instanceof MetaMaskConnectionError) {
```
zzmp, line 36: These messages should be encoded into the errors, so it doesn't need to be duplicated. eg the Error that you extend from in src/errors.ts should have certain properties used to fill out the dialog, like

    abstract class WidgetError extends Error {
      header: string
      action: string
    }

src/components/Error/ErrorBoundary.tsx
```diff
@@ -28,19 +32,52 @@ export default class ErrorBoundary extends Component<PropsWithChildren<ErrorBoun
...
   }
 }
+
+/**
+ * This is a wrapper around the ErrorBoundary class component above.
+ * The class component handles regular Errors "thrown" during render.
+ *
+ * For other errors, which rely on reading from state, we need this functional component.
+ */
+export default function ErrorBoundary({ children, onError }: PropsWithChildren<ErrorBoundaryProps>) {
```
zzmp, line 76: I don't think this needs a wrapper. The context can be checked in the `render` function.

zzmp (changes requested):

src/errors.ts
```diff
@@ -1,6 +1,50 @@
...
     super(message)
     this.name = 'Integration Error'
   }
 }
+
+const DefaultErrorHeader = t`Please refresh the page and try again.`
+const DefaultErrorAction = t`Reload the page`
+
+export abstract class WidgetError extends Error {
+  header: string = DefaultErrorHeader
```
zzmp: nit: `DefaultErrorHeader` is a string, so no explicit typing is needed. Same with `action`.

src/errors.ts
```diff
@@ -1,6 +1,50 @@
...
+  action: string = DefaultErrorAction
+
+  constructor(config: { header?: string; action?: string; message?: string }) {
+    super(config.message)
+    if (config.header) {
```
zzmp, line 19: nit: I think this would be clearer with defaulting on one line, in which case you can type it and not call it a default:

    this.header = header ?? DefaultErrorHeader

src/errors.ts
```diff
@@ -1,27 +1,23 @@
...
-    }
+    this.header = config.header ?? DEFAULT_ERROR_HEADER
+    this.action = config.action ?? DEFAULT_ERROR_ACTION
+  }
+}
+
+export class IntegrationError extends WidgetError {
+  constructor(message: string) {
+    super({ message })
+    this.name = 'Integration Error'
```
zzmp, line 20: What does `this.name` do?

zzmp (approved):

src/errors.ts
just-toby: this was my first time seeing this, and I just followed the `IntegrationError` example - but it seems like it's just [common practice ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name) and the names are probably shown in debugging tools sometimes.

PR #346, opened 2023-01-03
feat: track widget width and add responsive styling

just-toby:
adds width tracking to the widget. this is the first diff in this workstream, so i've also included the change to `release.yaml` to disable automatic releases.

implementation:
width-tracking logic lives in `WidgetWrapper`. this was a simple styled component in `Widget`, but in this PR we're making it a standalone component to house the new logic. The width state is managed by a React Context at this level. We try to initialize the state based on the `width` prop from the integrator, but can't always rely on it because it can be a string like "100%". The general case is handled by a `ResizeObserver`.

usage:
currently we use it to add some vertical padding to the input/output components. but i've created a hook so we can access the widget width anywhere.

`const widgetWidth = useWidgetWidth()`


.github/workflows/release.yaml
```diff
@@ -1,10 +1,5 @@
 name: Release
 
-on:
-  push:
-    branches:
-      - main
-
```
JFrankfurt, line 7: was this intentional?

src/components/WidgetWrapper.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+    // We default to 360px if width is undefined
+    expect(component.getByTestId(widgetWidthValueTestId).textContent).toBe('360')
+    expect(component.getByTestId(widgetWidthTypeTestId).textContent).toBe('number')
+  })
+
+  // todo: write a test that verifies widgetWidth is correct when the width prop is a string
+  // e.g. <WidgetWrapper width="100%"> or <WidgetWrapper width="100px">
+  // This logic depends on ResizeObserver and Layout, so it must be an e2e test.
```
JFrankfurt, line 56: looks like this todo is still outstanding--do you intend to resolve it for this PR or leave it? If you're planning on leaving it, we should add a jira task to track.

src/components/WidgetWrapper.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+})
+
+export {}
```
JFrankfurt, line 59: I don't think you need to include this

.github/workflows/release.yaml
just-toby: @JFrankfurt yes, zach suggested we disable automatic release until these redesign changes are done: https://www.notion.so/uniswaplabs/Widget-UI-Updates-Eng-Plan-deeec4509ca448b796ed41caf63b2bc3

src/components/Swap/Input.tsx
```diff
@@ -28,22 +29,20 @@ const Balance = styled(ThemedText.Body2)`
   transition: color 0.25s ease-in-out;
 `
 
-const InputColumn = styled(Column)<{ disableHover?: boolean }>`
+const InputColumn = styled(Column)<{ disableHover?: boolean; widgetWidthPx: number }>`
```
zzmp, line 32: Instead of passing the `widgetWidth` directly to `InputColumn` and including more logic in the CSS, can you move the width check (`widgetWidth > 420`) to the `FieldWrapper` function?

Even better, if this can be labeled more abstractly (like `useIsWideWidget`) and the actual width is less important, can you move the logic into a hook within `hooks/useWidgetWidth`?

src/components/Swap/Output.tsx
```diff
@@ -10,12 +11,12 @@ import { FieldWrapper } from './Input'
 
 export const colorAtom = atom<string | undefined>(undefined)
 
-const OutputWrapper = styled(FieldWrapper)<{ hasColor?: boolean | null }>`
+const OutputWrapper = styled(FieldWrapper)<{ hasColor?: boolean | null; widgetWidthPx: number }>`
```
zzmp, line 14: See comment on `InputColumn` - logic should not live within the CSS.
Also note that this may be overriding the CSS for `InputColumn` (which I think is your intention).

src/components/Widget.tsx
```diff
@@ -109,13 +81,6 @@ export interface TestableWidgetProps extends WidgetProps {
...
 
 export function TestableWidget(props: PropsWithChildren<TestableWidgetProps>) {
-  const width = useMemo(() => {
-    if (props.width && props.width < 300) {
-      console.warn(`Widget width must be at least 300px (you set it to ${props.width}). Falling back to 300px.`)
-      return 300
-    }
-    return props.width ?? 360
-  }, [props.width])
   const locale = useMemo(() => {
```
zzmp, line 75: As a follow-up - in a separate PR, as this is unrelated - it would be nice to move this locale logic to the `I18nProvider` as well.

zzmp (changes requested):
Haven't reviewed the whole thing yet, but here's an initial look

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+    box-sizing: border-box;
+    font-family: ${({ theme }) => (typeof theme.fontFamily === 'string' ? theme.fontFamily : theme.fontFamily.font)};
+
+    @supports (font-variation-settings: normal) {
+      font-family: ${({ theme }) => (typeof theme.fontFamily === 'string' ? undefined : theme.fontFamily.variable)};
+    }
+  }
+`
+
+type WidgetWidthContextType = {
```
zzmp, line 36: Move the context and provider to the same file as the hook, and out of this component's file.
Doing so, you can export a setter and getter from that file without exposing the actual Context, and thus guard against improper usage by future authors.

src/hooks/useWidgetWidth.ts
```diff
@@ -0,0 +1,10 @@
+import { WidgetWidthContext } from 'components/WidgetWrapper'
+import { useContext } from 'react'
+
+/**
+ * Gets the current width of the widget.
+ */
+export default function useWidgetWidth(): number {
```
zzmp, line 7: I don't think this is actually used. Consider instead just exporting a hook called eg `useIsWidgetNarrow` or `useIsWidgetWide`, or even returning an enum here that just defines `WIDE` or `NARROW`.

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+  border-radius: ${({ theme }) => theme.borderRadius}em;
+  box-sizing: border-box;
+  color: ${({ theme }) => theme.primary};
+  display: flex;
+  flex-direction: column;
+  font-size: 16px;
+  font-smooth: always;
+  font-variant: none;
+  min-height: 360px;
+  min-width: 300px;
```
zzmp, line 20: Should there be a max-width?

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+  padding: ${HORIZONTAL_PADDING}px;
+  position: relative;
+  user-select: none;
+  width: ${({ width }) => width && (isNaN(Number(width)) ? width : `${width}px`)};
```
zzmp, line 24: Can you move this transformation to a util? Also note that the falsey case cannot occur, and should just be removed so it's guarded against by typing.

    export function toLength(length: number | string) {
      if (isNaN(Number(length)) {
        return length
      } else {
        return `${length}px`
      }
    }

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+`
+
+type WidgetWidthContextType = {
+  widgetWidth: number
+}
+
+export const WidgetWidthContext = createContext<WidgetWidthContextType>({
+  widgetWidth: 0,
```
zzmp, line 41: If this _only_ contains widgetWidth, it does not need to be an object, it can just be a number:

    const WidgetWidthContext = createContext<number>(0)

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+})
+
+interface WidgetWrapperProps {
+  width: number | string | undefined
+  className?: string | undefined
```
zzmp, line 46: Is this for styled-components? If so, you should include an inline comment: `// required for styled-components`.

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+    return props.width ?? 360
+  }, [props.width])
+
+  /**
+   * We need to manually track the width of the widget because the width prop could be a string
+   * like "100%" or "400px" instead of a number.
+   */
+  const wrapperRef = useRef<HTMLDivElement>(null)
+  const [wrapperWidth, setWidgetWidth] = useState(isNaN(Number(width)) ? 360 : (width as number))
```
zzmp, line 63: If you use a util like `toLength` (see above), then you won't need to repeat the logic for the default case here.

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+      return 300
+    }
+    return props.width ?? 360
+  }, [props.width])
+
+  /**
+   * We need to manually track the width of the widget because the width prop could be a string
+   * like "100%" or "400px" instead of a number.
+   */
+  const wrapperRef = useRef<HTMLDivElement>(null)
```
zzmp, line 62: nit: just call this `ref`, it's the only one here.

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [wrapperWidth, setWidgetWidth] = useState(isNaN(Number(width)) ? 360 : (width as number))
+  useEffect(() => {
+    const observer = new ResizeObserver((entries) => {
+      // contentRect doesn't include padding or borders
```
zzmp, line 66: Is there an alternate value that _does_ include padding?

zzmp (changes requested):

src/components/Widget.tsx
just-toby: https://github.com/Uniswap/widgets/pull/351

src/components/WidgetWrapper.tsx
just-toby: i don't think so...it seems like this is the className that we allow the integrator to set on the Widget.

src/components/WidgetWrapper.tsx
just-toby: not from resizeObserver, unfortunately

src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,88 @@
...
+      const { width } = entries[0].contentRect
+      setWidgetWidth(Math.round(width) + 2 * HORIZONTAL_PADDING)
```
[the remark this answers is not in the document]
just-toby: in a haze i thought we'd want only round numbers, but this is probably not true

src/components/WidgetWrapper.tsx
just-toby: that's not in the spec, but i'll check with fred and follow up with this if it turns out we do want it

src/hooks/useWidgetWidth.tsx
```diff
@@ -0,0 +1,20 @@
+import { createContext, PropsWithChildren, useContext } from 'react'
+
+const WidgetWidthContext = createContext<number>(0)
+
+interface WidgetWidthProviderProps {
+  width: number
+}
+
+export function WidgetWidthProvider(props: PropsWithChildren<WidgetWidthProviderProps>) {
```
zzmp, line 9: nit: I prefer to destructure: `{ width, children }: PropsWithChildren<WidgetWidthProviderProps> {`

src/hooks/useWidgetWidth.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return <WidgetWidthContext.Provider value={props.width}>{props.children}</WidgetWidthContext.Provider>
+}
+
+export function useWidgetWidth(): number {
+  return useContext(WidgetWidthContext)
+}
+
+export function useIsWideWidget(): boolean {
+  const widgetWidth = useContext(WidgetWidthContext)
```
zzmp, line 18: use `useWidgetWidth` instead of re-referring to the context:

    return useWidgetWidth() > 420

package.json
```diff
@@ -198,6 +198,7 @@
     "qs": "^6.9.4",
     "react-cosmos": "^5.6.6",
     "react-test-renderer": "^18.0.0",
+    "resize-observer-polyfill": "^1.5.1",
```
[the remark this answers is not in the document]
zzmp: You should include it in dependencies and make sure it is included in the non-test code, because ResizeObserver is not supported on Safari on iOS 12.5, which has >0.2% market share.

We define valid browsers in our package.json under browserslist, and we support all browsers with >0.2% market share (except Opera Mini and dead browsers). You can check whether an API is available under this subset of browsers using https://caniuse.com/?search=ResizeObserver (feel free to ask for a demo).

zzmp (changes requested):

src/components/WidgetWrapper.tsx
```diff
@@ -1,5 +1,6 @@
 import { WidgetWidthProvider } from 'hooks/useWidgetWidth'
 import { PropsWithChildren, useEffect, useMemo, useRef, useState } from 'react'
+import ResizeObserverPolyfill from 'resize-observer-polyfill'
```
zzmp, line 3: Import this as `ResizeObserver`, not `ResizeObserverPolyfill`.

test/setup.ts
```diff
@@ -1,4 +1,6 @@
+import ResizeObserverPolyfill from 'resize-observer-polyfill'
```
zzmp, line 1: You can back out this change, because the polyfill is now imported in the implementation file, so it is not necessary to special-case it for testing.

zzmp (changes requested):

zzmp (approved):
Rename to `ResizeObserver` and back out the changes to test/setup.ts and this lgtm!

PR #348, opened 2023-01-03
feat: styling updates to widget input step

just-toby:
design spec: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10575%3A115233&t=VFd894XSQCCzBYD0-4

updates sizings, spacings, and colors for the main widget root component, and the input/output fields. This PR focuses on the "input" step, which includes the initial view with and without tokens selected.
this also updates the colors in the theme to match the new design.

image
image

### DIFF
--- a/src/components/ActionButton.tsx
+++ b/src/components/ActionButton.tsx
@@ -56,7 +56,7 @@ const actionCss = css`
 export const Overlay = styled(Row)<{ hasAction: boolean }>`
   border-radius: ${({ theme }) => theme.borderRadius}em;
   flex-direction: row-reverse;
-  margin-top: 12px;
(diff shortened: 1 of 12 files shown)


vm (changes requested):
I think widget uses `em` in particular for some reason?

just-toby:
@vm I think we should REM since we don't have a need to be scaling specific elements of the page separately. REM is a way to enforce consistent scaling throughout the entire page.

just-toby:
screenshot of the "wide" version:
Screenshot 2023-01-05 at 12 11 47 PM

vm (approved):

PR #349, opened 2023-01-04
test: upgrade tests to use react@18

zzmp:


e2e/connect.test.tsx
```diff
@@ -14,66 +14,71 @@ const HARDHAT_ACCOUNT_DISPLAY_STRING = `${hardhat.account.address?.substring(
   6
 )}...${hardhat.account.address?.substring(hardhat.account.address.length - 4)}`
 
+declare global {
+  // eslint-disable-next-line no-var
+  var IS_REACT_ACT_ENVIRONMENT: boolean
+}
+
+beforeAll(() => {
+  // End-to-end tests do not need to run in an act environment.
```
zzmp, line 23: Without this, there is a flood of errors on the screen, because a lot of our libraries depend on async events (eg web3-react).

JFrankfurt: dare I ask if it's possible to mock that?

JFrankfurt (approved):

e2e/connect.test.tsx
zzmp: I don't think we want to mock this for e2e tests, so that it's simulating a closer-to-production environment.
It's theoretically possible, but it'd be a large ask, and probably require some refactoring to make it mockable.

PR #350, opened 2023-01-04
feat: disable auto deploys redesign

just-toby:
see the eng plan here: https://www.notion.so/uniswaplabs/Widget-UI-Updates-Eng-Plan-deeec4509ca448b796ed41caf63b2bc3

while the redesign work is ongoing, we'll pause automatic releases.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,9 +1,11 @@
 name: Release
 
 on:
-  push:
-    branches:
-      - main
+  # push:
+  #   branches:
+  #     - main
+  # manual trigger
+  workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -1,10 +1,5 @@
 name: Release
 
-on:
```
zzmp, line 3: Can you add a manual trigger for now? See https://github.com/Uniswap/interface/blob/main/.github/workflows/release.yaml#L5-L6

zzmp (changes requested):

zzmp (approved):

JFrankfurt (approved):

PR #352, opened 2023-01-04
fix: simplify async error throwing

zzmp:
Simplifies async error throwing using the method detailed in https://medium.com/trabe/catching-asynchronous-errors-in-react-using-error-boundaries-5e8a5fd7b971.

Follow-ups:
- [ ] Convert async errors to use this method.
- [ ] Convert EIP-1193 error codes (constants/eip1193.ts) to use WidgetError.
- [ ] Convert utils/swapErrorToUserReadableMessage.tsx to use WidgetError.


just-toby (approved):

just-toby:
LGTM but we should confirm this is working (with `MetamaskConnectionError`) before landing

PR #355, opened 2023-01-05
feat: connect wallet changes

just-toby:
[widget design spec](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10369%3A106894&t=SDVcX2Pi6dTwROIw-4)

in the new design, the main action button is a "Connect Wallet" button which opens the wallet modal. this PR moves the "open wallet modal" functionality from the Header down to the main action button. also, we hide the Toolbar in the wallet disconnected state.


src/components/ActionButton.tsx
```diff
@@ -88,7 +88,16 @@ export default function ActionButton({
   wrapperProps,
   ...rest
 }: ActionButtonProps) {
-  const textColor = useMemo(() => (color === 'accent' ? 'onAccent' : 'currentColor'), [color])
+  const textColor = useMemo(() => {
+    switch (color) {
+      case 'accent':
```
zzmp, line 93: This feels like a failing of the design system. How come `accentSoft` does not get an `onAccentSoft` color? Should `onAccentSoft` alias to `accent`?

src/theme/theme.ts
```diff
@@ -1,6 +1,7 @@
 export interface Colors {
   // surface
   accent: string
+  accentSoft: string
```
zzmp, line 4: Just verifying: this name comes from the design system?

zzmp (approved):

src/components/ActionButton.tsx
zzmp: Could you bring this up with Design? It doesn't need to block this PR, but I would like to make sure it's brought up.

just-toby: yeah i'll start a discussion with design about this ... i'm also confused why the widget has a much smaller color set. if it's not intentional, then i'd like to copy over the color definitions from `interface` since the new widget designs are using the same system

PR #356, opened 2023-01-06
feat: max width for widget

just-toby:
adds max width to the widget container as per the spec 

https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10905%3A109428&t=gY6rvL8clpqdZ5EE-4

### DIFF
--- a/src/components/WidgetWrapper.tsx
+++ b/src/components/WidgetWrapper.tsx
@@ -19,6 +19,7 @@ const StyledWidgetWrapper = styled.div<{ width: number | string }>`
   font-size: 16px;
   font-smooth: always;
   font-variant: none;
+  max-width: 600px;
   min-height: 360px;
   min-width: 300px;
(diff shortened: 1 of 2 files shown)


src/components/WidgetWrapper.tsx
```diff
@@ -47,6 +48,10 @@ export default function WidgetWrapper(props: PropsWithChildren<WidgetWrapperProp
       console.warn(`Widget width must be at least 300px (you set it to ${props.width}). Falling back to 300px.`)
       return 300
     }
+    if (props.width && props.width > 600) {
```
zzmp, line 51: Nest these so you don't recheck props.width:

    if (props.width) {
      if (props.width < 300) {
        ...
      } else if (...) {
        ...
      }
    }

zzmp (approved):
One nit, otherwise lgtm!

PR #357, opened 2023-01-06
feat: updates to widget toolbar

just-toby:
design spec: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10305%3A105759&t=gY6rvL8clpqdZ5EE-4

moves the Approval/Allowance UI from the ActionButton to the Toolbar. it isn't matching the design spec exactly yet, but this is where it should live. i'll update the styling in a follow up

also, removes some props from the different Button types. these components are all getting the same value which can come from a shared hook.

also note that we hide the toolbar if either the input or output currency is unset, per the design

PR #358, opened 2023-01-07
feat: update toolbar caption for loading and trade

just-toby:
- disable the main button when approval/allowance isn't set
- add a gas estimate in the toolbar
- add an basic unit test for the Toolbar component - can't test very many states without an e2e test, so i'll follow up with that later


src/components/Swap/SwapActionButton/index.tsx
```diff
@@ -1,7 +1,10 @@
 import { useWeb3React } from '@web3-react/core'
 import { SupportedChainId } from 'constants/chains'
 import { ChainError, useSwapInfo } from 'hooks/swap'
+import { SwapApprovalState } from 'hooks/swap/useSwapApproval'
 import { useIsWrap } from 'hooks/swap/useWrapCallback'
+import { AllowanceState } from 'hooks/usePermit2Allowance'
+import { usePermit2 } from 'hooks/useSyncFlags'
```
zzmp, line 7: `usePermit2 as usePermit2Enabled`

src/components/Swap/SwapActionButton/index.tsx
```diff
@@ -16,17 +19,29 @@ export default function SwapActionButton() {
...
+    approval,
+    allowance,
     trade: { trade },
   } = useSwapInfo()
+  const permit2Enabled = usePermit2()
   const isWrap = useIsWrap()
-  const isDisabled = useMemo(
-    () =>
+  const isDisabled = useMemo(() => {
+    if (permit2Enabled) {
```
zzmp, line 29: These are booleans just like the rest of the list - include them in the compound conditional statement, instead of making new expressions.

zzmp: This belongs in #357, otherwise #357 breaks this behavior.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -1,37 +1,58 @@
...
+  )
+}
+
+function GasEstimate({ gasUseEstimateUSD }: GasEstimateProp) {
+  const theme = useTheme()
+
+  return (
+    <CaptionRow gap={0.25}>
+      <GasIconWrapper color={theme.secondary} />
+      <ThemedText.Body2 color="secondary">${gasUseEstimateUSD?.toFixed(2) ?? ' –'}</ThemedText.Body2>
```
zzmp, line 54: You should be using conedison to format numbers, not bespoke logic.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -68,57 +89,62 @@ export function MissingInputs() {
...
+          </Loading>
+        }
+      />
+      <GasEstimate gasUseEstimateUSD={gasUseEstimateUSD} />
+    </>
   )
 }
 
-export function WrapCurrency({ inputCurrency, outputCurrency }: { inputCurrency: Currency; outputCurrency: Currency }) {
+interface WrapCurrencyProps {
```
zzmp, line 108: nit: Pulling this and TradeProps into separate interfaces is a choice of personal preference - I'm fine with it either way. They are not heavily used interfaces - they are both one-offs.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -68,57 +89,62 @@ export function MissingInputs() {
...
-              </ThemedText.Caption>
-              <Rule />
-            </>
-          )}
-          <RoutingDiagram trade={trade} />
-        </Column>
-      </Tooltip>
-      <Price trade={trade} outputUSDC={outputUSDC} />
+      <CaptionRow>
+        <Price trade={trade} outputUSDC={outputUSDC} />
```
zzmp, line 145: You've gutted the routing diagram without replacing it anywhere. Is that to be done in a follow-up PR? Right now it is just dead codepaths.

zzmp (changes requested):

src/components/Swap/Toolbar/Caption.tsx
just-toby: yeah, i removed it intentionally - i figured with releases disabled this wouldn't be a problem until we add it back in the new location.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -19,19 +22,44 @@ const Loading = styled.span`
...
+  height: 1em;
+  width: 1em;
+`
+
 interface CaptionProps {
   icon?: Icon
   caption: ReactNode
 }
 
+interface GasEstimateProp {
```
zzmp, line 40: `GasEstimateProps` (pluralized) to be consistent.

src/components/Swap/Toolbar/Toolbar.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  })
+
+  it('should render a request for approval', () => {
+    const component = renderComponent(
+      <SwapInfoProvider>
+        <Toolbar />
+      </SwapInfoProvider>,
+      {
+        initialAtomValues: [
+          [stateAtom, getInitialTradeState({ amount: '1000' })],
```
zzmp, line 65: nit: just '1', or this may break if the users balance is too low in a future refactor

src/components/Swap/Toolbar/Toolbar.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+describe('Toolbar', () => {
+  it('should not render without currencies specified', () => {
+    const component = renderComponent(<Toolbar />)
+    expect(component.queryByTestId('toolbar')).toBeNull()
+  })
+
+  it('should render with both currencies specified, no amount input', () => {
+    const component = renderComponent(
+      <SwapInfoProvider>
```
zzmp, line 48: Should this provider be auto-included and mocked (as an additional option) as part of `renderComponent`/`renderHook`?

src/components/Swap/Toolbar/Toolbar.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    chainId: 1,
+    account: '0x0000000000000000000000000000000000000000',
+  }),
+}))
+
+jest.mock('../../../hooks/usePermit2Allowance', () => ({
+  __esModule: true,
+  ...(jest.requireActual('../../../hooks/usePermit2Allowance') as Module),
+  default: () => ({
+    state: 1 /* AllowanceState.REQUIRED */,
```
zzmp, line 24: Can you use the actual value? It would be much less brittle. I don't know if this is possible without moving the enum to a separate file.

```
    state: requireActual('../../../hooks/usePermit2Allowance').AllowanceState.REQUIRED
```

src/components/Swap/Toolbar/Toolbar.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+jest.mock('@web3-react/core', () => ({
+  ...(jest.requireActual('@web3-react/core') as Module),
+  useWeb3React: () => ({
+    isActive: true,
+    chainId: 1,
+    account: '0x0000000000000000000000000000000000000000',
+  }),
+}))
+
+jest.mock('../../../hooks/usePermit2Allowance', () => ({
```
zzmp, line 20: Can you specify the mocked file as `'hooks/usePermit2Allowance'` (absolute filepath) instead of using a relative filepath?

zzmp (changes requested):

src/components/Swap/Toolbar/Toolbar.test.tsx
zzmp: This could be done as a follow-up. I'm more interested in whether you think this would make future tests easier to write. If so, I'd be happy to own this piece.

just-toby: i think it would make sense to move this wrapping of the UI under test into the "test infra" section of the code. we could have an API that the test writer uses to enable different providers for wrapping lower level components, if needed. although, i'm not sure if there is any harm to always wrapping ...

just-toby: i don't think it's urgent until we get to the point where we're doing the same thing in multiple test files and copy-pasting between them just to write tests

src/components/Swap/Toolbar/Toolbar.test.tsx
just-toby: yes!

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -22,40 +21,33 @@ const Loading = styled.span`
...
 }
 
-function GasEstimate({ gasUseEstimateUSD }: GasEstimateProp) {
-  const theme = useTheme()
-
+function GasEstimate({ gasUseEstimateUSD }: GasEstimateProps) {
   return (
     <CaptionRow gap={0.25}>
-      <GasIconWrapper color={theme.secondary} />
+      <Gas color={'secondary'} />
```
zzmp, line 50: `<Gas color="secondary" />` (because it is no longer dynamic).

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -145,10 +133,10 @@ export interface TradeProps {
   impact?: PriceImpact
 }
 
-export function Trade({ trade, outputUSDC, impact, gasUseEstimateUSD }: TradeProps & GasEstimateProp) {
+export function Trade({ trade, outputUSDC, impact, gasUseEstimateUSD }: TradeProps & GasEstimateProps) {
```
zzmp, line 136: `TradeProps` should just extend `GasEstimateProps` instead of using an intersection type.

zzmp (approved):
two small changes requested, then this lgtm!
no re-review necessary as long as those changes are made.

PR #361, opened 2023-01-10
feat: Convert settings into popover

tinaszheng:
This PR converts the settings menu from a slide-in dialogue to a popover menu

https://user-images.githubusercontent.com/59578595/211571182-0b35482d-f415-4fe0-8e22-46b9b68efee5.mov

Some other notes:
- removed a `Header button` css property that was effecting sizing of all child buttons - we should define these at an individual level
- followup to do mobile-friendly modal view 
- followup to add drop shadows

 Questions:
- should we add a min-width for the menu?

### DIFF
--- a/src/components/Popover.tsx
+++ b/src/components/Popover.tsx
@@ -8,7 +8,8 @@ import { Layer } from 'theme'
 
 const BoundaryContext = createContext<HTMLDivElement | null>(null)
 
-export const BoundaryProvider = BoundaryContext.Provider
+/** Defines a boundary component past which a Popover should not overflow. */
+export const PopoverBoundaryProvider = BoundaryContext.Provider
 
 const PopoverContainer = styled.div<{ show: boolean }>`
   background-color: ${({ theme }) => theme.dialog};
--- a/src/components/Popover.tsx
+++ b/src/components/Popover.tsx
@@ -85,9 +85,18 @@ export interface PopoverProps {
   placement: Placement
   offset?: number
   contained?: true
+  showArrow?: boolean
 }
 
-export default function Popover({ content, show, children, placement, offset, contained }: PopoverProps) {
+export default function Popover({
+  content,
+  show,
+  children,
+  placement,
+  offset,
+  contained,
+  showArrow = true,
+}: PopoverProps) {
   const boundary = useContext(BoundaryContext)
   const reference = useRef<HTMLDivElement>(null)
 
@@ -138,12 +147,14 @@ export default function Popover({ content, show, children, placement, offset, co
         createPortal(
           <PopoverContainer show={show} ref={setPopover} style={styles.popper} {...attributes.popper}>
             {content}
-            <Arrow
-              className={`arrow-${attributes.popper?.['data-popper-placement'] ?? ''}`}
-              ref={setArrow}
-              style={styles.arrow}
-              {...attributes.arrow}
-            />
+            {showArrow && (
+              <Arrow
+                className={`arrow-${attributes.popper?.['data-popper-placement'] ?? ''}`}
+                ref={setArrow}
+                style={styles.arrow}
+                {...attributes.arrow}
+              />
+            )}
           </PopoverContainer>,
           boundary
         )}
--- a/src/components/Popover.tsx
+++ b/src/components/Popover.tsx
@@ -8,7 +8,7 @@ import { Layer } from 'theme'
 
 const BoundaryContext = createContext<HTMLDivElement | null>(null)
 
-/** Defines a boundary component past which a Popover should not overflow. */
+/* Defines a boundary component past which a Popover should not overflow. */
 export const PopoverBoundaryProvider = BoundaryContext.Provider
 
 const PopoverContainer = styled.div<{ show: boolean }>`
--- a/src/components/Swap/Settings/index.tsx
+++ b/src/components/Swap/Settings/index.tsx
@@ -1,3 +1,4 @@
+import { useCloseOnEscape } from 'hooks/useCloseOnEscape'
 import { Settings as SettingsIcon } from 'icons'
 import { useState } from 'react'
 import styled from 'styled-components/macro'
@@ -11,6 +12,7 @@ import TransactionTtlInput from './TransactionTtlInput'
 export function SettingsPopover() {
   const [boundary, setBoundary] = useState<HTMLDivElement | null>(null)
 
+  // TODO: add back reset settings functionality
   return (
     <Column gap={1} style={{ paddingTop: '1em' }} ref={setBoundary} padded>
       <PopoverBoundaryProvider value={boundary}>
@@ -31,6 +33,8 @@ const SettingsButton = styled(IconButton)`
 export default function Settings() {
   const [open, setOpen] = useState(false)
   const [wrapper, setWrapper] = useState<HTMLDivElement | null>(null)
+
+  useCloseOnEscape(() => setOpen(false))
 
   return (
     <div ref={setWrapper}>
--- a/src/components/Swap/Settings/index.tsx
+++ b/src/components/Swap/Settings/index.tsx
@@ -1,4 +1,4 @@
-import { useCloseOnEscape } from 'hooks/useCloseOnEscape'
+import { useOnEscapeHandler } from 'hooks/useOnEscapeHandler'
 import { Settings as SettingsIcon } from 'icons'
 import { useState } from 'react'
 import styled from 'styled-components/macro'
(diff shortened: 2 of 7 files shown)


src/components/Popover.tsx
```diff
@@ -138,12 +147,14 @@ export default function Popover({ content, show, children, placement, offset, co
         createPortal(
           <PopoverContainer show={show} ref={setPopover} style={styles.popper} {...attributes.popper}>
             {content}
-            <Arrow
-              className={`arrow-${attributes.popper?.['data-popper-placement'] ?? ''}`}
-              ref={setArrow}
-              style={styles.arrow}
-              {...attributes.arrow}
-            />
+            {showArrow && (
```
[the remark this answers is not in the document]
just-toby: see `Tooltip` - that's used a bunch and seems to have the arrow

src/components/Swap/Settings/index.tsx
```diff
@@ -1,72 +1,44 @@
...
+    transform: rotate(45deg);
+    transition: transform 0.25s;
   }
 `
 
-export default function Settings({ disabled }: { disabled?: boolean }) {
+export default function Settings() {
   const [open, setOpen] = useState(false)
-  const [hover, setHover] = useState(false)
+  const [wrapper, setWrapper] = useState<HTMLDivElement | null>(null)
```
just-toby, line 33: for my sake, can you explain the benefit of using `useState` instead of `useRef` here? i have never seen it before and want to understand.

just-toby (approved):

src/components/Swap/Settings/index.tsx
tinaszheng: oo this is copied from other places where we have `BoundaryProvider`s, i think in general they behave similarly except in the `useState` version, the `wrapper` value will actually cause a rerender to `BoundaryProvider` since updates to `wrapper` are not referentially stable. havent dug too much into how these boundary providers work yet tho~

zzmp: It is so that the component will rerender once the wrapper is set ([see docs](https://reactjs.org/docs/hooks-reference.html#:~:text=Keep%20in%20mind%20that%20useRef%20doesn%E2%80%99t%20notify%20you%20when%20its%20content%20changes.%20Mutating%20the%20.current%20property%20doesn%E2%80%99t%20cause%20a%20re%2Drender.%20If%20you%20want%20to%20run%20some%20code%20when%20React%20attaches%20or%20detaches%20a%20ref%20to%20a%20DOM%20node%2C%20you%20may%20want%20to%20use%20a%20callback%20ref%20instead.)).

src/components/Swap/Settings/index.tsx
```diff
@@ -1,72 +1,44 @@
...
-      />
-      {open && (
-        <Dialog color="module" onClose={() => setOpen(false)}>
-          <SettingsDialog />
-        </Dialog>
-      )}
-    </>
+    <div ref={setWrapper}>
+      <BoundaryProvider value={wrapper}>
+        <Popover showArrow={false} offset={10} show={open} placement="top-end" content={<SettingsPopover />}>
```
zzmp, line 38: Past usage of Popover was for tooltip and tooltip-like usages, so they are missing some functionality necessary for the more Dialog-like Popover.

Specifically, the Popover should dismiss both when you hit <ESC> and when you click off of it. See the Dialog for an example implementation.

src/components/Swap/Settings/index.tsx
```diff
@@ -1,72 +1,44 @@
...
-        icon={SettingsIcon}
-      />
-      {open && (
-        <Dialog color="module" onClose={() => setOpen(false)}>
-          <SettingsDialog />
-        </Dialog>
-      )}
-    </>
+    <div ref={setWrapper}>
+      <BoundaryProvider value={wrapper}>
```
zzmp, line 37: Add doc comment for BoundaryProvider (eg `/** Defines a boundary component past which a Popover should not overflow. */`). I would even consider renaming it `PopoverBoundaryProvider`.

src/components/Swap/Settings/index.tsx
```diff
@@ -1,72 +1,44 @@
...
-  const resetSettingsBase = useResetAtom(settingsAtom)
-  const resetSettings = useCallback(() => {
-    onSettingsReset?.()
-    resetSettingsBase()
-  }, [onSettingsReset, resetSettingsBase])
+
   return (
-    <>
-      <Header title={<Trans>Settings</Trans>} ruled>
-        <TextButton onClick={resetSettings}>
```
zzmp, line 28: Double-check with @infredible RE: removing the reset functionality.
If it is to be removed, remove `onSettingsReset` entirely, including converting the underlying atom to a simpler atom (from a ResettableAtom). As-is, this leaves in a lot of dead code to support resetting settings.

zzmp (changes requested):

src/components/Swap/Settings/index.tsx
tinaszheng: seems more appropriate to convert the existing `Popover` than trying to coalesce Dialog into this new functionality though. Popover is essentially the same except you have different enter/exit states

src/components/Swap/Settings/index.tsx
tinaszheng: since this isn't unique to just the settings popover, ill add the comment to the BoundaryProvider component

src/components/Swap/Settings/index.tsx
tinaszheng: good callout, re-adding reset is being added to design backlog so can do a followup PR to add that back to the UI once its ready.

src/components/Swap/Settings/index.tsx
tinaszheng: asked in figma about dismissing behavior

src/components/Swap/Settings/index.tsx
zzmp: Great! Can you add a comment to settingsAtom noting that this will be re-added back in?

src/components/Swap/Settings/index.tsx
zzmp: Feels like terrible UX to me to only be able to dismiss the settings popover by clicking the cog icon again. I feel pretty strongly that it should retain the Dialog exit behaviors.

zzmp (changes requested):

src/components/Swap/Settings/index.tsx
zzmp: that's exactly what I'd wanted, ty

src/hooks/useCloseOnEscape.tsx
```diff
@@ -0,0 +1,11 @@
+import { useEffect } from 'react'
+
+export function useCloseOnEscape(onClose?: () => void) {
```
zzmp, line 3: nit: This doesn't necessarily close - it just triggers a callback on Escape.

Could you rename the file and all uses to `useOnEscapeHandler(onEscape?: () => void) {`?

src/components/Swap/Settings/index.tsx
```diff
@@ -1,72 +1,48 @@
...
+export function SettingsPopover() {
   const [boundary, setBoundary] = useState<HTMLDivElement | null>(null)
-  const { onSettingsReset } = useAtomValue(swapEventHandlersAtom)
-  const resetSettingsBase = useResetAtom(settingsAtom)
-  const resetSettings = useCallback(() => {
-    onSettingsReset?.()
-    resetSettingsBase()
-  }, [onSettingsReset, resetSettingsBase])
+
+  // TODO: add back reset settings functionality
```
zzmp, line 15: tag the todo with your name or (preferably) the ticket ID (eg WEB-123). This way, if it is forgotten, it will be easy to look up the corresponding ticket in the future.

src/components/Swap/Settings/index.tsx
```diff
@@ -1,72 +1,48 @@
...
 
-const SettingsButton = styled(IconButton)<{ hover: boolean }>`
-  ${SettingsIcon} {
-    transform: ${({ hover }) => hover && 'rotate(45deg)'};
-    transition: ${({ hover }) => hover && 'transform 0.25s'};
-    will-change: transform;
+const SettingsButton = styled(IconButton)`
+  ${SettingsIcon}:hover {
+    transform: rotate(45deg);
+    transition: transform 0.25s;
```
zzmp, line 29: This is a regression, as it will not rotate back onmouseleave. What if you move transition outside of the hover pseudo-selector?

zzmp (changes requested):

zzmp (approved):
approving assuming you resolve the 3 outstanding comments
if any require further discussion, please re-request review

src/components/Swap/Settings/index.tsx
tinaszheng: it doesn't currently rotate back on onmouseleave i believe? at least not when i try it on the token details page: https://app.uniswap.org/#/tokens/ethereum/0xf1b99e3e573a1a9c5e6b2ce818b617f0e664e86b

i can add it though

PR #362, opened 2023-01-10
feat: update liquidity and balance warnings in toolbar

just-toby:
- updated the `Caption` component to support colors and setting the tooltip content
- updated copy for the liquidity and balances warnings


src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -28,20 +29,32 @@ const CaptionRow = styled(Row)<{ gap: number; shrink?: number }>`
...
   gasUseEstimateUSD: CurrencyAmount<Token> | undefined
 }
 
-function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary' }: CaptionProps) {
+function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', tooltip }: CaptionProps) {
   return (
-    <CaptionRow gap={0.5}>
-      <LargeIcon icon={Icon} color={color} />
+    <CaptionRow gap={0.5} shrink={0}>
+      {Boolean(tooltip) ? (
```
zzmp, line 51: This should not need to be wrapped in a Boolean, because the tooltip will never be returned - it is a ternary.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -55,13 +68,14 @@ function GasEstimate({ gasUseEstimateUSD }: GasEstimateProps) {
         <>
           <Gas color="secondary" />
           <ThemedText.Body2 color="secondary">
-            {formatCurrencyAmount({ amount: gasUseEstimateUSD, isUsdPrice: true })}
+            {!gasUseEstimateUSD ? '-' : formatCurrencyAmount({ amount: gasUseEstimateUSD, isUsdPrice: true })}
```
zzmp, line 71: Pull this value up to a const after L64, as its shared across states.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -28,20 +29,32 @@ const CaptionRow = styled(Row)<{ gap: number; shrink?: number }>`
   height: 100%;
 `
 
+interface CaptionTooltip {
+  content: ReactNode
+  placement?: Placement
```
zzmp, line 34: Do you need to pass in placement? Will it not always be bottom, as part of the Caption?
I would rather this interface be removed, and CaptionProps just have `tooltip?: ReactNode`.

zzmp (approved):
Assuming you can make the 3 changes, lgtm!
If you have any questions about them, ask for re-review and we can discuss.

src/components/Swap/Toolbar/Caption.tsx
just-toby: yes, it's not always bottom. new design has it to the right/left

PR #363, opened 2023-01-10
feat: support opacity version of disabled ActionButton

just-toby:
the [design spec ](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=1708%3A26043&t=wBjZPXKjLmMuPj2y-0) calls for disabled buttons to change their opacity, not their color.

this adds an option to ActionButton to support that


src/components/ActionButton.tsx
```diff
@@ -7,16 +7,20 @@ import Button from './Button'
 import Row, { RowProps } from './Row'
 import Tooltip from './Tooltip'
 
-const StyledButton = styled(Button)`
+const StyledButton = styled(Button)<{ useDisabledColor?: boolean }>`
```
tinaszheng, line 10: i'd probably not prefix this variable with `use` since thats more of a hooks convention 🤔 

you could also probably just pass in the `disabledBehavior` value here directly?

just-toby: i wanted to have as little logic in the CSS as possible, so i thought a boolean would be better. i will update the name though

zzmp (commented):
deferring to @tinaszheng for this review

tinaszheng (approved):

PR #365, opened 2023-01-10
feat: bridge banners

cartcrom:
https://user-images.githubusercontent.com/39385577/211671297-03d35ba0-eb08-44d6-b7c4-0b902bc03829.mov


src/components/Widget.tsx
```diff
@@ -75,32 +78,35 @@ export default function Widget(props: PropsWithChildren<WidgetProps>) {
...
+              {
+                // UI configuration must be passed to initial atom values, or the first frame will render incorrectly.
+              }
+              <AtomProvider initialValues={useInitialFlags(props as Flags)}>
+                <WidgetUpdater {...props} />
+                <Web3Provider {...(props as Web3Props)}>
+                  <BlockNumberProvider>
+                    <ThemeProvider theme={props.theme}>
+                      <WidgetContainer width={props.width} className={props.className}>
+                        <WidgetWrapper>
```
cartcrom, line 95: Needed to add container that didn't include background/border so that banner could be under widget rather than inside

src/components/BridgeBanner.tsx
```diff
@@ -0,0 +1,81 @@
...
+`
+const BridgeLink = styled(ExternalLink)<{ color: string }>`
+  align-items: center;
+  border-radius: 0.5em;
+  color: ${({ color }) => color};
+  display: flex;
+  font-size: 1em;
+  justify-content: space-between;
+  padding: 0.75em;
+  text-decoration: none !important;
```
tinaszheng, line 28: `!important` should be used as a last resort 🤔 is there another way u can set this property, maybe by making the selector more specific?

src/components/BridgeBanner.tsx
```diff
@@ -0,0 +1,81 @@
...
+export function BridgeBanner() {
+  const { chainId } = useWeb3React()
+
+  if (!isSupportedChainId(chainId)) {
+    return null
+  }
+
+  const { label, logoUrl, bridge, color, backgroundColor } = getChainInfo(chainId)
+
+  return bridge && color && backgroundColor ? (
```
tinaszheng, line 65: i would also use a guard-style like return here:

```
if (!bridge || !color || !backgroundColor) return null

return (...)
```

src/components/WidgetContainer.tsx
```diff
@@ -0,0 +1,86 @@
...
+    })
+    const current = ref.current
+    if (current) {
+      observer.observe(ref.current)
+    }
+    return () => {
+      if (current) {
+        observer.unobserve(current)
+      }
+    }
```
tinaszheng, line 78:
```suggestion
    return () => current && observer.unobserve(current)
```

src/cosmos/Swap.fixture.tsx
```diff
@@ -64,6 +64,7 @@ function Fixture() {
   const [defaultOutputAmount] = useValue('defaultOutputAmount', { defaultValue: 0 })
 
   const [hideConnectionUI] = useValue('hideConnectionUI', { defaultValue: false })
+  const [hideL2BridgeBanner] = useValue('hideL2BridgeBanner', { defaultValue: false })
```
tinaszheng, line 67: thoughts on making it hidden by default 🤔 thinking of external integrators who have defined custom wrapper styles and hopefully they wont get thrown off by upgrading

src/components/BridgeBanner.tsx
cartcrom: was leftover from copying from interface, the !important seems to not actually be _imporant_ here so I'm removing 👍

tinaszheng: sweeet

src/components/WidgetContainer.tsx
cartcrom: @tinaszheng I can't inline that way because useEffect cleanups can't return null, it has to be void or destructor...

do you think...

```
return () => (current ? observer.unobserve(current) : undefined)
```
or
```
return () => {
    current && observer.unobserve(current)
}
```

is better?

src/cosmos/Swap.fixture.tsx
cartcrom: hmmmm I'm not sure... that makes sense to me, but then I also feel like integrators might not realize this functionality is there unless it's on by default? I'm not sure who should make that decision wdyt

src/components/WidgetContainer.tsx
tinaszheng: on mobile we have a rule against calling functions inside ternaries since it has caused a dumb bug in the past 😅 i think latter is better for that reason

cartcrom: sounds good

src/cosmos/Swap.fixture.tsx
tinaszheng: okie we can leave as is for now and change it later if needed

tinaszheng (approved):
image

just-toby (approved):

PR #366, opened 2023-01-10
feat: toolbar updates for price impacts

just-toby:
- use a new Caption type for price impacts
- update colors for price impact messages
- fix bug with displaying price impact below the output field

image

image

image

image

### DIFF
--- a/src/components/Swap/Toolbar/Caption.tsx
+++ b/src/components/Swap/Toolbar/Caption.tsx
@@ -1,9 +1,7 @@
 import { Trans } from '@lingui/macro'
 import { Placement } from '@popperjs/core'
 import { Currency, CurrencyAmount, Token } from '@uniswap/sdk-core'
-import Column from 'components/Column'
 import Row from 'components/Row'
-import Rule from 'components/Rule'
 import Tooltip from 'components/Tooltip'
 import { loadingCss } from 'css/loading'
 import { PriceImpact } from 'hooks/usePriceImpact'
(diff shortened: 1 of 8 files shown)


zzmp (changes requested):

src/components/Swap/Toolbar/index.tsx
```diff
@@ -71,8 +71,10 @@ export default memo(function Toolbar() {
         return <Caption.InsufficientLiquidity />
       }
       if (trade?.inputAmount && trade.outputAmount) {
-        return (
-          <Caption.Trade trade={trade} outputUSDC={outputUSDC} impact={impact} gasUseEstimateUSD={gasUseEstimateUSD} />
+        return impact ? (
+          <Caption.PriceImpactWarning impact={impact} trade={trade} />
```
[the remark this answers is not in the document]
just-toby: thanks, good catch. i think i can rely on `impact.warning` here since it's doing those checks.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -183,32 +179,44 @@ export function WrapCurrency({ inputCurrency, outputCurrency, gasUseEstimateUSD
...
         tooltip={{
+          content: <RoutingDiagram trade={trade} />,
+        }}
+      />
+      <GasEstimate gasUseEstimateUSD={gasUseEstimateUSD} />
+    </>
+  )
+}
+
+type PriceImpactWarningProps = Omit<TradeProps, 'gasUseEstimateUSD'> & { impact: PriceImpact }
```
zzmp, line 199: nit: It'd also be appropriate here to combine types everywhere, eg:

    export interface TradeProps {...}
    export function Trade({...}: TradeProps & GasEstimateProps) {
    export interface PriceImpactProps extends TradeProps {...}

I'd prefer this to using Omit, because it's less complex overall. wdyt?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -183,32 +179,44 @@ export function WrapCurrency({ inputCurrency, outputCurrency, gasUseEstimateUSD
...
+
+export function PriceImpactWarning({ impact }: PriceImpactWarningProps) {
```
zzmp, line 201: I'd rather this just be called `PriceImpact`, considering the warning is checked in Toolbar/index.tsx, not here.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -62,9 +60,7 @@ function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', too
 
 function GasEstimate({ gasUseEstimateUSD }: GasEstimateProps) {
   const isWideWidget = useIsWideWidget()
-  const displayEstimate = !gasUseEstimateUSD
-    ? '-'
-    : formatCurrencyAmount({ amount: gasUseEstimateUSD, isUsdPrice: true })
+  const displayEstimate = formatCurrencyAmount(gasUseEstimateUSD, NumberType.FiatGasPrice)
```
zzmp, line 63: This feels like a dangerous change to make in the same PR. Are you able to move it to its own PR, once conedison lands?
What does formatCurrencyAmount return if !gasUseEstimateUSD?

zzmp (changes requested):
The only real blockers are verifying the impact color in the Output section, and potentially separating the GasEstimate change. Otherwise just minor nits - looks great!

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -196,9 +204,11 @@ export function Trade({ trade, outputUSDC, gasUseEstimateUSD }: TradeProps) {
   )
 }
 
-type PriceImpactWarningProps = Omit<TradeProps, 'gasUseEstimateUSD'> & { impact: PriceImpact }
+interface PriceImpactProps {
+  impact: PriceImpactValue
+}
 
-export function PriceImpactWarning({ impact }: PriceImpactWarningProps) {
+export function PriceImpact({ impact }: TradeProps & PriceImpactProps) {
```
zzmp, line 211: You don't need TradeProps here at all. You don't need to back out the changes to WrapCurrencyProps or TradeProps, though.

nit: WrapCurrency could just be called Wrap (similarly, WrapCurrencyProps->WrapProps).

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -1,12 +1,11 @@
...
 import { Placement } from '@popperjs/core'
+import { formatPriceImpact } from '@uniswap/conedison/format'
 import { Currency, CurrencyAmount, Token } from '@uniswap/sdk-core'
-import Column from 'components/Column'
 import Row from 'components/Row'
-import Rule from 'components/Rule'
 import Tooltip from 'components/Tooltip'
 import { loadingCss } from 'css/loading'
-import { PriceImpact } from 'hooks/usePriceImpact'
+import { PriceImpact as PriceImpactValue } from 'hooks/usePriceImpact'
```
zzmp, line 8: This isn't a concrete value, it's a type. What about `IPriceImpact` (`I` for Interface), or `PriceImpactType`?

zzmp (changes requested):

src/components/Swap/Toolbar/Caption.tsx
just-toby: thanks. names...

tinaszheng (approved):
v nice

PR #367, opened 2023-01-11
test: unit testing around permit2

zzmp:
Adds unit testing for the permit2 hooks and they're dependencies (eg block fast-forwarding through the transactions updater).


zzmp:
This PR is very large, so I've broken it into more easily reviewable chunks:
#368, #369, #370, #371, #372, #373

PR #368, opened 2023-01-11
test: fast forward on future receipts

zzmp:


src/hooks/web3/index.tsx
```diff
@@ -43,7 +43,12 @@ export interface ProviderProps {
 }
 
 export function TestableProvider({ provider, children }: PropsWithChildren<{ provider: JsonRpcProvider }>) {
-  const connectors = useMemo(() => [initializeWeb3ReactConnector(JsonRpcConnector, { provider })], [provider])
+  const connector = useMemo(() => initializeWeb3ReactConnector(JsonRpcConnector, { provider }), [provider])
+  useEffect(() => {
+    connector[0].activate()
+  }, [connector])
```
just-toby, line 49: just checking, we want this change in the product code too? (not just the test code)

no mention of this change in the description and i'm not sure how it relates to the test, so i want to double check

just-toby (approved):

src/hooks/web3/index.tsx
zzmp: This activates the hardhat connection so that unit tests have a functional web3 connection. Without this the unit tests dependent on a functional web3 connection will not work.
This is already done in production code, but was not being done in test code - here it is a testing-only change.

PR #373, opened 2023-01-11
test: permit2 unit tests

zzmp:


src/hooks/usePermit2Allowance.ts
```diff
@@ -70,7 +70,7 @@ export default function usePermit2Allowance(amount?: CurrencyAmount<Token>, spen
   // Signature and PermitAllowance will expire, so they should be rechecked at an interval.
   const [now, setNow] = useState(Date.now())
   // Calculate now such that the signature will still be valid for the submitting block.
-  useInterval(() => setNow((Date.now() - STANDARD_L1_BLOCK_TIME) / 1000), STANDARD_L1_BLOCK_TIME, true)
+  useInterval(() => setNow((Date.now() + STANDARD_L1_BLOCK_TIME) / 1000), STANDARD_L1_BLOCK_TIME, true)
```
vm, line 73: what's going on with this change?

src/hooks/usePermit2Allowance.test.ts
```diff
@@ -0,0 +1,268 @@
...
+  })
+
+  describe('with no token', () => {
+    it('defaults to allowed', () => {
+      const { result } = renderHook(() => usePermit2Allowance())
+      expect(result.current).toMatchObject({ state: AllowanceState.ALLOWED })
+    })
+  })
+
+  it('loads', async () => {
```
vm, line 46: need better description here

src/hooks/usePermit2Allowance.test.ts
```diff
@@ -0,0 +1,268 @@
...
+    const { result } = renderHook(() => usePermit2Allowance(AMOUNT, SPENDER))
+    await waitFor(() => expect(useTokenAllowance).toHaveBeenCalledWith(TOKEN, addressMatcher, PERMIT2_ADDRESS))
+    expect(usePermitAllowance).toHaveBeenCalledWith(TOKEN, addressMatcher, SPENDER)
+    expect(result.current).toMatchObject({ state: AllowanceState.LOADING })
+  })
+
+  describe('approves', () => {
```
vm, line 53: better desc

vm: should each of these should be separate tests? otherwise the entire test will fail vs seeing exactly what went wrong.

src/hooks/usePermit2Allowance.ts
zzmp: It's the one bug I caught in my tests 🤣 

"still be valid for the submitting block" means we need to pretend we're in the future, not in the past.

vm: huge, worth it then

src/hooks/usePermit2Allowance.test.ts
zzmp: these _are_ the separate tests. they track the different starting states for approval, and how that settles.
they all deal with intermediate state, so I don't think I can break them down any further.

zzmp: I split these a bit more, but they still test intermediate states, so some will remain longer.

vm (approved):

PR #375, opened 2023-01-11
feat: centered header row title

cartcrom:
Part of [WEB-2712](https://uniswaplabs.atlassian.net/browse/WEB-2712), center's the title on the header component. Also simplifies the wrapping of the component and removes redundant properties.

Also removed unnused "recent transactions" page so that I could simplify the header further

new:
image

old:
image

[WEB-2712]: https://uniswaplabs.atlassian.net/browse/WEB-2712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Dialog.tsx
```diff
@@ -67,37 +64,46 @@ export function Provider({ value, children }: ProviderProps) {
 const OnCloseContext = createContext<(() => void) | undefined>(undefined)
 
 const HeaderRow = styled(Row)`
```
just-toby, line 66: should this be

```suggestion
const HeaderRow = styled(Row)<{iconSize?: number}>`
```

? i am new to styled components, so i'm not sure if we should always use the prop types

src/components/Dialog.tsx
```diff
@@ -67,37 +64,46 @@ export function Provider({ value, children }: ProviderProps) {
...
+  :hover {
+    cursor: pointer;
+    opacity: 0.6;
+  }
+`
+
+const Title = styled.div`
+  left: 50%;
+  position: absolute;
+  top: 50%;
```
just-toby, line 90: i could see why we want absolute positioning on the horizontal axis, but should we stick with flex / `align-items: center` for vertical spacing here?

just-toby: nit / non blocking

just-toby (approved):

src/components/Dialog.tsx
cartcrom: excellent point, will do ty

PR #378, opened 2023-01-11
feat: update review swap page

cartcrom:
new:
image
old:
Screen Shot 2023-01-13 at 3 43 22 PM


src/components/Swap/Summary/Details.tsx
```diff
@@ -1,23 +1,47 @@
...
+  color: ${({ theme }) => theme.secondary};
+  margin-right: 0.25em;
+`
 const Value = styled.span<{ color?: Color }>`
   color: ${({ color, theme }) => color && theme[color]};
   white-space: nowrap;
 `
+const TokenAmount = styled(Text)<{ widgetWidth: number }>`
+  color: ${({ theme }) => theme.primary};
+  font-size: ${({ widgetWidth }) => (widgetWidth < 400 ? (widgetWidth < 350 ? '24px' : '30px') : '36px')};
```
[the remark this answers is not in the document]
just-toby: this is more of a nit from me, but @zzmp has set the precedent of keeping business logic out of CSS definitions so I want to try to uphold that. he can chime in here if i'm getting this wrong

src/components/Swap/Summary/Details.tsx
```diff
@@ -27,61 +51,109 @@ interface DetailProps {
...
+  usdcAmount?: CurrencyAmount<Currency>
+}
+
+function Amount({ tooltipText, label, amount, usdcAmount }: AmountProps) {
+  const width = useWidgetWidth()
+
+  let formattedAmount = formatCurrencyAmount(amount, NumberType.TokenTx)
+  if (formattedAmount.length > 9) {
+    formattedAmount =
+      width >= 420
```
just-toby, line 76: use `useIsWideWidget` hook here

src/components/Swap/Summary/Details.tsx
```diff
@@ -27,61 +51,109 @@ interface DetailProps {
...
+        ? formatCurrencyAmount(amount, NumberType.SwapTradeAmount)
+        : formatCurrencyAmount(amount, NumberType.TokenNonTx)
+  }
+
+  return (
+    <Row flex align="space-between">
```
just-toby, line 82: i don't think `space-between` is a valid value for the `align-items` attribute. are you sure this is having any effect?

just-toby: also, space-between is the default value for the `justify` prop, so you may not need this

src/components/Swap/Price.tsx
```diff
@@ -35,16 +39,25 @@ export default function Price({ trade, outputUSDC }: PriceProps) {
     }
   }, [base, executionPrice, inputAmount, outputAmount, outputUSDC])
 
+  return [
+    `${1} ${price.baseCurrency.symbol} = ${formatPrice(price).substring(1)} ${price.quoteCurrency.symbol}`,
+    usdcPrice && formatCurrencyAmount(usdcPrice, NumberType.FiatTokenPrice),
+  ]
```
just-toby, line 45: is there an opportunity to memoize here ? not sure if it's important, but it is calling to formatting functions which do some work that we don't want to duplicate

just-toby (changes requested):

src/components/Swap/Summary/Details.tsx
cartcrom: it feels wrong to me to import both useWidgetWidth and useIsWideWidget

src/components/Swap/Price.tsx
cartcrom: oooh good point ty

src/components/Swap/Summary/Details.tsx
cartcrom: good to know, ty

src/components/Swap/Summary/Details.tsx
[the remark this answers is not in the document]
cartcrom: great point, will move into component attributes instead of css

just-toby (approved):
lgtm

src/components/Swap/Summary/Details.tsx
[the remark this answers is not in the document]
zzmp: whee!

src/components/Swap/Summary/Details.tsx
```diff
@@ -27,61 +45,111 @@ interface DetailProps {
...
+      const localizedMaxSent = formatCurrencyAmount(trade.maximumAmountIn(slippage.allowed), NumberType.TokenTx)
+      const maxSentString = `${localizedMaxSent} ${inputCurrency.symbol}`
+      estimationMessage = t`Output is estimated. You will send at most ${maxSentString} or the transaction will revert.`
+      rows.push([t`Maximum sent`, maxSentString])
     }
 
-    rows.push([t`Slippage tolerance`, `${slippage.allowed.toFixed(2)}%`, slippage.warning])
-
-    return rows
+    return [rows, estimationMessage]
```
zzmp, line 150: nit: I find it clearer to return an object, so you can destructure and retain the names; vs an array, where names can change when destructuring and similar types may be accidentally swapped.

zzmp (changes requested):

src/components/Dialog.tsx
```diff
@@ -104,12 +104,13 @@ export function Header({ title }: HeaderProps) {
 
 export const Modal = styled.div<{ color: Color }>`
   background-color: ${({ color, theme }) => theme[color]};
-  border-radius: ${({ theme }) => theme.borderRadius * 0.75}em;
+  border-radius: ${({ theme }) => theme.borderRadius}em;
```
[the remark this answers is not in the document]
cartcrom: The only dialog that still has the old-style borders in figma are transaction error dialogs, I'm checking with fred on whether that's intentional or if it can be removed from all.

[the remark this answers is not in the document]
cartcrom: Fred has confirmed all dialogs should be changed 👍

src/components/Swap/Price.tsx
```diff
@@ -12,12 +12,16 @@ interface PriceProps {
...
+  base?: 'input' | 'output'
+}
 
-  const [base, setBase] = useState<'input' | 'output'>('input')
-  const onClick = useCallback(() => setBase((base) => (base === 'input' ? 'output' : 'input')), [])
+export function useTradeExchangeRate({
+  trade,
+  outputUSDC,
+  base = 'input',
+}: ExchangeRateProps): [string, string | undefined] {
```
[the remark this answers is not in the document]
cartcrom: I decided to just completely decouple useTradeExchangeRate from PriceProps

src/components/Swap/Price.tsx
```diff
@@ -35,16 +39,28 @@ export default function Price({ trade, outputUSDC }: PriceProps) {
     }
   }, [base, executionPrice, inputAmount, outputAmount, outputUSDC])
 
+  return useMemo(
+    () => [
+      `${1} ${price.baseCurrency.symbol} = ${formatPrice(price).substring(1)} ${price.quoteCurrency.symbol}`,
```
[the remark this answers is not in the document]
cartcrom: What do you mean by this exactly?

src/components/Swap/Price.tsx
```diff
@@ -35,16 +39,28 @@ export default function Price({ trade, outputUSDC }: PriceProps) {
...
+      usdcPrice && formatCurrencyAmount(usdcPrice, NumberType.FiatTokenPrice),
+    ],
+    [price, usdcPrice]
+  )
+}
+
+/** Displays the price of a trade. If outputUSDC is included, also displays the unit price. */
+export default function Price({ trade, outputUSDC }: PriceProps) {
+  const [base, setBase] = useState<'input' | 'output'>('input')
```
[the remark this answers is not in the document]
cartcrom: agree, will do

src/components/Widget.tsx
```diff
@@ -38,14 +38,16 @@ const slideOutRight = keyframes`
 `
 
 export const DialogWrapper = styled.div`
-  border-radius: ${({ theme }) => theme.borderRadius * 0.75}em;
-  height: calc(100% - 0.5em);
+  border: ${({ theme }) => `1px solid ${theme.outline}`};
```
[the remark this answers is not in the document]
cartcrom: yes

src/theme/breakpoints.ts
```diff
@@ -0,0 +1,6 @@
+export enum BREAKPOINTS {
+  WIDE = 420,
+  MEDIUM = 400,
+  SMALL = 375,
+  EXTRA_SMALL = 350,
+}
```
cartcrom, line 6: @zzmp is this naming convention bad? should I follow interface norms and do  L M S XS

zzmp: You should follow interface norms. (I think this is fine, but if interface already has defined norms you should use those for consistency.)

cartcrom: Since the interface naming convention is used for screen width, I think it actually could make sense to use a different naming convention here for widget width

src/components/Swap/Summary/Details.tsx
```diff
@@ -27,61 +40,114 @@ interface DetailProps {
...
+  const width = useWidgetWidth()
+  const [amountFontSize, amountLineHeight] =
+    width < WIDGET_BREAKPOINTS.MEDIUM
+      ? width < WIDGET_BREAKPOINTS.EXTRA_SMALL
+        ? ['24px', '30px']
+        : ['30px', '36px']
+      : ['36px', '44px']
+
+  let formattedAmount = formatCurrencyAmount(amount, NumberType.TokenTx)
+  if (formattedAmount.length > 9) {
```
tinaszheng, line 69: can u separate the `9` out into a constant

src/components/Swap/Price.tsx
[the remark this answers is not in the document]
tinaszheng: @cartcrom why do u need to remove the first char 👀

tinaszheng (approved):

tinaszheng (approved):

PR #379, opened 2023-01-11
fix: improve routing experience

zzmp:
- Omits rate-limited polygon RPC endpoints
- Fixes useTimeout to only trigger a callback once, after the timeout
- Fixes useRouterTrade to avoid infinite retries when erring


just-toby (approved):
lgtm

PR #380, opened 2023-01-12
lint: type multicall usage

zzmp:
Adds typings for multicall hooks to improve multicall ergonomics and reduce the chance of misuse (eg ReferenceErrors).
Although functional, this is a POC - it is better to move this into redux-multicall itself so it can be useful for interface and other consumers as well.

### DIFF
--- a/src/hooks/multicall.ts
+++ b/src/hooks/multicall.ts
@@ -1,4 +1,7 @@
+import { Contract } from '@ethersproject/contracts'
+import { CallState as CallStateBase } from '@uniswap/redux-multicall'
 import { useWeb3React } from '@web3-react/core'
+import { Interface } from 'ethers/lib/utils'
 import useBlockNumber from 'hooks/useBlockNumber'
 import multicall from 'state/multicall'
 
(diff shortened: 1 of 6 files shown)


src/hooks/multicall.ts
```diff
@@ -7,33 +10,53 @@ export { NEVER_RELOAD } from '@uniswap/redux-multicall' // re-export for conveni
 
 // Create wrappers for hooks so consumers don't need to get latest block themselves
 
-type MulticallParams<T extends (chainId: number | undefined, latestBlock: number | undefined, ...args: any) => any> =
-  Parameters<T> extends [any, any, ...infer P] ? P : never
+interface CallState<C extends Contract, M extends keyof C['functions']> extends Omit<CallStateBase, 'result'> {
+  result: C['functions'][M] extends (...args: any) => Promise<infer T> ? T | undefined : never
+}
+
+type MultipleContractParams<
```
zzmp, line 17: Note that `*ContractParams` is actually lossy, because it loses the names/documentation around the parameters and only retains their typing. Arguably, this should be explicit so that names are retained for a better devX. wdyt?

src/hooks/multicall.ts
```diff
@@ -1,4 +1,7 @@
+import { Contract } from '@ethersproject/contracts'
+import { CallState as CallStateBase } from '@uniswap/redux-multicall'
 import { useWeb3React } from '@web3-react/core'
+import { Interface } from 'ethers/lib/utils'
```
zzmp, line 4: This should be @ethersproject/contract-interfaces or something similar, _not_ from the ethers umbrella package

src/hooks/useArgentWalletContract.ts
```diff
@@ -8,9 +8,9 @@ import useIsArgentWallet from './useIsArgentWallet'
 export function useArgentWalletContract(): ArgentWalletContract | null {
```
zzmp, line 8: This type is now implied, and can be removed from the function definition.

PR #381, opened 2023-01-12
feat: Add bottom sheet modal for mobile settings view

tinaszheng:
fixes WEB-2755

This PR adds a custom bottom sheet modal that swipes in/out for mobile-friendly settings views

https://user-images.githubusercontent.com/59578595/212125169-131c3d58-1d5e-471c-afa3-41f75f010320.mov

| Light mode | Dark mode |
|------------|-------------|
|image|image|

Followups:
- Investigate how to freeze the screen behind the modal when it's open (WEB-2765)
- Add support for dismissing the modal when the user clicks on the backdrop (WEB-2767)
- Move timing amounts (like the `0.25s`) into design tokens?


src/components/BottomSheetModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+  open: boolean
+  title: string
+}>
+
+export function BottomSheetModal({ children, onClose, open, title }: BottomSheetModalProps) {
+  const [rootWrapper, setRootWrapper] = useState<HTMLDivElement | null>(null)
+
+  return (
+    <>
+      <RootElement ref={setRootWrapper} open={open} />
```
just-toby, line 82: is it correct for this to be self-closing? if so, why does the RootElement expect `children` below?

tinaszheng: `RootElement` itself doesnt get unmounted since it acts as an always-anchor for the modal to attach to. i also don't dismount the backdrop to enable the "exit" animations through css

just-toby: nice

just-toby (approved):
lgtm

tinaszheng:
waiting on next build to merge cc @zzmp

src/hooks/useIsMobileWidth.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+
+const MOBILE_BREAKPOINT_WIDTH = 900
```
zzmp, line 3: breakpoints should go in a theme file so they stay consistent throughout the codebase.
Can you move this into a new src/theme/breakpoints.ts file? It can be an enum imo, but up to you (eg `enum BREAKPOINTS { MOBILE = 900 }`).

src/css/font.ts
```diff
@@ -0,0 +1,18 @@
+import { css } from 'styled-components/macro'
+
+export const globalFontStyles = css`
```
zzmp, line 3: nit: these are the only font styles, they can just be called `font` or `fontStyles`.

src/components/Swap/Settings/index.tsx
```diff
@@ -41,13 +43,23 @@ export default function Settings() {
...
     <div ref={setWrapper}>
-      <PopoverBoundaryProvider value={wrapper}>
-        <Popover showArrow={false} offset={10} show={open} placement="top-end" content={<SettingsPopover />}>
+      {isMobile ? (
+        <>
           <SettingsButton onClick={() => setOpen(!open)} icon={SettingsIcon} />
-        </Popover>
-      </PopoverBoundaryProvider>
+          <BottomSheetModal title="Settings" onClose={() => setOpen(false)} open={open}>
+            <SettingsPopover />
```
zzmp, line 53: nit: Should this now be named something more generic than `SettingsPopover`?

src/components/BottomSheetModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+    visibility: hidden;
+  }
+
+  top: 0;
+  transition: visibility 0s linear 0s, opacity 0.25s;
+  visibility: visible;
+`
+
+const Wrapper = styled.div<{ open: boolean }>`
+  ${globalFontStyles};
```
zzmp, line 41: Move this to the Dialog's wrapping element, so that it covers all cases of portaled dialogs.

src/components/BottomSheetModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+`
+
+type BottomSheetModalProps = PropsWithChildren<{
+  onClose: () => void
+  open: boolean
+  title: string
+}>
+
+export function BottomSheetModal({ children, onClose, open, title }: BottomSheetModalProps) {
+  const [rootWrapper, setRootWrapper] = useState<HTMLDivElement | null>(null)
```
zzmp, line 78: nit: call this `ref` or `root` or `rootElement` to match the actual component name.

zzmp (changes requested):
Big changes to make:
- move fonts to Dialog to cover a more general case
- move breakpoint to src/theme/breakpoints.ts

Everything else is a nit. This looks great, and BottomSheetModal is much less intrusive than I'd expected :D
Requesting changes because of the two requests above, but if they're resolved with no questions and I'm away just discard my review.

src/css/font.ts
tinaszheng: for imported font styles i like to specify that these are indeed the global font styles to help discern if they're, like, specific to this component or not

src/components/Swap/Settings/index.tsx
tinaszheng: this is specific to settings tho!

tinaszheng: i just understood what you meant lol. renamed in a diff PR!

PR #382, opened 2023-01-12
fix: Remove spin state on un-hovering settings icon

tinaszheng:
[slack](https://uniswapteam.slack.com/archives/C04JP2P3F7T/p1673506938838779?thread_ts=1673506038.511189&cid=C04JP2P3F7T) convo for context

### DIFF
--- a/src/components/Swap/Settings/index.tsx
+++ b/src/components/Swap/Settings/index.tsx
@@ -24,13 +24,9 @@ export function SettingsPopover() {
 }
 
 const SettingsButton = styled(IconButton)`
+  // don't rotate back when un-hovering because it works weirdly with modal backdrop opening on top
   ${SettingsIcon}:hover {
     transform: rotate(45deg);
-    transition: transform 0.25s;
-  }
-
-  ${SettingsIcon} {
-    transform: rotate(0);
     transition: transform 0.25s;
   }
 `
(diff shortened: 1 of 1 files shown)


just-toby (approved):

src/components/Swap/Settings/index.tsx
```diff
@@ -24,15 +24,11 @@ export function SettingsPopover() {
 }
 
 const SettingsButton = styled(IconButton)`
+  // don't rotate back when un-hovering because it works weirdly with modal backdrop opening on top
```
zzmp, line 27: be more specific here, so it's less likely someone will "unwind" this change

zzmp (changes requested):

src/components/Swap/Settings/index.tsx
zzmp: also, use full sentence/grammar in your comment

zzmp: `// Don't rotate back when un-hovering so that clicking (and losing hover due to the modal backdrop) doesn't cause unintentional back-rotation.`?

zzmp (approved):
assuming you change the comment to be more descriptive, this looks great.

src/components/Swap/Settings/index.tsx
tinaszheng: ugh the gen z in me is crying

PR #383, opened 2023-01-12
feat: expanded toolbar

just-toby:
design spec: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10575%3A112105&t=kN2DYO2XVWBwgUwg-0#343932931

- implement expanding toolbar with TradeSummary and OrderRouting details appearing below the existing Caption when expanded
- move the location of the RoutingDiagram tooltip based on the design spec
- add 2 new components, TradeSummary and OrderRouting, that appear in the expanded state ... with some tests


src/components/Swap/Toolbar/ToolbarOrderRouting.tsx
```diff
@@ -0,0 +1,39 @@
...
+  justify-content: space-between;
+  margin: 0.75em 1em;
+`
+
+interface ToolbarOrderRoutingProps {
+  trade?: InterfaceTrade
+}
+
+export default function ToolbarOrderRouting({ trade }: ToolbarOrderRoutingProps) {
+  const hopCount = trade?.routes?.reduce((acc, curr) => acc + (curr?.pools?.length ?? 0), 0)
```
just-toby, line 21: to reviewer: please confirm that this is the correct value to show

src/components/Swap/Summary/index.tsx
```diff
@@ -64,7 +64,7 @@ interface EstimateProps {
   trade: InterfaceTrade
 }
 
-function Estimate({ trade, slippage }: EstimateProps) {
+export function Estimate({ trade, slippage }: EstimateProps) {
```
tinaszheng, line 67: since this is exported now maybe name this something more specific, like `SwapEstimate` 🤔

src/components/Swap/Toolbar/ToolbarTradeSummary.tsx
```diff
@@ -0,0 +1,119 @@
...
+  slippage: Slippage
+}
+
+export default function ToolbarTradeSummary({ gasUseEstimateUSD, impact, trade, slippage }: ToolbarTradeSummaryProps) {
+  const currencySymbol = trade?.outputAmount?.currency?.symbol ?? ''
+  return (
+    <TradeSummaryColumn gap={0.75}>
+      <SummaryRow
+        name={t`Network fee`}
+        value={(gasUseEstimateUSD ? '~' : '') + formatCurrencyAmount(gasUseEstimateUSD, NumberType.FiatGasPrice)}
```
tinaszheng, line 84: i prefer string interpolation for merging strings like `${gasUseEstimateUSD ? '~' : ''}${formatCurrencyAmount(gasUseEstimateUSD, NumberType.FiatGasPrice)}` mostly just for consistency!

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -59,24 +76,27 @@ function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', too
...
 }
 
-function GasEstimate({ gasUseEstimateUSD }: GasEstimateProps) {
+function GasEstimate({ gasUseEstimateUSD, trade }: GasEstimateProps) {
   const isWideWidget = useIsWideWidget()
   // TODO(just-toby): use formatCurrencyAmount from conedison
   const displayEstimate = !gasUseEstimateUSD
     ? '-'
     : formatCurrencyAmount({ amount: gasUseEstimateUSD, isUsdPrice: true })
+  if (gasUseEstimateUSD == null) {
```
tinaszheng, line 85: you could move this line to right after line 80 for a minorrr perf optimization

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -59,24 +76,27 @@ function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', too
...
+    return null
+  }
   return (
     <CaptionRow gap={0.25}>
-      {isWideWidget ? (
-        <>
+      <>
+        {trade ? (
```
tinaszheng, line 91: what happens here if its a narrow size widget? it should display the gas fee in a tooltip right?

just-toby: i was displaying the gas fee text in a tooltip, but the design actually prefers to show that information under the routing diagram, so i am replacing my label with the RoutingDiagram. I still need to update the RoutingDiagram component to match the design

tinaszheng:
questionnn: is the expanding component supposed to animate down? thats my assumption based on existing interface expander 🤔 looks like the current ver just jumps down? i think the arrow button is supposed to rotate too

tinaszheng:
also tiny nit, it looks like the "Auto Router" label isnt vertically centered. you might have to add a display: flex to the auto router wrapper:
image

src/test/utils.ts
```diff
@@ -0,0 +1,49 @@
+import { CurrencyAmount, Token } from '@uniswap/sdk-core'
+import { FeeAmount, Pool, Route as RouteV3 } from '@uniswap/v3-sdk'
+
+// helper function to make amounts more readable
+export const amount = (raw: TemplateStringsArray) => (parseInt(raw[0]) * 1e6).toString()
+
+export const USDC = new Token(1, '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48', 6, 'USDC')
```
zzmp, line 7: these tokens should already exist in constants/tokens.ts.
Instead of recreating, could you alias them here?

src/components/Swap/Toolbar/ToolbarOrderRouting.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { InterfaceTrade } from 'state/routing/types'
+import { renderComponent } from 'test'
+import { buildMultiV3Route, buildSingleV3Route, DAI, USDC } from 'test/utils'
+
+import ToolbarOrderRouting from './ToolbarOrderRouting'
+
+describe('ToolbarOrderRouting', () => {
+  it('renders correctly, loading', () => {
+    const component = renderComponent(<ToolbarOrderRouting />)
+    expect(component.getByText('Order routing')).toBeTruthy()
```
zzmp, line 11: These tests don't check anything but the title. Is that intentional?
Would this be suited for a snapshot test?

src/components/Swap/Toolbar/index.tsx
```diff
@@ -63,7 +74,7 @@ export default memo(function Toolbar() {
           <Caption.Wrap
             inputCurrency={inputCurrency}
             outputCurrency={outputCurrency}
-            gasUseEstimateUSD={gasUseEstimateUSD}
+            gasUseEstimateUSD={expanded ? null : gasUseEstimateUSD}
```
zzmp, line 77: Will this ever be expanded?

src/components/Swap/Toolbar/index.tsx
```diff
@@ -51,7 +62,7 @@ export default memo(function Toolbar() {
     }
 
     if (state === TradeState.LOADING) {
-      return <Caption.LoadingTrade gasUseEstimateUSD={gasUseEstimateUSD} />
+      return <Caption.LoadingTrade gasUseEstimateUSD={expanded ? null : gasUseEstimateUSD} />
```
zzmp, line 65: Will this ever be expanded?

src/components/Swap/Toolbar/index.tsx
```diff
@@ -113,8 +139,21 @@ export default memo(function Toolbar() {
...
 
   return (
-    <ToolbarRow flex justify="space-between" data-testid="toolbar">
-      {caption}
-    </ToolbarRow>
+    <ToolbarColumn>
+      <ToolbarRow flex justify="space-between" data-testid="toolbar">
+        {caption}
+      </ToolbarRow>
+      {expanded && (
```
zzmp, line 146: Shouldn't expanded only be available for a valid trade?

src/components/Swap/Toolbar/ToolbarTradeSummary.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const TradeAttributeValue = styled(ThemedText.Body2)`
+  color: ${({ theme, color }) => color ?? theme.primary};
+`
+
+interface SummaryRowProps {
+  name: string
+  value: string
+  color?: Color
+  tooltip?: {
```
zzmp, line 37: Do you think separating this out into `nameTooltip` and `valueTooltip` would be more readable? I'm unsure.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -28,26 +28,43 @@ const CaptionRow = styled(Row)<{ gap: number; shrink?: number }>`
...
-  gasUseEstimateUSD: CurrencyAmount<Token> | undefined
+  trade?: InterfaceTrade
+  gasUseEstimateUSD?: CurrencyAmount<Token> | null
 }
 
 function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', tooltip }: CaptionProps) {
   return (
     <CaptionRow gap={0.5} shrink={0}>
-      {tooltip ? (
+      {Icon == null ? null : tooltip ? (
```
zzmp, line 67: use `===`, never `==`.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -59,24 +76,27 @@ function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', too
   )
 }
 
-function GasEstimate({ gasUseEstimateUSD }: GasEstimateProps) {
+function GasEstimate({ gasUseEstimateUSD, trade }: GasEstimateProps) {
   const isWideWidget = useIsWideWidget()
+  if (gasUseEstimateUSD == null) {
```
zzmp, line 81: `===`.
Also, this is a guard statement, so feel free to one-line it: `if (!gasUseEstimateUSD) return null`.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -59,24 +76,27 @@ function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', too
...
+    return null
+  }
   // TODO(just-toby): use formatCurrencyAmount from conedison
```
zzmp, line 84: rm comment?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -156,7 +176,7 @@ interface WrapProps {
   outputCurrency: Currency
 }
 
-export function Wrap({ inputCurrency, outputCurrency, gasUseEstimateUSD }: WrapProps & GasEstimateProps) {
+export function Wrap({ inputCurrency, outputCurrency, gasUseEstimateUSD, trade }: WrapProps & GasEstimateProps) {
```
zzmp, line 179: This should not need `trade`.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -199,7 +231,13 @@ interface PriceImpactProps {
   impact: PriceImpactType
 }
 
-export function PriceImpact({ impact }: PriceImpactProps) {
+export const PriceImpactWarningTooltipContent = () => (
```
zzmp, line 234: nit: I prefer this inline.

src/components/Swap/Toolbar/Caption.tsx
zzmp: when is this new condition triggered? it seems confusing - is it necessary?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -59,24 +76,27 @@ function Caption({ icon: Icon = AlertTriangle, caption, color = 'secondary', too
   )
 }
 
-function GasEstimate({ gasUseEstimateUSD }: GasEstimateProps) {
+function GasEstimate({ gasUseEstimateUSD, trade }: GasEstimateProps) {
```
zzmp, line 79: This has become more than just the `GasEstimate`. It should be renamed. `TradeTooltip`?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -134,7 +154,7 @@ export function MissingInputs() {
   return <Caption icon={Info} caption={<Trans>Enter an amount</Trans>} />
 }
 
-export function LoadingTrade({ gasUseEstimateUSD }: GasEstimateProps) {
+export function LoadingTrade({ gasUseEstimateUSD, trade }: GasEstimateProps) {
```
zzmp, line 157: There is necessarily _not_ a trade here, as it is still loading.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -180,17 +200,29 @@ export interface TradeProps {
   outputUSDC?: CurrencyAmount<Currency>
 }
 
-export function Trade({ trade, outputUSDC, gasUseEstimateUSD }: TradeProps & GasEstimateProps) {
+interface ExpandProps {
+  expanded: boolean
+  toggleExpanded: () => void
```
zzmp, line 205: This is a handler, so it should be called `on*`. `onExpand` or `onToggleExpand`?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -180,17 +200,29 @@ export interface TradeProps {
   outputUSDC?: CurrencyAmount<Currency>
 }
 
-export function Trade({ trade, outputUSDC, gasUseEstimateUSD }: TradeProps & GasEstimateProps) {
+interface ExpandProps {
+  expanded: boolean
```
zzmp, line 204: nit: this is often called `open` in other instances of expandos in this code base.

src/components/Swap/Toolbar/index.tsx
```diff
@@ -72,9 +83,23 @@ export default memo(function Toolbar() {
       }
       if (trade?.inputAmount && trade.outputAmount) {
         return impact?.warning ? (
-          <Caption.PriceImpact impact={impact} />
+          <Caption.PriceImpact
+            impact={impact}
+            expanded={expanded}
+            toggleExpanded={() => {
```
zzmp, line 89: If you alias this to a callback you can make it a one-liner:

    const onExpand = useCallback(() => {
      setExpanded((expanded) => !expanded)
    }, [])
    ...
    onToggleExpand={onToggleExpand}

src/components/Swap/Toolbar/index.tsx
```diff
@@ -72,9 +83,23 @@ export default memo(function Toolbar() {
...
+              setExpanded(!expanded)
```
zzmp, line 90: Do not ever pass an old value into a state setter, as it risks render loops.
Instead, use a setter function to retrieve the old value. In this case:

    setExpanded((expanded) => !expanded)
    
The reason to not set using the old value is that, in a hook, it could look like this, triggering rerenders because it depends on what it is setting: `useEffect(() => setExpanded(!expanded), [expanded])`.

zzmp (changes requested):

src/components/Swap/Toolbar/index.tsx
just-toby: this is out of date now - but i don't think we should assume that we need a valid trade to show the summary / routing info components - they should be able to handle an invalid trade and show a pending state

src/test/utils.ts
just-toby: i think we have these new definitions on purpose actually - we want versions of the tokens without the `name` field set, because the v3 SDK doesn't set that field. I didn't write these lines, i just copied them over to this util file

src/components/Swap/Toolbar/index.tsx
tinaszheng: ^ eh i disagree here, `expanded` is a boolean primitive value and there's no race condition with state setting here, so this feels unnecessary

src/components/Swap/Toolbar/index.tsx
tinaszheng: would recommend not using `useCallback` here as it adds unnecessary memory overhead whereas creating the handler function on every render is rather cheap. `useMemo` and `useCallback` should really be reserved for either expensive calculations or keeping props referentially stable

[here's](https://www.developerway.com/posts/how-to-use-memo-use-callback) a decent blog post that talks more about it

src/components/Swap/Toolbar/Caption.tsx
tinaszheng: oo got it, can you add a TODO in the code or add a "followup item" in the PR description just so we know for posterity that this PR needs the quick follow to be feature complete :3

src/components/Swap/Toolbar/ToolbarTradeSummary.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+interface ToolbarTradeSummaryProps {
+  rows: SummaryRowProps[]
+}
+
+export default function ToolbarTradeSummary({ rows }: ToolbarTradeSummaryProps) {
+  return (
+    <TradeSummaryColumn gap={SUMMARY_COLUMN_GAP_EM}>
+      {rows.map((row, i) => (
+        <SummaryRow key={i} {...row} />
```
tinaszheng, line 77: would recommend using `row.name` here instead of `i` as the key because if you have different number of items from rerenders (like if a middle row disappears), then react wouldn't properly know which rows have actually disappeared

might not be totally applicable for this case if `rows` never changes but good practice to try to not use indices as keys if you can :P

src/components/Swap/Toolbar/index.tsx
```diff
@@ -94,10 +117,52 @@ export default memo(function Toolbar() {
...
     isWrap,
     trade,
-    outputUSDC,
     impact,
+    open,
+    onExpand,
+    outputUSDC,
   ])
 
+  const tradeSummaryRows: SummaryRowProps[] = useMemo(() => {
```
tinaszheng, line 126: this is an example of when it's a good idea to use `useMemo` :D

src/components/Swap/Toolbar/index.tsx
```diff
@@ -94,10 +117,52 @@ export default memo(function Toolbar() {
...
+    const currencySymbol = trade?.outputAmount?.currency.symbol ?? ''
+    const rows: SummaryRowProps[] = []
+    rows.push({
```
tinaszheng, line 129: seems cleaner to just code them in the array directly instead of doing `.push` 3 times 🤔

src/components/Swap/Toolbar/index.tsx
```diff
@@ -94,10 +117,52 @@ export default memo(function Toolbar() {
...
+        ? {
+            icon: Info,
+            content: <SwapInputOutputEstimate trade={trade} slippage={slippage} />,
+          }
+        : undefined,
+    })
+    return rows
+  }, [gasUseEstimateUSD, impact?.percent, impact?.warning, slippage, trade])
+
+  const expandedHeight = useMemo(() => {
```
tinaszheng, line 161: can you leave a comment above this one to explain what the logic for calculating height is? my brain turns off when seeing this code lol

src/test/utils.ts
tinaszheng: would it break tests to use the version with `name` set? 🤔 feels fine to use the existing ones as is

tinaszheng (approved):
mostly looks great just a few nits!

will defer to zach for his comments but looks good to me

src/test/utils.ts
just-toby: yes, that causes the tests to fail

PR #385, opened 2023-01-12
feat: update colors and icon in token select dialog

just-toby:
design spec: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10369%3A107306&t=kN2DYO2XVWBwgUwg-4

update the colors used in the token select dialog, and add a search icon to the search bar

image

image

### DIFF
--- a/src/components/TokenSelect/index.tsx
+++ b/src/components/TokenSelect/index.tsx
@@ -1,7 +1,7 @@
 import { t, Trans } from '@lingui/macro'
 import { Currency } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
-import { StringInput } from 'components/Input'
+import { inputCss, StringInput } from 'components/Input'
 import { useConditionalHandler } from 'hooks/useConditionalHandler'
 import { useCurrencyBalances } from 'hooks/useCurrencyBalance'
 import useNativeCurrency from 'hooks/useNativeCurrency'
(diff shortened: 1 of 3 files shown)


src/components/TokenSelect/index.tsx
```diff
@@ -13,16 +15,27 @@ import { ThemedText } from 'theme'
...
+  border-radius: ${({ theme }) => theme.borderRadius}em;
+  justify-content: flex-start;
+  padding: calc(0.75em - 1px);
+
+  &:focus-within {
+    border-color: ${({ theme }) => theme.accentSoft};
+  }
+`
+
+const StyledSearchIcon = styled(Search)`
```
zzmp, line 37: This shouldn't be necessary. Instead, use it directly as `<SearchIcon color="secondary" />`.
If that doesn't work, change the wrapping in src/icons/index.tsx so that the color is applied correctly.

src/components/TokenSelect/index.tsx
```diff
@@ -13,16 +15,27 @@ import { ThemedText } from 'theme'
...
 import Rule from '../Rule'
 import NoTokensAvailableOnNetwork from './NoTokensAvailableOnNetwork'
 import TokenButton from './TokenButton'
 import TokenOptions, { TokenOptionsHandle } from './TokenOptions'
 import TokenOptionsSkeleton from './TokenOptionsSkeleton'
 
-const SearchInput = styled(StringInput)`
-  ${inputCss}
+const SearchInputContainer = styled(Row)`
+  background-color: ${({ theme }) => theme.module};
```
zzmp, line 26: Why do you depart from inputCss? Instead, should you be changing inputCss?
This is used for other inputs, so it should be used to keep them consistent.

zzmp (changes requested):

src/components/TokenSelect/index.tsx
just-toby: yeah works for me

zzmp (approved):

PR #387, opened 2023-01-12
feat: common bases

just-toby:
design spec: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10575%3A119214&t=kN2DYO2XVWBwgUwg-4

- adds common tokens to the top of the list in the CurrencySelect dialog
- currently using an existing list of "common bases" - will update to include UNI (to match the spec) in a future PR

### DIFF
--- a/src/components/TokenSelect/CommonBases.test.tsx
+++ b/src/components/TokenSelect/CommonBases.test.tsx
@@ -1,6 +1,7 @@
 import { SupportedChainId } from '@uniswap/sdk-core'
 import { BASES_TO_CHECK_TRADES_AGAINST } from 'constants/routing'
-import { renderComponent } from 'test'
+import { USDC_MAINNET } from 'constants/tokens'
+import { renderComponent, userEvent, waitFor } from 'test'
 
 import CommonBases from './CommonBases'
 
@@ -13,11 +14,22 @@ describe('CommonBases', () => {
     })
   }
 
-  it('calls the select function', () => {
+  it('calls the select function on click', () => {
     const callback = jest.fn()
     const component = renderComponent(<CommonBases chainId={SupportedChainId.MAINNET} onSelect={callback} />)
     component.getByText('USDC').click()
-    expect(callback).toHaveBeenCalled()
+    expect(callback).toHaveBeenCalledWith(USDC_MAINNET)
+  })
+
+  it('calls the select function on enter', async () => {
+    const user = userEvent.setup()
+    const callback = jest.fn()
+    const component = renderComponent(<CommonBases chainId={SupportedChainId.MAINNET} onSelect={callback} />)
+    await user.tab()
+    await user.type(component.container, '{enter}')
+    waitFor(() => {
+      expect(callback).toHaveBeenCalledWith(USDC_MAINNET)
+    })
   })
 
   it('renders correct currencies, mainnet', () => {
--- a/src/components/TokenSelect/CommonBases.test.tsx
+++ b/src/components/TokenSelect/CommonBases.test.tsx
@@ -1,0 +1,42 @@
+import { SupportedChainId } from '@uniswap/sdk-core'
+import { BASES_TO_CHECK_TRADES_AGAINST } from 'constants/routing'
+import { renderComponent } from 'test'
+
+import CommonBases from './CommonBases'
+
+describe('CommonBases', () => {
+  const itRendersCorrectCurrenciesForChainId = (chainId: SupportedChainId) => {
+    const component = renderComponent(<CommonBases chainId={chainId} onSelect={jest.fn()} />)
+    const tokens = BASES_TO_CHECK_TRADES_AGAINST[chainId]
+    tokens.forEach((token) => {
+      expect(component.getAllByText(token.symbol as string)).toBeTruthy()
+    })
+  }
+
+  it('calls the select function', () => {
+    const callback = jest.fn()
+    const component = renderComponent(<CommonBases chainId={SupportedChainId.MAINNET} onSelect={callback} />)
+    component.getByText('USDC').click()
+    expect(callback).toHaveBeenCalled()
+  })
+
+  it('renders correct currencies, mainnet', () => {
+    itRendersCorrectCurrenciesForChainId(SupportedChainId.MAINNET)
+  })
+
+  it('renders correct currencies, optimism', () => {
+    itRendersCorrectCurrenciesForChainId(SupportedChainId.OPTIMISM)
+  })
+
+  it('renders correct currencies, arbitrum', () => {
+    itRendersCorrectCurrenciesForChainId(SupportedChainId.ARBITRUM_ONE)
+  })
+
+  it('renders correct currencies, polygon', () => {
+    itRendersCorrectCurrenciesForChainId(SupportedChainId.POLYGON)
+  })
+
+  it('renders correct currencies, celo', () => {
+    itRendersCorrectCurrenciesForChainId(SupportedChainId.CELO)
+  })
+})
--- a/src/components/TokenSelect/CommonBases.tsx
+++ b/src/components/TokenSelect/CommonBases.tsx
@@ -56,14 +56,7 @@ export default function CommonBases({
             flex
             tabIndex={0}
             data-testid={`common-base-${currency.symbol}`}
-            onKeyPress={
-              !isSelected
-                ? (e) => {
-                    console.log({ e })
-                    e.key === 'Enter' && onSelect(currency)
-                  }
-                : undefined
-            }
+            onKeyPress={!isSelected ? (e) => e.key === 'Enter' && onSelect(currency) : undefined}
             onClick={!isSelected ? () => onSelect(currency) : undefined}
             active={isSelected}
             key={currencyId(currency)}
--- a/src/components/TokenSelect/CommonBases.tsx
+++ b/src/components/TokenSelect/CommonBases.tsx
@@ -1,0 +1,66 @@
+import { Currency } from '@uniswap/sdk-core'
+import Row from 'components/Row'
+import TokenImg from 'components/TokenImg'
+import { BASES_TO_CHECK_TRADES_AGAINST } from 'constants/routing'
+import { transparentize } from 'polished'
+import styled from 'styled-components/macro'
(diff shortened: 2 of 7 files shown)


src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+  margin-right: 0.5em;
+`
+
+export default function CommonBases({
+  chainId,
+  onSelect,
+  selectedCurrency,
+}: {
+  chainId?: number
+  selectedCurrency?: Currency | null
```
zzmp, line 41: `selectedCurrency`/`onSelect` are asymmetrical.
Ether use `selected`/`onSelect` (my preference) or `selectedCurrency`/`onSelectCurrency`.

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onSelect: (currency: Currency) => void
+}) {
+  const bases = typeof chainId !== 'undefined' ? BASES_TO_CHECK_TRADES_AGAINST[chainId] ?? [] : []
```
zzmp, line 44: Instead of a ternary, I think it'd be more readable to use a guard which would then narrow the type:

    if (!chainId) return null
    
You could expand on this through to your other ternary:

    if (!chainId) return null
    const bases = BASES_TO_CHECK_TRADES_AGAINST[chainId]
    if (bases.length === 0) return null

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return bases.length > 0 ? (
+    <BasesContainer gap={0.5} flex justify="start">
+      {bases.map((currency: Currency) => {
+        const isSelected = selectedCurrency?.equals(currency)
+
+        return (
+          <BaseWrapper
+            tabIndex={0}
+            onKeyPress={(e) => !isSelected && e.key === 'Enter' && onSelect(currency)}
```
zzmp, line 54: The listener should only be passed if isSelected, as adding too many listeners can reduce page performance:

    onKeyPress={isSelected ? ({key}) => key === 'Enter' && onSelect(currency) : undefined}

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+            onClick={() => !isSelected && onSelect(currency)}
```
zzmp, line 55: The listener should only be passed if !isSelected (see other comment).

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+    background-color: ${({ theme }) => transparentize(0.76, theme.active)};
+    color: ${({ theme }) => theme.active};
+    cursor: pointer;
+  }
+
+  padding: 0.5em 0.75em 0.5em;
+`
+
+const StyledTokenImg = styled(TokenImg)`
+  margin-right: 0.5em;
```
zzmp, line 32: Instead of modifying TokenImg can you add a gap in the BaseWrapper, so there are less bespoke elements?

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+  background-color: ${({ theme, active }) => (active ? transparentize(0.76, theme.active) : 'transparent')};
+  border: 1px solid ${({ theme, active }) => (active ? theme.active : theme.outline)};
+  border-radius: 1em;
+  color: ${({ theme, active }) => (active ? theme.active : theme.primary)};
+
+  display: flex;
+  :hover {
+    background-color: ${({ theme }) => transparentize(0.76, theme.active)};
+    color: ${({ theme }) => theme.active};
+    cursor: pointer;
```
zzmp, line 25: `cursor` should be moved up, as it will only apply on `:hover` anyways.

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  padding: 0.5em 0.75em 0.5em;
```
zzmp, line 28: mv `padding` and `display` to be grouped with other properties, before `:hover`.

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { ThemedText } from 'theme'
+import { currencyId } from 'utils/currencyId'
+
+const BasesContainer = styled(Row)`
+  margin: 0 1.25em;
+`
+
+const BaseWrapper = styled.div<{ active?: boolean }>`
+  align-items: center;
+  background-color: ${({ theme, active }) => (active ? transparentize(0.76, theme.active) : 'transparent')};
```
zzmp, line 16: Why transparentize? Is there not an appropriate theme color?

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,66 @@
...
+  border: 1px solid ${({ theme, active }) => (active ? theme.active : theme.outline)};
+  border-radius: 1em;
+  color: ${({ theme, active }) => (active ? theme.active : theme.primary)};
+
+  display: flex;
+  :hover {
```
zzmp, line 22: I think this should be `:hover, :focus` so that tabbing into an element (where the 'Enter' keypress is applicable) also shows the same UI.

src/components/TokenSelect/CommonBases.tsx
zzmp, line 16: Would it be more readable if you took the ternary properties, which match `:hover`, and put them in a separate css macro, so that you don't need to write it out twice? Then you can just use `${({ active }) => active && activeCss};`.

src/components/TokenSelect/CommonBases.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+describe('CommonBases', () => {
+  const itRendersCorrectCurrenciesForChainId = (chainId: SupportedChainId) => {
+    const component = renderComponent(<CommonBases chainId={chainId} onSelect={jest.fn()} />)
+    const tokens = BASES_TO_CHECK_TRADES_AGAINST[chainId]
+    tokens.forEach((token) => {
+      expect(component.getAllByText(token.symbol as string)).toBeTruthy()
+    })
+  }
+
+  it('calls the select function', () => {
```
zzmp, line 16: You should also test this for the keypress.

zzmp (changes requested):

src/components/TokenSelect/CommonBases.tsx
just-toby: actually this should be added as a theme color

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,81 @@
...
+  border: 1px solid ${({ theme }) => theme.outline};
+  border-radius: 1em;
+  color: ${({ theme, active }) => (active ? theme.active : theme.primary)};
+  cursor: pointer;
+  padding: 0.5em 0.75em 0.5em 0.5em;
+
+  ${({ active }) => active && activeCss};
+
+  :hover,
+  :focus {
```
zzmp, line 29: Should this be focus-within, like in inputCss? Test by tabbing.

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,81 @@
...
+        const isSelected = selected?.equals(currency)
+        return (
+          <BaseWrapper
+            flex
+            tabIndex={0}
+            data-testid={`common-base-${currency.symbol}`}
+            onKeyPress={
+              !isSelected
+                ? (e) => {
+                    console.log({ e })
```
zzmp, line 62: rm

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,81 @@
...
+                  }
+                : undefined
+            }
+            onClick={!isSelected ? () => onSelect(currency) : undefined}
+            active={isSelected}
+            key={currencyId(currency)}
+            gap={0.25}
+          >
+            <TokenImg token={currency} />
+            <ThemedText.ButtonMedium fontWeight={600} lineHeight="1.25em">
```
zzmp, line 73: Should this be a new button text type?

zzmp (changes requested):

src/components/TokenSelect/CommonBases.tsx
just-toby: no, tabbing is working for me like this

src/components/TokenSelect/CommonBases.tsx
just-toby: i don't think so, because in the figma it's labeled as "button medium" too. i think the entire design system definitions (mostly colors) is due for an update, we'll need to follow up on that

src/components/TokenSelect/CommonBases.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <BasesContainer gap={0.5} flex justify="start">
+      {bases.map((currency: Currency) => {
+        const isSelected = selected?.equals(currency)
+        return (
+          <BaseWrapper
+            flex
+            tabIndex={0}
+            data-testid={`common-base-${currency.symbol}`}
+            onKeyPress={!isSelected ? (e) => e.key === 'Enter' && onSelect(currency) : undefined}
```
tinaszheng, line 59: i'd separate this callback into a separate function for readability :P

src/theme/index.tsx
```diff
@@ -21,6 +21,7 @@ export const brand = brandLight
 
 const stateColors = {
   active: 'hsl(221, 96%, 64%)',
+  activeSoft: transparentize(0.76, 'hsl(221, 96%, 64%)'),
```
tinaszheng, line 24: wait similar comment as zach here, can this be done with an alpha value in the `hsl` or do u need transparentize

tinaszheng (approved):
mostly looks good, just one q about the transparent color but non blocking

PR #390, opened 2023-01-13
fix: polish design issues in widget

just-toby:
- add missing border to settings input
- use 'primary' text color for disable buttons
- fix casing for input sub header

https://www.notion.so/uniswaplabs/widget-design-feedbacks-835bfad85aac4ad3a00550fcea18bf99

design spec: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10305%3A105762&t=GMFoRG5hj2E6QPEd-0

image

image

### DIFF
--- a/src/components/Swap/Settings/TransactionTtlInput.tsx
+++ b/src/components/Swap/Settings/TransactionTtlInput.tsx
@@ -11,6 +11,7 @@ import { Label } from './components'
 
 const Input = styled(Row)`
   ${inputCss}
+  border: 1px solid ${({ theme }) => theme.outline};
 `
 
 export default function TransactionTtlInput() {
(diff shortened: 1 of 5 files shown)


src/components/Swap/Settings/TransactionTtlInput.tsx
```diff
@@ -11,6 +11,7 @@ import { Label } from './components'
 
 const Input = styled(Row)`
   ${inputCss}
+  border: 1px solid ${({ theme }) => theme.outline};
```
zzmp, line 14: Why does the existing container border not suffice?

zzmp, line 14: This should be backed out. Instead:
- Change the inputCss to use theme.outline for the border value. This will fix this issue, and proactively add a border around the token select search input (also spec'ed out in figma).
- In optionCss, on line 9, change theme.outline to `''` to avoid overwriting this, so that they stay in sync should inputCss change.

zzmp (changes requested):

zzmp (approved):

PR #400, opened 2023-01-13
revert: bridge banners

zzmp:
Reverts #389, #384, and #365, as they introduced a bug rendering errors by nesting the DialogProvider inside of the ErrorBoundary (instead of vice versa).


just-toby (approved):
works for me

PR #401, opened 2023-01-13
fix: resize skeleton

zzmp:
Updates the skeleton to match the redesigns for a width of 360px.

### DIFF
--- a/src/components/Swap/Skeleton.tsx
+++ b/src/components/Swap/Skeleton.tsx
@@ -10,7 +10,6 @@ import ReverseButton from './ReverseButton'
 const LoadingWrapper = styled.div`
   display: flex;
   flex-direction: column;
-  height: 100%;
   justify-content: space-between;
 `
 const Blob = styled.div<{ height: string; width: string; radius?: number; isModule?: boolean }>`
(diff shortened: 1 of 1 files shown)


src/components/Swap/Skeleton.tsx
```diff
@@ -37,44 +36,25 @@ const InputColumn = styled(Column)`
...
-  width: 100%;
+const OutputColumn = styled(InputColumn)`
+  padding-bottom: 3em;
+  padding-top: 3.5em;
 `
 
 const ButtonColumn = styled(Column)`
   padding-bottom: 0em;
-  padding-top: 0.75em;
+  padding-top: 0.55em;
```
just-toby, line 49: should this be 0.5em?

just-toby (approved):

src/components/Swap/Skeleton.tsx
zzmp: This needs to be 0.55em or it will be off by 1px.

zzmp: It's hard to see unless you disable the brandedFooter.

PR #405, opened 2023-01-17
feat: update ui for tx status dialog

just-toby:
design for this screen: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10369%3A106419&t=ijfa7EEMSPInWNSl-4

Screenshot 2023-01-17 at 9 52 19 AM
Screenshot 2023-01-17 at 9 52 34 AM

updates the fonts and placement of the UI elements on the "waiting for confirmation" and "success" screens

### DIFF
--- a/src/theme/type.tsx
+++ b/src/theme/type.tsx
@@ -40,6 +40,12 @@ export function H3(props: TextProps) {
   )
 }
 
+export function H4(props: TextProps) {
+  return (
+    <TextWrapper className="headline headline-4" fontSize={20} fontWeight={500} lineHeight="28px" noWrap {...props} />
+  )
+}
+
 export function Subhead1(props: TextProps) {
   return (
     <TextWrapper className="subhead subhead-1" fontSize={16} fontWeight={500} lineHeight="24px" noWrap {...props} />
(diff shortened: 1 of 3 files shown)


src/theme/type.tsx
```diff
@@ -40,6 +40,12 @@ export function H3(props: TextProps) {
   )
 }
 
+export function H4(props: TextProps) {
+  return (
+    <TextWrapper className="headline headline-4" fontSize={20} fontWeight={500} lineHeight="28px" noWrap {...props} />
```
tinaszheng, line 45: is there a mismatch between figma and our design tokens? i see a `headlineSmall` in the design spec but these numbers do indeed match up 👀

tinaszheng (approved):

src/theme/type.tsx
just-toby: yes, it seems like our ThemedText needs an audit. [i have a task for it here](https://uniswaplabs.atlassian.net/browse/WEB-2782?atlOrigin=eyJpIjoiMTQ0MzBlNjY4N2JhNGNjYzgzMjMyNWM2NjUyODBhOGMiLCJwIjoiaiJ9)

PR #407, opened 2023-01-17
feat: Move settings menu into expandable components

tinaszheng:
This PR moves the inner content of the settings menu into `Expando` components

https://user-images.githubusercontent.com/59578595/213023578-d96515e8-4be6-45d3-b3e6-224a59794413.mov

https://user-images.githubusercontent.com/59578595/213023731-b2ea3384-faed-4af1-9368-b2b2c0596bef.mov

Followups:
- add the "routing api" settings option
- looks like the "max slippage" option has different sub designs too, will change in a followup PR

### DIFF
--- a/src/components/BottomSheetModal.tsx
+++ b/src/components/BottomSheetModal.tsx
@@ -62,6 +62,10 @@ const Wrapper = styled.div<{ open: boolean }>`
     height: unset;
     position: fixed;
     top: unset;
+
+    * {
+      box-sizing: border-box;
+    }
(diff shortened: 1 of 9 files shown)


just-toby (approved):
lgtm ... just noticed that the icon animation here is different than the 180deg rotation i'm using in the toolbar.

tinaszheng:
@just-toby good catch, updated this animation + converted all animations to be consistent with the rotation that currently exists on interface

PR #409, opened 2023-01-18
feat: use bottom sheet for routing diagram

just-toby:
based on the [design spec](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=1708%3A26043&t=5QXnxfgYI7whcjzT-0)

the routing diagram tooltip should also become a bottom sheet on smaller screens. this PR updates the two callsites for of RoutingDiagram to optionally wrap it in the BottomSheetModal.

also, fixes a width issue with the bottom sheet dialog

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -116,7 +116,7 @@ export const Modal = styled.div<{ color: Color }>`
   padding: 0.5em;
   position: absolute;
   top: 0;
-  width: 100%;
+  width: calc(100% - 1em);
   z-index: ${Layer.DIALOG};
 `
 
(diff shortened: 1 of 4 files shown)


src/components/Dialog.tsx
```diff
@@ -116,7 +116,7 @@ export const Modal = styled.div<{ color: Color }>`
   padding: 0.5em;
   position: absolute;
   top: 0;
-  width: 100%;
+  width: calc(100% - 1em);
```
tinaszheng, line 119: o nice, i have a different fix in my PR that does `right: 0` which doesnt need a custom calculation

tinaszheng: fixed now so this can be removed :3

tinaszheng (approved):

PR #411, opened 2023-01-18
Web 2711 price impact speedbumps

cartcrom:
Replaces "acknowledge price impact" section of the confirm swap button with a price impact speedbump, which will appear if the user clicks "Review Swap" on a high impact trade, or if they are on a Review Swap page whose trade's price impact is updated to a warning level.

Once a price impact is acknowledged, further impact update speedbumps will not appear.

Also updates LargeIcon component to accept strokeWidth prop

https://user-images.githubusercontent.com/39385577/213253793-81ef9025-2c90-4493-818d-295e7edeb606.mov

### DIFF
--- a/src/components/Swap/Speedbump/index.tsx
+++ b/src/components/Swap/Speedbump/index.tsx
@@ -1,0 +1,68 @@
+import { t, Trans } from '@lingui/macro'
+import Column from 'components/Column'
+import { StatusHeader } from 'components/Error/ErrorDialog'
+import { PriceImpact } from 'hooks/usePriceImpact'
+import { AlertTriangle, LargeIcon, X } from 'icons'
+import styled from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+const Body = styled(Column)`
+  height: calc(100% - 2.5em);
+`
+const SpeedbumpWrapper = styled.div`
+  padding: 1.75em 1.25em;
+`
+
+interface SpeedbumpDialogProps {
+  impact: PriceImpact
+  onContinue: () => void
+  onClose: () => void
+}
+
+export function SpeedbumpDialog({ impact, onContinue, onClose }: SpeedbumpDialogProps) {
+  return (
+    <SpeedbumpWrapper>
+      <div onClick={onClose}>
+        <LargeIcon icon={X} color="primary" />
+      </div>
+
+      <Body flex align="stretch" padded gap={0.75}>
+        <StatusHeader icon={AlertTriangle} iconColor="critical" iconSize={4}>
+          <Column>
+            <ThemedText.H3>
+              <Trans>Warning</Trans>
+            </ThemedText.H3>
+            <ThemedText.Body1>
+              {t`This transaction will result in a ${impact?.toString()} price impact on the market price of this pool. Do you wish to continue? `}
+            </ThemedText.Body1>
+          </Column>
+        </StatusHeader>
+        {/* <Heading gap={0.75} flex justify="center">
+          <Summary
+            input={inputAmount}
+            output={outputAmount}
+            inputUSDC={inputUSDC}
+            outputUSDC={outputUSDC}
+            impact={impact}
+            open={open}
+          />
+          <Price trade={trade} />
+        </Heading>
+        <Expando
+          title={<Subhead impact={impact} slippage={slippage} />}
+          open={open}
+          onExpand={onExpand}
+          height={6}
+          gap={open ? 0 : 0.75}
+        >
+          <Column gap={0.5}>
+            <Details trade={trade} slippage={slippage} gasUseEstimateUSD={gasUseEstimateUSD} impact={impact} />
+            <Estimate trade={trade} slippage={slippage} />
+          </Column>
+        </Expando> */}
+
+        {/* <ConfirmButton trade={trade} highPriceImpact={impact?.warning === 'error'} onConfirm={onConfirm} /> */}
+      </Body>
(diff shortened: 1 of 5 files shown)


src/components/Swap/Speedbump/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+import Column from 'components/Column'
+import { OnCloseDialogContext } from 'components/Dialog'
+import Row from 'components/Row'
+import { PriceImpact } from 'hooks/usePriceImpact'
+import { LargeAlert, LargeIcon, X } from 'icons'
+import { PropsWithChildren, useContext } from 'react'
+import styled, { css } from 'styled-components/macro'
+import { ThemedText } from 'theme'
+
+const SpeedbumpWrapper = styled.div`
```
just-toby, line 12: should `Speedbump`  be two words ? like `SpeedBump` ?

just-toby: also, should/can this be a `styled(Column)` ? i've found `Column`  to have pretty sensible defaults

src/components/Swap/Speedbump/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  padding: 1em;
+  text-align: center;
+`
+
+const Body = styled.div`
```
just-toby, line 20: same here for `styled(Column)`

src/components/Swap/Speedbump/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+`
+const ContinueButton = styled(Button)`
+  ${SpeedbumpButtonStyle}
+  background-color: ${({ theme }) => theme.criticalSoft};
+  color: ${({ theme }) => theme.critical};
+`
+const CancelButton = styled(TextButton)`
+  ${SpeedbumpButtonStyle}
+  color: ${({ theme }) => theme.secondary};
+`
```
just-toby, line 56: do you think we should be using `ActionButton` for these? (and adding to its logic for bg/text colors) ?

src/components/Swap/Speedbump/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+interface SpeedbumpDialogProps {
+  impact: PriceImpact
+  onAcknowledge: () => void
+}
+
+export function SpeedbumpDialog({ impact, onAcknowledge, children }: PropsWithChildren<SpeedbumpDialogProps>) {
+  const onClose = useContext(OnCloseDialogContext)
```
just-toby, line 64: we tend to have a pattern of keeping `Context` themselves private, and exposing a hook for the specific thing you need, to prevent improper use of the Context. it would look like exporting this in the Dialog file:

```
export const useOnDialogClose = () =>  {
    return useContext(OnCloseDialogContext)
}
```

i don't think there's much difference in this case, and i only mention it because of the pattern i've seen elsewhere.

src/components/Swap/Summary/index.tsx
```diff
@@ -91,13 +89,44 @@ interface SummaryDialogProps {
...
+        <>
+          <Header title={<Trans>Review swap</Trans>} />
+          <Body flex align="stretch">
+            <Details {...props} />
+          </Body>
+          <ConfirmButton {...props} onAcknowledgeNewTrade={onAcknowledgeNewTrade} />
+        </>
+      ) : (
+        props.impact && (
+          <SpeedbumpDialog onAcknowledge={onAcknowledgePriceImpact} impact={props.impact}>
```
just-toby, line 124: this boolean logic seems like it has the possibility to render `null` - can we combine the two checks to we always fallback to the non-warning view?

like: `props.impact && showSpeedbump ? {speedbump} : {normal view}`

src/components/Swap/Summary/index.tsx
```diff
@@ -91,13 +89,44 @@ interface SummaryDialogProps {
...
+    if (!showSpeedbump && !ackPriceImpact && props.impact?.warning === 'error') {
+      setShowSpeedbump(true)
+    }
+  }, [ackPriceImpact, props.impact?.warning, showSpeedbump])
+
+  useEffect(() => {
+    if (showSpeedbump && props.impact?.warning !== 'error') {
+      setShowSpeedbump(false)
+    }
+  }, [ackPriceImpact, props.impact, showSpeedbump])
```
just-toby, line 110: this `useEffect` is intended to catch price changes, right? i.e. catching the edge case of the `impact.warning` changing while the speedbump is visible?

cartcrom: Yeah, if the impact price goes under the speedbump threshold, we hide it. One the otherhand, if the impact price goes from under the threshold to above the threshold, we wait to show the speedbump after the user has accepted the new price (hence why setShowSpeedbump(true) is in a useCallback)

src/components/Swap/Speedbump/index.tsx
cartcrom: good point, will update to Speedbump.

As for column, I was having some trouble w/ some of the defaults there but I'll revisit.

src/components/Swap/Speedbump/index.tsx
cartcrom: I'd rather not in this PR, that would involve changing both ActionButton and the BaseButton in Button.tsx to allow overriding colors. Action button also has lots of complicated features that this button doesn't need

just-toby (approved):
nice

PR #413, opened 2023-01-18
fix: remove toolbar when on mismatched chain

just-toby:
Screenshot 2023-01-18 at 9 52 26 AM

fixes a bug where we show an empty toolbar in the `ChainError.MISMATCHED_CHAINS` case

we can just default to showing the ActionButton here, since it already has the error message and a button to switch.

(draft until design confirms this behavior)


tinaszheng (approved):
nice!

PR #415, opened 2023-01-18
chore: Use themed text instead of css for Expando headers

tinaszheng:
no UI changes!


src/components/Dialog.tsx
```diff
@@ -115,8 +115,8 @@ export const Modal = styled.div<{ color: Color }>`
   overflow: hidden;
   padding: 0.5em;
   position: absolute;
+  right: 0;
```
tinaszheng, line 118: did not make it into the last PR where i allegedly fixed this 🤦

just-toby (approved):

PR #416, opened 2023-01-18
chore: Use themed timing enums for css transition durations

tinaszheng:


just-toby (approved):
lgtm. side note, do you think it would be cool to have a custom lint rule which forbids these values from being introduced in new places? seems like it would be some lift to create an eslint plugin... but once we have that, we could introduce a bunch of custom rules.

tinaszheng:
@just-toby that could be a good idea for the backlog although it seems hard to enforce since some of these vars can still be hard coded through interpolation like the example of
```transition: color 0.125s linear, opacity ${({ isLoading }) => (isLoading ? 0 : 0.25)}s ease-in-out;```

PR #417, opened 2023-01-18
fix: design polish on inputs

just-toby:
a handful of small changes based on the [design feedback doc](https://www.notion.so/uniswaplabs/Widget-design-feedback-835bfad85aac4ad3a00550fcea18bf99)

- change balance color to always be "secondary"
- reduce the stroke width of the gas icon even more :D
- default to using the output currency as the base, in the exchange rate
- change the font size of the exchange rate
- slightly adjust the sizing/spacing of the TokenButtons (now we've diverged from the design spec, but it's a request from the recent round of design feedback. so i figured i'd make the change and discuss in review tomorrow)


src/components/TokenSelect/TokenButton.tsx
```diff
@@ -9,10 +9,10 @@ import Row from '../Row'
 import TokenImg from '../TokenImg'
 
 const StyledTokenButton = styled(Button)<{ approved?: boolean }>`
-  border-radius: ${({ theme }) => theme.borderRadius}em;
-  min-height: 2.25em;
+  border-radius: 1em;
```
tinaszheng, line 12: you can define this in terms of `theme.borderRadius.medium` once that PR goes in :P

tinaszheng (approved):

PR #421, opened 2023-01-19
fix: toolbar and expando fixes

just-toby:
- fixed the click-to-open logic on the toolbar...now the entire toolbar is clickable
- fixed the vertical padding issue in Expando by allowing the callsite to _not_ specify a fixed expanded height, and just transitioning to a large max-height value to fit the given content. I chose an arbitrary max height of 20rem which can be changed in the future if needed.


src/components/Swap/Price.tsx
```diff
@@ -54,7 +54,13 @@ export default function Price({ trade, outputUSDC }: PriceProps) {
   const [exchangeRate, usdcPrice] = useTradeExchangeRate(trade, outputUSDC, defaultBase ? 'input' : 'output')
 
   return (
-    <TextButton color="primary" onClick={onClick}>
+    <TextButton
+      color="primary"
+      onClick={(e) => {
+        onClick()
+        e.stopPropagation()
```
tinaszheng, line 61: why the `stopPropagation` here 🤔 this feels sus

tinaszheng (dismissed):
Nice nice, playing around with this the animation speed here feels kind of off since the timing is now interpolating between 0 and 20em. One change I'd suggest is to have the consumers themselves pass in `max-height` to a more reasonable bound and keep the animation timing as `AnimationSpeed.Medium`

i think the max-height approach is good tho 👍

src/components/Swap/Price.tsx
just-toby: this is because `Price` should handle clicks without the underlying `Toolbar` also receiving them. without this, when you click on `Price` to swap the exchange rate, the `Toolbar` will also expand.

is there a better way to achieve this?

tinaszheng: okay yeah i was having the same issue with my tooltips on mobile and was trying to think of a way around it

so `stopPropagation` would work but its considered a v risky maneuver as it prevents any other event handler on the page from being run. for example if you decide to add a click handler onto the whole page or if you want to do analytics tracking, it would prevent the event from reaching those callbacks. in the past these issues have been rly gnarly to debug so i would recommend not using `stopPropagation` if u can :3 

oneee alternative i can think of is to have all the clickable elements within the toolbar be `button` (i think it may already be like this?) and then detect whether or not the event was originated from a button that is nottt the current element if that makes sense

tinaszheng: another idea if we can't rely on the `button` type is to label other elements with a classname like `.clickable` and the handler can check to see if it originated from a component with `clickable` as a classname 🤔

src/components/Swap/Toolbar/index.tsx
```diff
@@ -186,7 +171,17 @@ export default memo(function Toolbar() {
...
+          flex
+          justify="space-between"
+          data-testid="toolbar"
+          onClick={(e: MouseEvent<HTMLDivElement>) => {
+            const el = e.target as HTMLDivElement
+            if (!el.classList.contains(TradePriceToggledClass)) {
+              setOpen((open) => !open)
+            }
+          }}
+        >
```
tinaszheng, line 184: i'd define this function up in the component for readability

tinaszheng: and then also do the guard-style `if` so like:

```
// Comment here about preventing Toolbar from opening when clicking on the Trade price component
if (el.classList.contains(TradePriceToggledClass)) return

setOpen(open => !open)
```

just-toby: i avoid if statements without braces, so my change is slightly different than the suggestion.

src/components/Expando.tsx
```diff
@@ -57,7 +63,12 @@ interface ExpandoProps extends ColumnProps {
   open: boolean
   onExpand: () => void
   // The absolute height of the expanded container, in em.
-  height: number
+  // If not provided, the container will expand to fit its contents up to {maxHeight}em.
+  height?: number
+  // The maximum height of the expanded container, in em.
```
zzmp, line 68: Document the max maxHeight of 20rem.

src/components/Swap/Price.tsx
```diff
@@ -54,7 +56,14 @@ export default function Price({ trade, outputUSDC }: PriceProps) {
   const [exchangeRate, usdcPrice] = useTradeExchangeRate(trade, outputUSDC, defaultBase ? 'input' : 'output')
 
   return (
-    <TextButton color="primary" onClick={onClick}>
+    <TextButton
+      color="primary"
+      onClick={(e: React.MouseEvent<HTMLButtonElement>) => {
+        onClick()
+        const el = e.target as HTMLDivElement
```
zzmp, line 63: You should not be passing state using CSS classes. Use controlled components instead.

src/components/Swap/Settings/MaxSlippageSelect.tsx
```diff
@@ -27,6 +27,10 @@ const Custom = styled(BaseButton)<{ selected: boolean }>`
   ${inputCss}
 `
 
+const ExpandoContent = styled(Row)`
```
zzmp, line 30: Is this generic enough that it should be another export of Expando.tsx?
I'm not sure if it would be good to share, or make derivative components more confusing. Up to you.

src/components/Swap/Settings/TransactionTtlInput.tsx
```diff
@@ -25,6 +25,7 @@ const InputContainer = styled(Row)`
   display: flex;
   gap: 0.5em;
   justify-content: flex-start;
+  margin: 1em 0 0;
```
zzmp, line 28: If `ExpandoContent` is exported, this should derive from `ExpandoContent`.

zzmp (changes requested):

src/components/Expando.tsx
```diff
@@ -57,7 +63,12 @@ interface ExpandoProps extends ColumnProps {
...
+  // If relying on auto-sizing, this should be something close to (but still larger than)
```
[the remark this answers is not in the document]
just-toby: sure, this is a good simplification.

src/components/Swap/Settings/MaxSlippageSelect.tsx
just-toby: no, i think the issue here is more that we can't easily add a bit of styling without creating a new named Styled Component. This use case isn't specific to Expando or unique enough to need to be centralized in the Expando file imo. If anything, i think it shows that we should allow adding margin to`Row`.

src/components/Swap/Price.tsx
just-toby: updated the approach here to not use CSS classes. 

i'm not sure i understand your Controlled Components suggestion. the goal here is not to pass state, but just to prevent nested onClicks from firing without completely preventing the click event from propagating to the rest of the page (i.e. we don't want to call `stopPropagation` like we do in `interface` - see Tina's comment above.

i think the new approach is better than using CSS classes! let me know if you have a specific implementation in mind here.

src/components/Expando.tsx
zzmp: "up to 20rem." Or "up to MAX_HEIGHT rem."

src/components/Expando.tsx
```diff
@@ -35,16 +35,24 @@ const bottomCss = css`
...
   position: relative;
-  transition: height ${AnimationSpeed.Medium}, padding ${AnimationSpeed.Medium};
+  transition: max-height ${AnimationSpeed.Medium}, padding ${AnimationSpeed.Medium};
   ${({ showBottomGradient }) => showBottomGradient && bottomCss}
 `
 
-const InnerColumn = styled(Column)<{ height: number }>`
-  height: ${({ height }) => height}em;
+const InnerColumn = styled(Column)<{ height?: number; maxHeight?: number }>`
+  max-height: ${({ height, maxHeight }) => Math.min(height ?? MAX_HEIGHT, maxHeight ?? MAX_HEIGHT)}em;
```
zzmp, line 55: You can move this logic to an in-file helper function.

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -171,19 +171,19 @@ export interface TradeProps {
...
+  return <ExpandIcon $expanded={expanded} onClick={toggleOpen} />
 }
 
 export function Trade({
   trade,
   outputUSDC,
   gasUseEstimateUSD,
   expanded,
-  onToggleExpand,
+  toggleOpen,
```
zzmp, line 186: As a callback, this should still be on*, eg onToggleOpen.

src/components/Swap/Toolbar/index.tsx
```diff
@@ -186,7 +175,20 @@ export default memo(function Toolbar() {
...
-        <ToolbarRow flex justify="space-between" data-testid="toolbar">
+        <ToolbarRow
+          flex
+          justify="space-between"
+          data-testid="toolbar"
+          onClick={(e) => {
+            // e.currentTarget is ToolbarRow, the element this listener is attached to.
+            // e.target is the element that was clicked, which could be ToolbarRow or a child.
+            // The user interaction should be handled by the child.
+            if (e.currentTarget !== e.target) {
```
zzmp, line 186: I'd prefer that children that alter the click behavior also call e.stopPropagation, instead of having the parent handler here check for child targeting.

I can see how either could be argued for, but this feels more confusing / less idiomatic to me. I also think that it will be harder to debug, as I tend to start looking for event handlers on the child and bubbling up (following the DOM model) - so if something goes wrong it will be harder to pinpoint it to a parents event handler.

zzmp (changes requested):

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -171,19 +171,27 @@ export interface TradeProps {
...
 }
 
-const Expander = ({ expanded, onToggleExpand }: ExpandProps) => {
-  return <ExpandIcon onClick={onToggleExpand} $expanded={expanded} />
+const Expander = ({ expanded, onToggleOpen }: ExpandProps) => {
+  return (
+    <ExpandIcon
+      $expanded={expanded}
+      onClick={(e: React.MouseEvent) => {
+        e.stopPropagation()
```
zzmp, line 182: nit: end the fn with stopProp

zzmp (approved):

PR #422, opened 2023-01-19
fix: collapse toolbar when action button clicked

just-toby:
automatically collapse the toolbar when the main action button is clicked.


src/components/Swap/Toolbar/index.tsx
```diff
@@ -37,6 +37,28 @@ const ToolbarRow = styled(Row)`
...
+}>({
+  open: false,
+  collapse: () => null,
+  toggleOpen: () => null,
+})
+
+export const Provider = ({ children }: PropsWithChildren) => {
+  const [open, setOpen] = useState(false)
+  const toggleOpen = useCallback(() => setOpen((open) => !open), [])
+  const collapse = useCallback(() => setOpen(false), [])
```
tinaszheng, line 53: dont need to wrap these in `useCallbacks` since theyre just simple function definitions :P

src/components/Swap/Toolbar/index.tsx
```diff
@@ -37,6 +37,28 @@ const ToolbarRow = styled(Row)`
   padding: 0 1em;
 `
 
+const Context = createContext<{
```
tinaszheng, line 40: i'd name this something more specific like `ToolbarContext`

tinaszheng (dismissed):
more of a question than a requested change but curious why you didn't choose to do this using controlled version of `setOpen(false) && setOpen(await onReviewSwapClick())` 🤔 the provider/context method seems like a roundabout way of doing the same thing

src/components/Swap/Toolbar/index.tsx
just-toby: i think we prefer this name elsewhere in the codebase if the Context is not exported.

just-toby:
> more of a question than a requested change but curious why you didn't choose to do this using controlled version of setOpen(false) && setOpen(await onReviewSwapClick()) 🤔 the provider/context method seems like a roundabout way of doing the same thing

@tinaszheng the `setOpen` in the `ActionButton` is not for the Expando, it's for the transaction `SummaryDialog` which opens when you click it. The ActionButton and the Toolbar are siblings, so i'm using Context to share state between them, for now.

In the near future I want to refactor these two components to be more explicitly coupled, because they share state (as of this PR) and should only appear  in certain configurations together. But i think this change is out of scope this sprint.

zzmp (approved):

PR #428, opened 2023-01-19
fix: hint color in dark theme

just-toby:
see the input placeholder color here: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10298%3A104518&t=MPVEUGMwd2mLGl3t-4

### DIFF
--- a/src/theme/index.tsx
+++ b/src/theme/index.tsx
@@ -73,7 +73,7 @@ export const darkTheme: Colors = {
   onAccent: white,
   primary: white,
   secondary: 'hsl(227, 21%, 67%)',
-  hint: 'hsl(225, 10%, 47.1%)',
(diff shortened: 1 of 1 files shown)


src/theme/index.tsx
```diff
@@ -73,7 +73,7 @@ export const darkTheme: Colors = {
   onAccent: white,
   primary: white,
   secondary: 'hsl(227, 21%, 67%)',
-  hint: 'hsl(225, 10%, 47.1%)',
+  hint: 'hsla(225, 18%, 44%)',
```
tinaszheng, line 76: this doesnt have an alpha value so u can just use `hsl`

tinaszheng (approved):

PR #430, opened 2023-01-20
fix: widget borders, radius, and button radius

just-toby:
removes an unnecessary border around the widget at the top level. somehow the widget had 2 borders without us noticing.

also fixes some border radii that are a little off


tinaszheng:
wait im not a fan of manually scaling these border radii 🤔 should these just be separate theme tokens instead?

just-toby:
@tinaszheng updated the border radii to use theme-defined values

src/components/WidgetWrapper.tsx
```diff
@@ -10,8 +10,8 @@ const HORIZONTAL_PADDING = 8
 const StyledWidgetWrapper = styled.div<{ width: number | string }>`
   -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
   background-color: ${({ theme }) => theme.container};
-  border: ${({ theme }) => `1px solid ${theme.outline}`};
-  border-radius: ${({ theme }) => theme.borderRadius}em;
+  /* border: ${({ theme }) => `1px solid ${theme.outline}`}; */
```
tinaszheng, line 13: did u mean to remove this?

tinaszheng (approved):

PR #432, opened 2023-01-20
fix: tooltips in animated dialogs

cartcrom:
Adds a trigger to update popover positioning, used by Dialog.tsx correctly place tooltips after slide-in animation


src/components/Swap/Settings/index.tsx
```diff
@@ -23,7 +23,7 @@ export function SettingsMenu() {
   // TODO (WEB-2754): add back reset settings functionality
   return (
     <Column gap={1} style={{ paddingTop: '1em', paddingBottom: '1em' }} ref={setBoundary} padded>
-      <PopoverBoundaryProvider value={boundary}>
+      <PopoverBoundaryProvider value={{ boundary }}>
```
cartcrom, line 26: Should I be memoizing these?

src/components/Swap/Summary/Details.tsx
```diff
@@ -57,6 +58,10 @@ function Detail({ label, value, color }: DetailProps) {
   )
 }
 
+const ToolTipBody = styled(ThemedText.Caption)`
+  max-width: 220px;
+`
+
```
cartcrom, line 64: I don't think this needs to be dynamic as this size looks good on both small 300px and large widget widths, but am opening to changing if wanted

src/components/Dialog.tsx
```diff
@@ -115,14 +126,82 @@ export const Modal = styled.div<{ color: Color }>`
   flex-direction: column;
   height: 100%;
   left: 0;
-  overflow: hidden;
```
just-toby, line 118: just checking - removing this doesn't break the entrance/exit animation for modals? you can test with the token selection modal (focus on the corners)

src/components/Dialog.tsx
```diff
@@ -115,14 +126,82 @@ export const Modal = styled.div<{ color: Color }>`
...
+  position: absolute;
+  top: 0;
+  width: 100%;
+
+  @supports (overflow: clip) {
+    overflow: clip;
+  }
+`
+
+export const DialogAnimationLengthMs = ms`250`
```
just-toby, line 166: we have a theme variable for this value now in `animations.ts`

src/components/Dialog.tsx
```diff
@@ -115,14 +126,82 @@ export const Modal = styled.div<{ color: Color }>`
...
+  const popoverRef = useRef<HTMLDivElement>(null)
+  const [updatePopover, setUpdatePopover] = useState(false)
+  useEffect(() => {
+    setTimeout(() => {
+      setUpdatePopover(true)
+    }, DialogAnimationLengthMs + PopoverAnimationUpdateDelay)
+  }, [])
+
+  return (
+    <PopoverBoundaryProvider value={{ boundary: popoverRef.current, updateTrigger: updatePopover }}>
```
just-toby, line 195: to clarify - with this, we are adding a 2nd PopoverBoundaryProvider to the widget, right? we might want to leave a comment here or with the main PopoverBoundaryProvider to tell future readers that Popovers in Dialogs are not subject to the same provider?

cartcrom: There isn't really a main popover provider, since the provider in Swap doesn't wrap dialogs

src/components/Dialog.tsx
cartcrom: great catch, fixed by adding border radius to hidden wrapper

src/theme/animations.ts
```diff
@@ -1,5 +1,11 @@
-export enum AnimationSpeed {
-  Fast = '0.125s',
-  Medium = '0.25s',
-  Slow = '0.5s',
+export enum TransitionDuration {
```
just-toby, line 1: `TransitionDurationMs` ?

just-toby (approved):
lgtm

src/components/Popover.tsx
```diff
@@ -98,7 +99,7 @@ export default function Popover({
   contained,
   showArrow = true,
 }: PopoverProps) {
-  const boundary = useContext(BoundaryContext)
+  const { boundary, updateTrigger } = useContext(BoundaryContext) || {}
```
tinaszheng, line 102: can you declare an `const EMPTY_OBJECT = {}` in a util file and use that instead so its not a new obj on every render 🤔 not a huge deal here since this value itself isnt used butttt more of a habit thing

cartcrom: I gave the context a default value of {} instead

tinaszheng (approved):

PR #433, opened 2023-01-20
fix: swap failed modal

zzmp:
Update swap failed modal to exclude the actual error object, as one is not provided by ethers.
If we want to capture an error object, it will need to be done separately, in SwapButton.tsx.

image

WEB-2793

### DIFF
--- a/src/components/Swap/Status/StatusDialog.tsx
+++ b/src/components/Swap/Status/StatusDialog.tsx
@@ -95,7 +95,7 @@ function TransactionStatus({ tx, onClose }: TransactionStatusProps) {
 export default function TransactionStatusDialog({ tx, onClose }: TransactionStatusProps) {
   return tx.receipt?.status === 0 ? (
     <ErrorDialog
-      header={<Trans>Your swap failed</Trans>}
+      header={<Trans>Your swap failed.</Trans>}
       message={
         <Trans>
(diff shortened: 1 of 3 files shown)


yyip-dev (commented):
How are we testing this update? Is there a before/after screenshots you can include?

yyip-dev (commented):
How are we testing this update? Is there a before/after screenshots you can include?

zzmp:
> How are we testing this update? Is there a before/after screenshots you can include?

I was unable to test this directly, but I flipped a boolean to simulate it, and produced the screenshot now included in the description.

just-toby (approved):
LGTM

src/components/Swap/Status/StatusDialog.tsx
```diff
@@ -95,14 +95,14 @@ function TransactionStatus({ tx, onClose }: TransactionStatusProps) {
 export default function TransactionStatusDialog({ tx, onClose }: TransactionStatusProps) {
   return tx.receipt?.status === 0 ? (
     <ErrorDialog
-      header={
+      header={<Trans>Your swap failed</Trans>}
```
tinaszheng, line 98:
```suggestion
      header={<Trans>Your swap failed.</Trans>}
```

yyip-dev (approved):

PR #434, opened 2023-01-20
fix: use conedison signTypedData with fallbacks

zzmp:
Use conedison's signTypedData utility for better x-wallet compat.

See https://github.com/Uniswap/interface/pull/5875


just-toby (approved):
ez - assuming the conedison PR was also reviewed and is good, this change nothing

PR #437, opened 2023-01-24
fix: integration fixes

just-toby:


src/components/Swap/Toolbar/index.tsx
```diff
@@ -79,6 +79,18 @@ export default memo(function Toolbar() {
...
+      !isWrap &&
+      !(state === TradeState.NO_ROUTE_FOUND || (trade && !trade.swaps)) &&
+      trade?.inputAmount &&
+      trade?.outputAmount
+    ) {
+      onToggleOpen()
+    }
+  }, [insufficientBalance, isWrap, onToggleOpen, state, trade])
+
   const caption = useMemo(() => {
```
zzmp, line 94: Instead of duplicating the logic, just return a pair, `[caption, isTogglable]`

src/components/Swap/Toolbar/index.tsx
```diff
@@ -194,14 +206,14 @@ export default memo(function Toolbar() {
...
-        <ToolbarRow flex justify="space-between" data-testid="toolbar" onClick={onToggleOpen}>
+        <ToolbarRow flex justify="space-between" data-testid="toolbar" onClick={maybeToggleOpen}>
           {caption}
         </ToolbarRow>
       }
       styledTitleWrapper={false}
       showBottomGradient={false}
       open={open}
-      onExpand={onToggleOpen}
+      onExpand={maybeToggleOpen}
```
zzmp, line 216: If this is unnecessary just rm it from the codebase.

src/components/Swap/Toolbar/index.tsx
```diff
@@ -194,14 +206,14 @@ export default memo(function Toolbar() {
   return (
     <StyledExpando
       title={
-        <ToolbarRow flex justify="space-between" data-testid="toolbar" onClick={onToggleOpen}>
+        <ToolbarRow flex justify="space-between" data-testid="toolbar" onClick={maybeToggleOpen}>
```
zzmp, line 209: onClick is handled in the parent (for the whole row), so it is unnecessary (and subverts the DOM model) on Trade and PriceImpact. Remove it there.

src/components/Swap/Toolbar/index.tsx
```diff
@@ -79,6 +79,18 @@ export default memo(function Toolbar() {
     return inputBalance && inputAmount && inputBalance.lessThan(inputAmount)
   }, [inputAmount, inputBalance])
 
+  const maybeToggleOpen = useCallback(() => {
```
zzmp, line 82: Define this after the caption, taking advantage of the (new; see below) `isTogglable` boolean within the callback.

zzmp (changes requested):

src/components/Swap/Toolbar/index.tsx
zzmp: I returned an object instead of a pair because arrays of ReactNodes require keys.

src/components/Swap/Toolbar/index.tsx
zzmp: I think this is necessary.

zzmp (approved):

PR #438, opened 2023-01-31
fix: show a better error message when there is a swap error

just-toby:
copies the Swap Error message parsing logic from `interface` and uses it to show a better error message in the widget swap flow.


src/utils/swapErrorToUserReadableMessage.ts
```diff
@@ -0,0 +1,52 @@
...
+  let reason: string | undefined
+
+  if (error.code) {
+    switch (error.code) {
+      case 4001:
+        return t`Transaction rejected`
+    }
+  }
+
+  while (error) {
```
tinaszheng, line 18: the while (error) + reassigning error loop seems kind of strange to me. can this be done by just assigning new variables?

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -90,9 +94,14 @@ export default function SwapButton({ disabled }: { disabled: boolean }) {
       // Only close the review modal if the swap submitted (ie no-throw).
       setOpen(false)
     } catch (e) {
-      console.error(e) // ignore error
+      throwAsync(
+        new SwapError({
+          header: 'Swap Failed',
```
zzmp, line 99: I think this should read "Swap Error", because the swap did not necessarily fail. Specifically, see https://github.com/Uniswap/widgets/blob/main/src/hooks/useUniversalRouter.ts#L62-L66.

src/utils/swapErrorToUserReadableMessage.test.ts
```diff
@@ -0,0 +1,113 @@
+import { swapErrorToUserReadableMessage } from './swapErrorToUserReadableMessage'
+
+describe('swapErrorToUserReadableMessage', () => {
+  test('should return "Transaction rejected" when error code is 4001', () => {
```
zzmp, line 4: nit: AFAIK the rest of the codebase uses `it`, not `test`. I think it reads better ("it should...").

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -90,9 +94,14 @@ export default function SwapButton({ disabled }: { disabled: boolean }) {
       // Only close the review modal if the swap submitted (ie no-throw).
       setOpen(false)
     } catch (e) {
-      console.error(e) // ignore error
+      throwAsync(
+        new SwapError({
```
zzmp, line 98: These errors are not fatal - they should be dismissable. Right now an error through `throwAsync` can only be dismissed through a page reload.

Can you special-case `SwapError` in `ErrorBoundary`?

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -90,9 +94,14 @@ export default function SwapButton({ disabled }: { disabled: boolean }) {
...
+          header: 'Swap Failed',
+          message: swapErrorToUserReadableMessage(e),
```
zzmp, line 100: The thrown error has already been converted to a user readable message. You should update those callsites, so that the error here can just be propagated up.

See https://github.com/Uniswap/widgets/blob/main/src/hooks/useUniversalRouter.ts#L58-L66
See https://github.com/Uniswap/widgets/blob/main/src/hooks/swap/useSendSwapTransaction.tsx#L125

zzmp (changes requested):

src/utils/swapErrorToUserReadableMessage.ts
zzmp: I'm not sure if it can, because this is done because there is an unknown number of nested errors here.
fwiw this is legacy code, so it may be better to not touch it unless we have documentation or can repro the errors that it was created to handle.

src/components/Error/ErrorBoundary.tsx
```diff
@@ -54,11 +60,24 @@ export default class ErrorBoundary extends Component<PropsWithChildren<ErrorBoun
   }
 
   renderErrorDialog(error: Error) {
-    const header = error instanceof WidgetError ? error.header : DEFAULT_ERROR_HEADER
-    const action = error instanceof WidgetError ? error.action : DEFAULT_ERROR_ACTION
+    const header =
+      error instanceof WidgetError ? (error.dismissable ? DISMISS_ERROR_HEADER : error.header) : DEFAULT_ERROR_HEADER
```
zzmp, line 64: I don't think you should be overwriting the header here:

    const isDismissable = error instanceof WidgetError && error.dismissable
    const header = error instanceof WidgetError ? (error.header) : (isDismissable ? DISMISS_ERROR_HEADER : DEFAULT_ERROR_HEADER)

src/components/Error/ErrorBoundary.tsx
```diff
@@ -1,4 +1,10 @@
-import { DEFAULT_ERROR_ACTION, DEFAULT_ERROR_HEADER, WidgetError } from 'errors'
+import {
+  DEFAULT_ERROR_ACTION,
+  DEFAULT_ERROR_HEADER,
+  DISMISS_ERROR_ACTION,
```
zzmp, line 4: I think this should be named DEFAULT_DISMISSABLE_ERROR_HEADER, etc. It is still used as a default (assuming it is only used when no other header is passed.).

src/components/Error/ErrorBoundary.tsx
```diff
@@ -1,4 +1,10 @@
-import { DEFAULT_ERROR_ACTION, DEFAULT_ERROR_HEADER, WidgetError } from 'errors'
+import {
+  DEFAULT_ERROR_ACTION,
```
zzmp, line 2: These defaults are only used in this file. Could you just define them here, to avoid importing them?

zzmp (changes requested):

src/components/Error/ErrorBoundary.tsx
just-toby: actually i think this version is has a problem too: we only set the `dismissable` attribute on `WidgetErrors` , so in this case we'd never actually show `DISMISS_ERROR_HEADER`

i made an update that should address this in a different way (setting the attributes as the defaults for `SwapError`s)

src/errors.ts
```diff
@@ -3,6 +3,9 @@ import { t } from '@lingui/macro'
 export const DEFAULT_ERROR_HEADER = t`Please refresh the page and try again.`
 export const DEFAULT_ERROR_ACTION = t`Reload the page`
 
+const DEFAULT_DISMISSABLE_ERROR_HEADER = t`Please try again.`
+const DEFAULT_DISMISSABLE_ERROR_ACTION = t`Dismiss`
+
 interface WidgetErrorConfig {
```
zzmp, line 9: Should this include `dismissable?: boolean`?

src/errors.ts
```diff
@@ -34,6 +38,18 @@ class ConnectionError extends WidgetError {
   }
 }
 
+export class SwapError extends WidgetError {
+  constructor(config: WidgetErrorConfig) {
+    super({
+      ...config,
```
zzmp, line 44: nit: I prefer this syntax:

    super({
      header: DEFAULT_DISMISSABLE_ERROR_HEADER,
      action: DEFAULT_DISMISSABLE_ERROR_ACTION,
      ...config,
    })

src/errors.ts
```diff
@@ -34,6 +38,18 @@ class ConnectionError extends WidgetError {
...
+      header: config.header ?? DEFAULT_DISMISSABLE_ERROR_HEADER,
```
zzmp, line 45: This is never used - all of your `SwapError`'s constructors pass in `Swap Error`.
Could you rm `DEFAULT_DISMISSABLE_ERROR_HEADER`, and inline

    `t`Swap Error`

here instead?

zzmp: Alternatively, define a `DismissableWidgetError` that takes on these defaults, and then derive `SwapError` from that.

zzmp (approved):
Down to nits. I'm happy to take another look if you make changes, but its not strictly necessary. 😄

src/errors.ts
zzmp: If you subclass a `DismissableWidgetError`, this should not be exposed in the config. It should auto-populate to `true` if you subclass from `DismissableWidgetError`.

src/errors.ts
```diff
@@ -24,6 +24,17 @@ export abstract class WidgetError extends Error {
   }
 }
 
+abstract class DismissableWidgetError extends WidgetError {
+  constructor(config: WidgetErrorConfig) {
+    super({
+      action: config.action ?? DEFAULT_DISMISSABLE_ERROR_ACTION,
+      header: config.header ?? DEFAULT_DISMISSABLE_ERROR_HEADER,
+      ...config,
```
zzmp, line 32: This could be buggy, in the case that action or header are explicitly undefined. They will then be overwritten on L32 to be undefined. Consider

    const a = {
      a: 1,
      ...{a: undefined}
    }

src/errors.ts
```diff
@@ -38,15 +49,10 @@ class ConnectionError extends WidgetError {
...
 
-export class SwapError extends WidgetError {
+export class SwapError extends DismissableWidgetError {
   constructor(config: WidgetErrorConfig) {
-    super({
-      ...config,
-      header: config.header ?? DEFAULT_DISMISSABLE_ERROR_HEADER,
-      action: config.action ?? DEFAULT_DISMISSABLE_ERROR_ACTION,
-    })
+    super(config)
```
zzmp, line 54: If you use

    super({ header: t`Swap Error`, ...config })

you can omit the header from all other callsites.

PR #439, opened 2023-01-31
feat: add modal zIndex to theme definition

just-toby:
gives bottom sheet modals a higher zIndex by default - this is also configurable by the integration theme, so anyone can specify where exactly the modals should appear on the z Index

I tested this by installing a build in `interface` and verifying that I could make the bottom sheet appear over the Mobile Nav Bar

### DIFF
--- a/src/components/BottomSheetModal.tsx
+++ b/src/components/BottomSheetModal.tsx
@@ -35,6 +35,7 @@ const BottomSheetModalBackdrop = styled.div<{ className?: string }>`
   top: 0;
   transition: visibility 0s linear 0s, opacity ${AnimationSpeed.Medium};
   visibility: visible;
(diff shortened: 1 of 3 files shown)


src/components/BottomSheetModal.tsx
```diff
@@ -35,6 +35,7 @@ const BottomSheetModalBackdrop = styled.div<{ className?: string }>`
   top: 0;
   transition: visibility 0s linear 0s, opacity ${AnimationSpeed.Medium};
   visibility: visible;
+  z-index: ${({ theme }) => theme.zIndex.modal - 1};
```
tinaszheng, line 38: clever!

tinaszheng (approved):

PR #440, opened 2023-01-31
feat: add dev script to create a local tarball for installation into …

just-toby:
…another local project

add an script to easily generate a local-only build. this creates a file called `uniswap-widgets-v0.0.1.tgz` in the current directory, which can be installed in another local package with the command `yarn add file:<absolute path to file>`

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "deduplicate": "yarn-deduplicate --strategy=highest",
     "lint": "eslint .",
     "test": "jest src",
-    "test:e2e": "jest e2e"
+    "test:e2e": "jest e2e",
+    "pack:local": "yarn cache clean && yarn build && json -I -f package.json -e \"this.version='0.0.1'\" && yarn pack && json -I -f package.json -e \"delete this.version\""
   },
   "browserslist": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,8 @@
     "deduplicate": "yarn-deduplicate --strategy=highest",
     "lint": "eslint .",
     "test": "jest src",
-    "test:e2e": "jest e2e"
+    "test:e2e": "jest e2e",
+    "pack:local": "yarn cache clean && yarn build && json -I -f package.json -e \"this.version='0.0.1'\" && yarn pack && json -I -f package.json -e \"delete this.version\""
```
zzmp, line 47: Why do you need to clear the yarn cache?

zzmp: The output tgz should be `.gitignore`d so it is not accidentally committed.

zzmp (changes requested):
I'd prefer something automated between widgets/interface to accomplish this task, and I think that the toolchain should be updated to accomodate it.

I'm fine with something like this PR in the interim, but I don't think it should be the end goal.

PR #441, opened 2023-02-01
feat: include tx info in OnTxSuccess

just-toby:
include the Transaction Info in the `OnTxSuccess` callback - this gives the integrator access to the trade info easily without caching it themselves.

`BREAKING CHANGE`

### DIFF
--- a/src/hooks/transactions/index.tsx
+++ b/src/hooks/transactions/index.tsx
@@ -66,7 +66,8 @@ export function useIsPendingApproval(token?: Token): boolean {
 }
 
 export type OnTxSubmit = (hash: string, tx: Transaction) => void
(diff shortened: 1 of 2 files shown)


src/hooks/transactions/index.tsx
```diff
@@ -66,7 +66,8 @@ export function useIsPendingApproval(token?: Token): boolean {
 }
 
 export type OnTxSubmit = (hash: string, tx: Transaction) => void
-export type OnTxSuccess = (hash: string, receipt: TransactionReceipt) => void
+type WithRequired<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>
```
tinaszheng, line 69: this seems like it should go in a `types.d.ts` file so it can be used globally without needing to import 👀

tinaszheng (approved):

PR #443, opened 2023-02-02
fix: hide connection UI based on integration prop

just-toby:
hide the "connect wallet" action button when this prop is true


vm (approved):
hideConnectButton?

just-toby:
> hideConnectButton?

it's not just the button!

just-toby:
also prefer not to change the integration API like that

PR #444, opened 2023-02-02
fix: color of connect wallet button

just-toby:
matches the color and hover state of this button to the one in `interface` as per design request

### DIFF
--- a/src/components/ActionButton.tsx
+++ b/src/components/ActionButton.tsx
@@ -78,7 +78,7 @@ export interface Action {
 
 type ActionButtonColor = keyof Pick<
   Colors,
-  'accent' | 'actionSoft' | 'accentSoft' | 'warningSoft' | 'interactive' | 'critical'
+  'accent' | 'activeSoft' | 'accentSoft' | 'warningSoft' | 'interactive' | 'critical'
 >
 
 interface BaseProps {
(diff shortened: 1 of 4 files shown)


src/theme/index.tsx
```diff
@@ -34,6 +34,8 @@ export const lightTheme: Colors = {
   // surface
   accent: brandLight,
   accentSoft: rgba(brandLight, 0.12),
+  action: brandLight,
+  actionSoft: rgba(brandLight, 0.24),
```
cartcrom, line 38: I don't think we should be introducing a new variable here. In interface, we have accentAction and accentActive, which are the same in darkmode (blue) and different in darkmode (pink vs blue). Since widget only uses one accent color in both themes, we don't need a sense of action vs active.

accentSoft should just be updated to 0.24 opacity instead of introducing action

cartcrom (changes requested):

src/theme/index.tsx
just-toby: makes sense, updated

src/components/ActionButton.tsx
```diff
@@ -107,6 +110,8 @@ export default function ActionButton({
         return 'onAccent'
       case 'accentSoft':
         return 'accent'
+      case 'activeSoft':
+        return 'active'
```
cartcrom, line 114: i think we don't need this anymore right

cartcrom (changes requested):

src/components/ActionButton.tsx
just-toby: ah yeah i think it's unused. thanks

cartcrom (approved):

PR #446, opened 2023-02-02
fix: estimated output tooltip in summary

just-toby:
adds the missing tooltip to the "price updated" action button
Screenshot 2023-02-06 at 10 55 27 AM
Screenshot 2023-02-06 at 10 55 17 AM


src/components/Swap/Summary/index.tsx
```diff
@@ -64,7 +68,12 @@ function ConfirmButton({
       return {
         color: 'accent',
         message: <Trans>Price updated</Trans>,
-        icon: AlertTriangle,
+        icon: () => <AlertTriangle color="accent" />,
```
cartcrom, line 71: Should we fix the issue where having tooltipContent causes the icon to be the wrong color, rather than having to manually recolor the icon here? It seems like a bug in ActionButton

cartcrom (changes requested):
@just-toby nit: pasting quick screen captures into PR description (I do cmd+ctrl+shift+4 to capture directly to clipboard) showing changes / new UI is always appreciated

src/components/Swap/Summary/index.tsx
just-toby: good call, fixed

just-toby:
@cartcrom added screenshots 🤝

cartcrom (approved):
Looks great

PR #449, opened 2023-02-03
chore: Transform quote results into Trades within rtk-query

tinaszheng:
Our logic previously was set up such that the transform from quote result -> Trade was happening outside of rtk-query. This resulted in a tricky edge case during refetching where input/outputs were unsynced. Steps to repro:

- Get a quote for 0.1 ETH -> DAI. rtk-query returns the quote response, and `data` and `currentData` get filled in.
- Revise quote to 0.1 USDC -> DAI. rtk-query refetches. `data` is the cached result from the previous response, with the route for `ETH -> DAI`. `currentData` is undefined during the refetch.
- We're now creating a `Trade` object using the USDC -> DAI inputs but using `data` that has an ETH -> DAI route. This would error, but we had a [line](https://github.com/Uniswap/widgets/blob/main/src/state/routing/utils.ts#L24) that caught this case:

```
if (parsedTokenIn.address !== currencyIn.wrapped.address) return undefined
if (parsedTokenOut.address !== currencyOut.wrapped.address) return undefined
```
where parsedTokenIn = token from response, and currencyIn = token from input 

This is fine, except for future trade types, this logic wouldn't apply similarly.

The fix is to simply apply the transform -> `Trade` from within rtk-query so that the data that is being transformed is always coupled with the correct corresponding inputs.

### DIFF
--- a/src/hooks/routing/types.ts
+++ b/src/hooks/routing/types.ts
@@ -9,3 +9,10 @@ export enum PoolType {
   V2Pool = 'v2-pool',
   V3Pool = 'v3-pool',
 }
+
+// swap router API special cases these strings to represent native currencies
+// all chains have "ETH" as native currency symbol except for polygon
+export enum SwapRouterNativeAssets {
+  MATIC = 'MATIC',
+  ETH = 'ETH',
+}
--- a/src/hooks/routing/types.ts
+++ b/src/hooks/routing/types.ts
@@ -4,3 +4,8 @@ export enum RouterPreference {
   API,
   CLIENT,
 }
+
+export enum PoolType {
+  V2Pool = 'v2-pool',
+  V3Pool = 'v3-pool',
+}
--- a/src/hooks/routing/useRouterTrade.ts
+++ b/src/hooks/routing/useRouterTrade.ts
@@ -9,7 +9,6 @@ import { useCallback, useMemo } from 'react'
 import { useGetQuoteArgs } from 'state/routing/args'
 import { useGetQuoteQueryState, useLazyGetQuoteQuery } from 'state/routing/slice'
 import { InterfaceTrade, NO_ROUTE, TradeState } from 'state/routing/types'
-import { computeRoutes, transformRoutesToTrade } from 'state/routing/utils'
 
 import { RouterPreference } from './types'
 
@@ -72,31 +71,21 @@ export function useRouterTrade(
   }, [fulfilledTimeStamp, pollingInterval, queryArgs, trigger])
   useTimeout(request, 200)
 
-  const quote = typeof data === 'object' ? data : undefined
-  const trade = useMemo(() => {
-    const routes = computeRoutes(currencyIn, currencyOut, tradeType, quote)
-    if (!routes || routes.length === 0) return
-    try {
-      return transformRoutesToTrade(routes, tradeType)
-    } catch (e: unknown) {
-      console.debug('transformRoutesToTrade failed: ', e)
-      return
-    }
-  }, [currencyIn, currencyOut, quote, tradeType])
-  const isValidBlock = useIsValidBlock(Number(quote?.blockNumber))
+  const quoteResult = typeof data === 'object' ? data : undefined
+  const isValidBlock = useIsValidBlock(Number(quoteResult?.blockNumber))
   const isValid = currentData === data && isValidBlock
-  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(quote?.gasUseEstimateUSD)
+  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(quoteResult?.gasUseEstimateUSD)
 
   return useMemo(() => {
     if (!amountSpecified || isError || queryArgs === skipToken) {
       return TRADE_INVALID
     } else if (data === NO_ROUTE) {
       return TRADE_NOT_FOUND
-    } else if (!trade) {
+    } else if (!quoteResult?.trade) {
       return TRADE_LOADING
     } else {
       const state = isValid ? TradeState.VALID : TradeState.LOADING
-      return { state, trade, gasUseEstimateUSD }
+      return { state, trade: quoteResult?.trade, gasUseEstimateUSD }
     }
-  }, [isError, amountSpecified, queryArgs, data, trade, isValid, gasUseEstimateUSD])
+  }, [amountSpecified, isError, queryArgs, data, quoteResult?.trade, isValid, gasUseEstimateUSD])
 }
--- a/src/hooks/routing/useRouterTrade.ts
+++ b/src/hooks/routing/useRouterTrade.ts
@@ -7,7 +7,7 @@ import useTimeout from 'hooks/useTimeout'
 import ms from 'ms.macro'
 import { useCallback, useMemo } from 'react'
 import { useGetQuoteArgs } from 'state/routing/args'
-import { useGetQuoteQueryState, useLazyGetQuoteQuery } from 'state/routing/slice'
+import { TradeResult, useGetQuoteQueryState, useLazyGetQuoteQuery } from 'state/routing/slice'
 import { InterfaceTrade, NO_ROUTE, TradeState } from 'state/routing/types'
 
 import { RouterPreference } from './types'
@@ -71,21 +71,21 @@ export function useRouterTrade(
   }, [fulfilledTimeStamp, pollingInterval, queryArgs, trigger])
   useTimeout(request, 200)
 
-  const quoteResult = typeof data === 'object' ? data : undefined
-  const isValidBlock = useIsValidBlock(Number(quoteResult?.blockNumber))
+  const tradeResult: TradeResult | undefined = typeof data === 'object' ? data : undefined
+  const isValidBlock = useIsValidBlock(Number(tradeResult?.blockNumber))
   const isValid = currentData === data && isValidBlock
-  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(quoteResult?.gasUseEstimateUSD)
+  const gasUseEstimateUSD = useStablecoinAmountFromFiatValue(tradeResult?.gasUseEstimateUSD)
 
   return useMemo(() => {
     if (!amountSpecified || isError || queryArgs === skipToken) {
       return TRADE_INVALID
     } else if (data === NO_ROUTE) {
       return TRADE_NOT_FOUND
-    } else if (!quoteResult?.trade) {
+    } else if (!tradeResult?.trade) {
       return TRADE_LOADING
     } else {
       const state = isValid ? TradeState.VALID : TradeState.LOADING
-      return { state, trade: quoteResult?.trade, gasUseEstimateUSD }
+      return { state, trade: tradeResult?.trade, gasUseEstimateUSD }
     }
-  }, [amountSpecified, isError, queryArgs, data, quoteResult?.trade, isValid, gasUseEstimateUSD])
+  }, [amountSpecified, isError, queryArgs, data, tradeResult?.trade, isValid, gasUseEstimateUSD])
 }
--- a/src/hooks/routing/useRouterTrade.ts
+++ b/src/hooks/routing/useRouterTrade.ts
@@ -7,8 +7,8 @@ import useTimeout from 'hooks/useTimeout'
 import ms from 'ms.macro'
 import { useCallback, useMemo } from 'react'
 import { useGetQuoteArgs } from 'state/routing/args'
-import { TradeResult, useGetQuoteQueryState, useLazyGetQuoteQuery } from 'state/routing/slice'
-import { InterfaceTrade, NO_ROUTE, TradeState } from 'state/routing/types'
+import { useGetQuoteQueryState, useLazyGetQuoteQuery } from 'state/routing/slice'
+import { InterfaceTrade, NO_ROUTE, TradeResult, TradeState } from 'state/routing/types'
 
 import { RouterPreference } from './types'
 
--- a/src/hooks/routing/useRouterTrade.ts
+++ b/src/hooks/routing/useRouterTrade.ts
@@ -7,7 +7,7 @@ import useTimeout from 'hooks/useTimeout'
 import ms from 'ms.macro'
 import { useCallback, useMemo } from 'react'
 import { useGetQuoteArgs } from 'state/routing/args'
-import { useGetQuoteQueryState, useLazyGetQuoteQuery } from 'state/routing/slice'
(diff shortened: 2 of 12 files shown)


src/utils/currencyId.ts
```diff
@@ -1,5 +1,12 @@
 import { Currency } from '@uniswap/sdk-core'
 
+// swap router API special cases these strings to represent native currencies
+// all chains have "ETH" as native currency symbol except for polygon
+export enum SwapRouterNativeAssets {
```
just-toby, line 5: this feels like it will break something in the future.

i think a similar thing caused the price charts to break recently. hopefully we are standardizing the DB keys for native assets

src/state/routing/utils.ts
```diff
@@ -1,70 +1,95 @@
...
   if (quoteResult.route.length === 0) return []
 
-  const parsedTokenIn = parseToken(quoteResult.route[0][0].tokenIn)
-  const parsedTokenOut = parseToken(quoteResult.route[0][quoteResult.route[0].length - 1].tokenOut)
+  const tokenIn = quoteResult.route[0]?.[0]?.tokenIn
+  const tokenOut = quoteResult.route[0]?.[quoteResult.route[0]?.length - 1]?.tokenOut
 
-  if (parsedTokenIn.address !== currencyIn.wrapped.address) return undefined
-  if (parsedTokenOut.address !== currencyOut.wrapped.address) return undefined
+  if (!tokenIn || !tokenOut) throw new Error('Expected both tokenIn and tokenOut to be present')
```
just-toby, line 33: do you want this to get caught by the `ErrorBoundary` and display the error message? if so, you should test this case. it might require you to use `useAsyncError` depending on how this is called

src/utils/currencyId.ts
tinaszheng: so these are constants supported by the routing-api, not token keys stored in any db. we'll have to pass in either a string `ETH` or just make an enum

src/state/routing/utils.ts
tinaszheng: depends, we throw errors in other parts of the quote flow too and those are suppressed by just returning `undefined` for a trade and console logging the error, so this is consistent with the existing flow

tinaszheng: i plan to address in a separate PR, but it would be nice to get to a point where we return all errors as part of rtk-query's default error handling mechanism, and we can access all errors through `const { data, error } = useGetQuoteQueryState` and then do something like this: https://github.com/Uniswap/mobile/blob/main/src/features/transactions/swap/useSwapWarnings.tsx#L58

src/state/routing/utils.ts
```diff
@@ -1,70 +1,95 @@
...
 
+      const isOnlyV2 = isV2OnlyRoute(route)
+      const isOnlyV3 = isV3OnlyRoute(route)
+
       return {
-        routev3: isV3Route(route) ? new V3Route(route.map(parsePool), currencyIn, currencyOut) : null,
-        routev2: !isV3Route(route) ? new V2Route(route.map(parsePair), currencyIn, currencyOut) : null,
-        inputAmount: CurrencyAmount.fromRawAmount(currencyIn, rawAmountIn),
-        outputAmount: CurrencyAmount.fromRawAmount(currencyOut, rawAmountOut),
+        routev3: isOnlyV3 ? new V3Route(route.map(parsePool), parsedCurrencyIn, parsedCurrencyOut) : null,
```
zzmp, line 54: nit: I prefer these booleans inlined, because they're only used once.

src/state/routing/utils.ts
```diff
@@ -89,6 +114,10 @@ const parsePair = ({ reserve0, reserve1 }: V2PoolInRoute): Pair =>
     CurrencyAmount.fromRawAmount(parseToken(reserve1.token), reserve1.quotient)
   )
 
-function isV3Route(route: V3PoolInRoute[] | V2PoolInRoute[]): route is V3PoolInRoute[] {
-  return route[0].type === 'v3-pool'
+function isV2OnlyRoute(route: (V3PoolInRoute | V2PoolInRoute)[]): route is V2PoolInRoute[] {
```
zzmp, line 117: nit: Can you replace both of these with a generic fn, because they essentially disambiguate a generic type?

    const isOnlyV2 = isVersionedRoute<V2PoolInRoute>(PoolType.V2Pool, route)
    const isOnlyV3 = isVersionedRoute<V3PoolInRoute>(PoolType.V3Pool, route)

    function isVersionedRoute<T extends V2PoolInRoute | V3PoolInRoute>(
      type: T['type'],
      route: (V3PoolInRoute | V2PoolInRoute)[]
    ): route is T[] {
      return route.every((pool) => pool.type === type)
    }

zzmp (changes requested):

src/state/index.tsx
```diff
@@ -17,6 +17,9 @@ export const store = configureStore({
     getDefaultMiddleware({
       thunk: true,
       serializableCheck: {
+        // meta.arg and meta.baseQueryMeta are defaults. trade is a nonserializable return value, but that's ok
+        // because we are not persisting it to a redux store
```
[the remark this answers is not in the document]
tinaszheng: i wanted to update this comment ya, i meant that its not in persisted store (aka like, local storage or asyncStorage on mobile), where things do have to be serialized

src/state/routing/utils.ts
tinaszheng: ah sorry they are copy pasted from the mobile stack where we have mixed routes support, and they're used in the mixed routes logic, which is coming up very soon so will keep these in

src/state/routing/utils.ts
tinaszheng: good idea!

src/hooks/routing/useRouterTrade.ts
```diff
@@ -72,31 +71,21 @@ export function useRouterTrade(
   }, [fulfilledTimeStamp, pollingInterval, queryArgs, trigger])
   useTimeout(request, 200)
 
-  const quote = typeof data === 'object' ? data : undefined
-  const trade = useMemo(() => {
-    const routes = computeRoutes(currencyIn, currencyOut, tradeType, quote)
-    if (!routes || routes.length === 0) return
```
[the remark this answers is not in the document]
tinaszheng: yup, moved into computeRoutes: https://github.com/Uniswap/widgets/blob/main/src/state/routing/utils.ts#L19

src/state/routing/slice.ts
```diff
@@ -56,14 +64,15 @@ export const routing = createApi({
 
               // NO_ROUTE should be treated as a valid response to prevent retries.
               if (typeof data === 'object' && data.errorCode === 'NO_ROUTE') {
-                return { data: NO_ROUTE as GetQuoteResult }
+                return { data: NO_ROUTE as TradeQuoteResult }
```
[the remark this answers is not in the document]
tinaszheng: so that function definition is a very complicated type using `QueryReturnValue` that rtk-query does not allow you to export 🙃 

but the return type is already inferred from us setting the type of `baseQuery` [here](https://github.com/Uniswap/widgets/blob/c4c49155081bbe6cbb63a92d802cc59da742267f/src/state/routing/slice.ts#L19)

i had to cast it like i had previously because the rtk-query typechecker didn't like that `TradeQuoteResult` is technically a union, and returning a subtype is seen as not valid

src/state/routing/types.ts
```diff
@@ -87,6 +87,7 @@ export interface QuoteResult {
 export const INITIALIZED = 'Initialized'
 export const NO_ROUTE = 'No Route'
 
-export type GetQuoteResult = QuoteResult | typeof INITIALIZED | typeof NO_ROUTE
+export type GetQuoteError = typeof INITIALIZED | typeof NO_ROUTE
```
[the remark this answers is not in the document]
tinaszheng: so NO_ROUTE is technically an error i think? it gives us a 500 iirc. if we use rtk-query `query` instead of `queryFn` it will give that result in the `error` field

not sure about initialized - i feel like we could do something different there instead of returning a str, but that could be a discussion later

they're not valid quote results with routes, so they feel like errors in my mind 🤔

src/state/routing/utils.ts
```diff
@@ -1,70 +1,95 @@
...
-  currencyOut: Currency | undefined,
-  tradeType: TradeType,
-  quoteResult: Pick<QuoteResult, 'route'> | undefined
-) {
-  if (!quoteResult || !quoteResult.route || !currencyIn || !currencyOut) return undefined
+  tokenInIsNative: boolean,
+  tokenOutIsNative: boolean,
+  quoteResult?: Pick<QuoteResult, 'route'>
+):
+  | {
```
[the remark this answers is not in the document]
tinaszheng: `Trade` is the class 😭

src/state/routing/utils.ts
```diff
@@ -1,70 +1,95 @@
...
+  const parsedCurrencyOut = tokenOutIsNative ? nativeOnChain(tokenOut.chainId) : parseToken(tokenOut)
 
   try {
     return quoteResult.route.map((route) => {
       if (route.length === 0) {
         throw new Error('Expected route to have at least one pair or pool')
       }
-      const rawAmountIn = route[0].amountIn
-      const rawAmountOut = route[route.length - 1].amountOut
+      const rawAmountIn = route[0]?.amountIn
```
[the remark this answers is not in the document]
tinaszheng: keeping the check to be `route.length` because it's more canonical but removing the `?` operators

zzmp (changes requested):

src/state/routing/utils.ts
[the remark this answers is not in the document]
zzmp: i agree .length is better - I thought maybe it didn't give the type assertion necessary to remove the optional operators. :)

src/state/index.tsx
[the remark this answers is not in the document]
zzmp: I'm still not sure *why* this is not being stored

src/utils/currencyId.ts
zzmp: can they be imported from the routing package?

src/state/index.tsx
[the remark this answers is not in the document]
tinaszheng: it is in redux but not in local storage. Does that answer your q? I updated the comment btw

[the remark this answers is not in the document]
tinaszheng: https://redux-toolkit.js.org/rtk-query/usage/persistence-and-rehydration - we don't have this set up but if we wanted state to persist through browser sessions, we could use this framework and that's where we actually have to really pay attention serialization

src/utils/currencyId.ts
tinaszheng: i think longer-ish term i'd prefer to have routing-api support `NATIVE` as a string and then have it be exported from sdk-core. That way it would match up with our string addresses for our graphql and amplitude endpoints.

src/state/routing/types.ts
```diff
@@ -87,6 +87,13 @@ export interface QuoteResult {
...
+export type GetQuoteResult = QuoteResult | GetQuoteError
```
zzmp, line 91: Shouldn't `GetQuoteResult` now be `TradeResult | GetQuoteError`? `QuoteResult` is no longer directly returned, and this affects the typing of the rtk-query slice. It is only named GetQuoteResult because it is returned by GetQuoteQuery, so if the query returns something else (TradeResultQuote), then things should be renamed.

zzmp (changes requested):
logic lgtm, but the type naming should be corrected to minimize aliasing (see comments) and the native token names should be imported or annotated as to why not.

zzmp (approved):
only nits left so switching to approve

src/state/routing/utils.ts
```diff
@@ -89,6 +116,17 @@ const parsePair = ({ reserve0, reserve1 }: V2PoolInRoute): Pair =>
...
-function isV3Route(route: V3PoolInRoute[] | V2PoolInRoute[]): route is V3PoolInRoute[] {
-  return route[0].type === 'v3-pool'
+function isVersionedRoute<T extends V2PoolInRoute | V3PoolInRoute>(
+  type: T['type'],
+  route: (V3PoolInRoute | V2PoolInRoute)[]
+): route is T[] {
+  return route.every((pool) => pool.type === type)
+}
+
+export function currencyAddressForSwapQuote(currency: Currency): string {
```
JFrankfurt, line 126: naming fn w/r/t its consumer instead of its functionality feels wrong to me

src/hooks/routing/useRouterTrade.ts
```diff
@@ -72,31 +71,21 @@ export function useRouterTrade(
...
       return TRADE_INVALID
     } else if (data === NO_ROUTE) {
       return TRADE_NOT_FOUND
-    } else if (!trade) {
+    } else if (!tradeResult?.trade) {
       return TRADE_LOADING
     } else {
       const state = isValid ? TradeState.VALID : TradeState.LOADING
-      return { state, trade, gasUseEstimateUSD }
+      return { state, trade: tradeResult?.trade, gasUseEstimateUSD }
```
JFrankfurt, line 88: this reads like `TradeState.LOADING` can be the result of two logic paths, is that right? Having the last `else` case include a ternary when we're concluding an existing if/else block feels messy.

JFrankfurt (approved):

src/state/routing/utils.ts
tinaszheng: could rename to `currencyAddressForRoutingApi` but technically this should just be used for getting swap quotes - will also be not a thing once i ask the routing team to support `NATIVE` as an exported constant

src/hooks/routing/useRouterTrade.ts
tinaszheng: can take a look at refactoring the loading states in another pr!

src/state/routing/types.ts
tinaszheng: thoughts on the name `getSwapTradeResult` for the queryFn/hook name?

src/hooks/routing/types.ts
```diff
@@ -4,3 +4,15 @@ export enum RouterPreference {
...
   CLIENT,
 }
+
+export enum PoolType {
+  V2Pool = 'v2-pool',
+  V3Pool = 'v3-pool',
+}
+
+// swap router API special cases these strings to represent native currencies
+// all chains have "ETH" as native currency symbol except for polygon
```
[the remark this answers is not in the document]
tinaszheng: going to add a TODO to import `NATIVE` once it's available (i just requested the feature)

PR #450, opened 2023-02-03
fix: improve swap error

just-toby:
improve default values for dismissable and swap error

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -51,7 +51,7 @@ class ConnectionError extends WidgetError {
 
 export class SwapError extends DismissableWidgetError {
   constructor(config: WidgetErrorConfig) {
-    super({ header: t`Swap Error`, ...config })
+    super({ ...config, header: config.header ?? t`Swap Error` })
     this.name = 'SwapError'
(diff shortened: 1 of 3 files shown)


src/errors.ts
```diff
@@ -51,7 +51,7 @@ class ConnectionError extends WidgetError {
 
 export class SwapError extends DismissableWidgetError {
   constructor(config: WidgetErrorConfig) {
-    super(config)
+    super({ header: t`Swap Error`, ...config })
```
zzmp, line 54: very subtle bug here:
In L30, if an `action` or `header` is explicitly `undefined`, it will still be set to the default, because you are setting config _then_ setting them again with the nullish operator.
Here, if `config` explicitly sets the header to `undefined`, you'll end up with the default, not the `Swap Error` header.

This may never occur, but is something to be aware of. In this case, I'd mirror L30 here:

    super({ ...config, header: config.header ?? t`Swap Error` })

zzmp (changes requested):

zzmp (approved):

PR #452, opened 2023-02-03
feat: update permit2 flow

cartcrom:
Updates permit2 flow to match new design.

example: (first pass includes full flow, second pass includes existing signature)

https://user-images.githubusercontent.com/39385577/216707995-7b7b2fc5-e733-4c26-9145-b02bc45b5680.mov


src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +30,176 @@ const PriceImpactText = styled.span`
...
+    children: <Trans>Cancel</Trans>,
+  }
+}
+
+enum ReviewState {
+  REVIEWING,
+  ALLOWANCE_PENDING,
+  ALLOWANCE_FAILED,
+  TRADE_CHANGED,
+  PENDING_SWAP,
```
tinaszheng, line 90: nit: `SWAP_PENDING` seems more consistent with the other enums 😅

src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +30,176 @@ const PriceImpactText = styled.span`
   color: ${({ theme }) => theme.error};
 `
 
+const PermitTooltipBody = styled.div`
+  max-width: 220px;
+`
```
just-toby, line 35: this is now being re-used in a lot of places, just to constrain the tooltip body. can we make a shared component exported from the tooltip code?

just-toby (changes requested):

src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +30,176 @@ const PriceImpactText = styled.span`
...
-          onAcknowledgeNewTrade()
-        },
-        children: <Trans>Accept</Trans>,
+  const prevAllowanceRef = useRef(allowance)
+  // Ensures swap isn't submitted until allowance has been properly updated post-permit2 flow
+  useEffect(() => {
+    if (prevAllowanceRef.current.state === AllowanceState.REQUIRED && allowance.state === AllowanceState.ALLOWED) {
+      // Prevents swap if trade has updated mid permit2 flow
+      if (doesTradeDiffer) setReviewState(ReviewState.TRADE_CHANGED)
+      else triggerSwap()
```
just-toby, line 138: it feels wrong to me to call `triggerSwap` in the effect. is there a success callback from the approval/permit steps that we can use instead?

src/components/Swap/Toolbar/index.tsx
```diff
@@ -184,18 +180,6 @@ export default memo(function Toolbar() {
     return null
   }
 
-  if (!insufficientBalance) {
```
just-toby, line 187: please remove the unused variables from this file above

just-toby (changes requested):

src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +30,176 @@ const PriceImpactText = styled.span`
...
   const doesTradeDiffer = useMemo(
     () => Boolean(trade && ackTrade && tradeMeaningfullyDiffers(trade, ackTrade)),
     [ackTrade, trade]
   )
+  useEffect(() => {
+    if (doesTradeDiffer && reviewState === ReviewState.REVIEWING) {
+      setReviewState(ReviewState.TRADE_CHANGED)
+    } else if (!doesTradeDiffer && reviewState === ReviewState.TRADE_CHANGED) {
+      setReviewState(ReviewState.REVIEWING)
+    }
```
tinaszheng, line 117: 🤔 feels like moving back to the `REVIEWING` state should be part of the callback for trade acked instead of in this useEffect

src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +30,176 @@ const PriceImpactText = styled.span`
...
-          onSwapPriceUpdateAck?.(ackTrade, trade)
-          setAckTrade(trade)
-          // Prompts parent to show speedbump if new trade has high impact
-          onAcknowledgeNewTrade()
-        },
-        children: <Trans>Accept</Trans>,
+  const prevAllowanceRef = useRef(allowance)
+  // Ensures swap isn't submitted until allowance has been properly updated post-permit2 flow
+  useEffect(() => {
+    if (prevAllowanceRef.current.state === AllowanceState.REQUIRED && allowance.state === AllowanceState.ALLOWED) {
```
tinaszheng, line 135: hmm this `useEffect` took me a bit to understand - would it be more accurate to have comment update describe this as "automatically trigger signing swap tx if allowance requirements are met"

src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +30,176 @@ const PriceImpactText = styled.span`
...
-  }, [ackTrade, doesTradeDiffer, isPending, onAcknowledgeNewTrade, onSwapPriceUpdateAck, trade])
+  }, [allowance])
+
+  const onClick = useCallback(() => {
+    if (allowance.state === AllowanceState.REQUIRED) {
+      triggerPermit2Flow()
+      // if the user finishes permit2 allowance flow, triggerSwap() is called by useEffect above once state updates
+    } else if (allowance.state === AllowanceState.ALLOWED) {
+      triggerSwap()
+    }
```
tinaszheng, line 161: having all the trigger states be defined/used in separate places in this file is kind of confusing

thoughts on moving all the swap state/callbacks its own hook? it can be something like

`const { onStartSwapFlow, currentState } = useSwapCallback` that defines the useEffects here that manage allowance/swap tx state

and then this `onClick` here can just call `onStartSwapFlow`

tinaszheng (commented):
mostly looks good, just have a few suggestions on how to make the flow more readable. love the attention to detail on all the possible states, like if a trade updates during a permit flow 🤝

src/components/Swap/Summary/index.tsx
cartcrom: I originally had built it that way, but found that allowance.approveAndPermit would return successfully before onConfirm was updated to include the new allowance. I'm definitely open to a better pattern here, I'm just not sure if there's a cleaner one

src/components/Swap/Summary/index.tsx
cartcrom: Sure

src/components/Swap/Summary/index.tsx
cartcrom: true, will do

src/components/Swap/Summary/index.tsx
cartcrom: This should probably be done is sep refactor PR tho since it touches unrelated files

src/components/Swap/Summary/index.tsx
cartcrom: This is to catch a situation where the trade changes twice, where the first change causes tradeMeaningfullyDiffers to be true, but then the second change brings it back into a range where tradeMeaningfullyDiffers is false (acknowledgement no longer needed)

src/components/Swap/Summary/index.tsx
cartcrom: great point ty

src/components/Swap/Summary/index.tsx
just-toby: i actually did this in a PR of mine because i had the same need: `SmallToolTipBody`. you could use it here but it's not blocking

src/components/Tooltip.tsx
```diff
@@ -43,3 +46,17 @@ export default function Tooltip({
     </Popover>
   )
 }
+
+interface TooltipTextProps extends TooltipBaseProps {
+  text?: ReactNode
+}
+
+export function TooltipText({ text, children, placement = 'auto', offset, contained }: TooltipTextProps) {
```
just-toby, line 54: you could also just make this component constrain the width

src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +99,126 @@ const PriceImpactText = styled.span`
...
   onAcknowledgeNewTrade,
+  allowance,
 }: {
   trade: InterfaceTrade
   onConfirm: () => Promise<void>
   onAcknowledgeNewTrade: () => void
+  allowance: Allowance
 }) {
-  const { onSwapPriceUpdateAck, onSubmitSwapClick } = useAtomValue(swapEventHandlersAtom)
+  //const [reviewState, setReviewState] = useState(ReviewState.REVIEWING)
```
just-toby, line 165: remove ?

just-toby (approved):

src/components/Swap/Summary/index.tsx
tinaszheng, line 165: JAIL

src/components/Swap/Summary/index.tsx
```diff
@@ -19,6 +22,75 @@ import Summary from './Summary'
...
+    setCurrentState(ReviewState.REVIEWING)
+  }, [onConfirm, onSubmitSwapClick, trade])
+
+  const prevAllowanceRef = useRef(allowance)
+  // Automatically triggers signing swap tx if allowance requirements are met
+  useEffect(() => {
+    if (prevAllowanceRef.current.state === AllowanceState.REQUIRED && allowance.state === AllowanceState.ALLOWED) {
+      // Prevents swap if trade has updated mid permit2 flow
+      if (doesTradeDiffer) setCurrentState(ReviewState.TRADE_CHANGED)
+      else triggerSwap()
```
tinaszheng, line 63: whoa the if/else without brackets is spooky 🤔 but i dont hate it lol

src/components/Swap/Summary/index.tsx
```diff
@@ -19,6 +22,75 @@ import Summary from './Summary'
...
+  }, [allowance])
+
+  const onStartSwapFlow = useCallback(() => {
+    if (allowance.state === AllowanceState.REQUIRED) {
+      triggerPermit2Flow()
+      // if the user finishes permit2 allowance flow, triggerSwap() is called by useEffect above once state updates
+    } else if (allowance.state === AllowanceState.ALLOWED) {
+      triggerSwap()
+    }
+  }, [allowance, triggerPermit2Flow, triggerSwap])
```
tinaszheng, line 87: okie one other suggestion here:

so we have signing calls being triggered in both callbacks and in `useEffects` - maybe would be best to keep all the logic in one compact `useEffect` like:

```
const [swapFlowStarted, setSwapFlowStarted] = useState(false)

useEffect(() => {
   if (!swapFlowStarted) return 

   if (allowance.state === AllowanceState.REQUIRED) {
      triggerPermit2Flow()
    } else if (allowance.state === AllowanceState.ALLOWED) {
      if (tradeDoesNotDiffer) {
            triggerSwap()
       }
    }
}, [...])
const onStartSwapFlow = () => setSwapFlowStarted(true)
```

could probably get rid of the need to call `triggerSwap` on line 63

src/components/Swap/Summary/index.tsx
cartcrom: o tru

src/components/Swap/Summary/index.tsx
cartcrom: before I would've done `(doesTradeDiffer) ? setCurrentState(ReviewState.TRADE_CHANGED) : triggerSwap()` but you mentioned that means jail

tinaszheng: ok yes slight jail for that

src/components/Swap/Summary/index.tsx
```diff
@@ -27,59 +74,125 @@ const PriceImpactText = styled.span`
...
   const doesTradeDiffer = useMemo(
     () => Boolean(trade && ackTrade && tradeMeaningfullyDiffers(trade, ackTrade)),
     [ackTrade, trade]
   )
-
-  const [isPending, setIsPending] = useState(false)
-  const onClick = useCallback(async () => {
-    setIsPending(true)
+  const onSwap = useCallback(async () => {
+    console.log('here')
```
tinaszheng, line 143: dont forget to delet

tinaszheng (approved):
🔥

PR #453, opened 2023-02-03
fix: exit animations for dialogs

just-toby:
for useUnmountingAnimation to work, the animations need to be applied to the root of the subtree that is being unmounted
however, the animations need to be applied to the actual visible content that you want to be animated

this solution i landed on, updates `useUnmountingAnimation` to allow the animating element to be different than the unmounting element


cartcrom (approved):
LGTM, ty for fixing

PR #454, opened 2023-02-04
feat: small design updates

just-toby:
- reduce vertical spacing between elements in the widget
- reduce the height of the toolbar
- add price impact tooltips everywhere we show the value (only shown if warning is "error" or "warning")
  - use a new component for this. added tests for new component
- add token icons to summary page


tinaszheng (approved):
image

PR #455, opened 2023-02-05
feat: Add mixed routes support

tinaszheng:
Fixes WEB-970

for reference (when reviewing) [here's](https://github.com/Uniswap/mobile/pull/2680) my similar PR for mobile

*note this is branching off of tina/trades because it depends on changes in that branch. will update the main branch once tina/trades merges in*

This PR adds support for mixed route types from routing-api. To test this locally, add `forceMixedRoutes: true` to the `DEFAULT_QUERY_PARAMS` in [routing/slice.ts](https://github.com/Uniswap/widgets/blob/main/src/state/routing/slice.ts#L15)

The layout for routes for 3+ hops looks funky and needs to updated to match the [spec](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10900%3A111833&t=QhB8tYHZgm9L2p9O-0) but can be a separate PR that merges into this one.

Screenshot 2023-02-05 at 5 52 54 PM

### DIFF
--- a/src/state/routing/utils.ts
+++ b/src/state/routing/utils.ts
@@ -1,3 +1,4 @@
+import { MixedRouteSDK } from '@uniswap/router-sdk'
 import { Currency, CurrencyAmount, Token } from '@uniswap/sdk-core'
 import { Pair, Route as V2Route } from '@uniswap/v2-sdk'
 import { FeeAmount, Pool, Route as V3Route } from '@uniswap/v3-sdk'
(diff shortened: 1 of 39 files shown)


src/state/routing/utils.ts
```diff
@@ -90,6 +96,17 @@ export function transformQuoteToTradeResult(args: GetQuoteArgs, quoteResult: Quo
           inputAmount,
           outputAmount,
         })) ?? [],
+    mixedRoutes:
+      routes
+        ?.filter(
+          (r): r is typeof routes[0] & { mixedRoute: NonNullable<typeof routes[0]['mixedRoute']> } =>
+            r.mixedRoute !== null
+        )
+        .map(({ mixedRoute, inputAmount, outputAmount }) => ({
```
just-toby, line 105: what is the point of this `.map` operation? it looks like you're not actually changing the value before returning it?

just-toby (approved):

src/state/routing/utils.ts
tinaszheng: i think it's to remove the extraneous fields from each `route` object to match the Typescript types

PR #456, opened 2023-02-06
feat: move warnings into ActionButton

just-toby:
design spec: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=1708%3A26043&t=rBDEvcuI5V1h50oH-0

summary of change: "Moved most swap blocking errors back into the button text as they are today (Enter an amount, Insufficient liquidity etc)"

i.e. making the widget UI more like the current /swap page.

to achieve this I moved the `SwapActionButton` into the Toolbar file/component context so it's easier to reason about (since the display logic for the Toolbar and Action Button are very closely related), and so it's easier to test.

NOTE: still missing some test cases until I figure out how to mock the routing API

Screenshot 2023-02-06 at 2 02 15 PM
Screenshot 2023-02-06 at 1 53 42 PM
Screenshot 2023-02-06 at 1 53 35 PM
Screenshot 2023-02-06 at 1 53 29 PM
Screenshot 2023-02-06 at 1 53 18 PM


src/components/Swap/Toolbar/index.tsx
```diff
@@ -169,33 +168,68 @@ export default memo(function Toolbar() {
...
+          <ToolbarOrderRouting trade={trade} />
+        </Column>
+      </StyledExpando>
+    )
+  }, [caption, error, inputCurrency, isExpandable, maybeToggleOpen, open, outputCurrency, trade, tradeSummaryRows])
+
+  const getSwapActionButton = useMemo(() => {
+    if (insufficientBalance) {
+      return (
+        <ActionButton disabled={true}>
```
tinaszheng, line 205:
```suggestion
        <ActionButton disabled>
```

src/components/Swap/Toolbar/index.tsx
```diff
@@ -169,33 +168,68 @@ export default memo(function Toolbar() {
...
+      >
+        <Column>
+          <ToolbarTradeSummary rows={tradeSummaryRows} />
+          <ToolbarOrderRouting trade={trade} />
+        </Column>
+      </StyledExpando>
+    )
+  }, [caption, error, inputCurrency, isExpandable, maybeToggleOpen, open, outputCurrency, trade, tradeSummaryRows])
+
+  const getSwapActionButton = useMemo(() => {
```
tinaszheng, line 202: this isnt a getter function since it returns the value, right? can you rename to `swapActionButton`? or if you meant to make this a function you shoudl use `useCallback` instead

src/components/Swap/Toolbar/index.tsx
```diff
@@ -169,33 +168,68 @@ export default memo(function Toolbar() {
     return rows
   }, [gasUseEstimateUSD, impact?.percent, impact?.warning, slippage, trade])
 
-  if (inputCurrency == null || outputCurrency == null || error === ChainError.MISMATCHED_CHAINS) {
-    return null
-  }
+  const getCaptionRow = useMemo(() => {
```
tinaszheng, line 171: same for this one

src/components/Swap/Toolbar/index.tsx
tinaszheng: but also you dont necessarily need to create children components like this - you can instead just separate these out into their own functional components, which accomplishes the same thing

src/test/index.tsx
```diff
@@ -46,9 +51,19 @@ export function TestableWidget(props: PropsWithChildren<TestableWidgetProps>) {
             <ReduxProvider store={store}>
               <AtomProvider initialValues={props.initialAtomValues}>
                 <Web3Provider provider={hardhat.provider}>
-                  <BlockNumberProvider>
-                    <TokenListProvider list={TokenList.tokens}>{props.children}</TokenListProvider>
-                  </BlockNumberProvider>
+                  <ConnectorsProvider
```
tinaszheng, line 54: what's this for 🤔

tinaszheng (approved):
left some suggestions - overall lgtm

src/test/index.tsx
just-toby: this is just to avoid an invariant failing in the test environment - i added the first test that runs code using the ConnectorsProvider

PR #467, opened 2023-02-09
chore: remove permit2 flag

just-toby:
removes the flag since it's always true now


src/hooks/useSyncFlags.ts
```diff
@@ -1,37 +1,20 @@
...
 }
-
-export function usePermit2() {
-  const { chainId } = useWeb3React()
-  const permit2 = useAtomValue(flagsAtom).permit2 ?? false
-  try {
-    // Detect if the Universal Router is not yet deployed to chainId.
-    // This is necessary so that we can fallback correctly on chains without a Universal Router deployment.
-    // It will be removed once Universal Router is deployed on all supported chains.
-    chainId && UNIVERSAL_ROUTER_ADDRESS(chainId)
```
zzmp, line 32: This is not always true, and will crash on some test nets.
https://github.com/Uniswap/widgets/pull/388 must be finished and merged before this flag is removed.

zzmp (changes requested):

just-toby:
i'll just re-open this when the time is right

PR #468, opened 2023-02-09
fix: swallow transaction rejected error

just-toby:
stop throwing `SwapError`s when the user rejects the Transaction. instead, if we detect that the user rejected, just notify the product code via the return value. for now, we swallow the error and just revert the UI so they can try again. but we may want to show a message inline in this situation

https://uniswaplabs.atlassian.net/browse/WEB-2888?atlOrigin=eyJpIjoiYTg1YWFiZDYwMGM2NDc3NWJmOGRiZGE3NGNmNGQyZTQiLCJwIjoiaiJ9

### DIFF
--- a/src/components/Swap/SwapActionButton/SwapButton.tsx
+++ b/src/components/Swap/SwapActionButton/SwapButton.tsx
@@ -69,7 +69,7 @@ export default function SwapButton({ disabled }: { disabled: boolean }) {
   const throwAsync = useAsyncError()
   const onSwap = useCallback(async () => {
     try {
-      await onSubmit(async () => {
+      const submitted = await onSubmit(async () => {
         const response = await swapCallback?.()
         if (!response) return
 
@@ -90,7 +90,9 @@ export default function SwapButton({ disabled }: { disabled: boolean }) {
       })
 
       // Only close the review modal if the swap submitted (ie no-throw).
-      setOpen(false)
+      if (submitted) {
+        setOpen(false)
+      }
     } catch (e) {
       throwAsync(e)
     }
--- a/src/components/Swap/SwapActionButton/useOnSubmit.ts
+++ b/src/components/Swap/SwapActionButton/useOnSubmit.ts
@@ -6,16 +6,20 @@ import { displayTxHashAtom, Field } from 'state/swap'
 import { TransactionInfo, TransactionType } from 'state/transactions'
 import { isAnimating } from 'utils/animations'
 
-/** Submits a transaction. */
+/**
+ * Returns a callback to submit a transaction.
+ *
+ * Returns a boolean indicating whether the transaction was submitted.
+ * */
 export default function useOnSubmit() {
   const addTransactionInfo = useAddTransactionInfo()
   const setDisplayTxHash = useUpdateAtom(displayTxHashAtom)
   const [, setInputAmount] = useSwapAmount(Field.INPUT)
 
   return useCallback(
-    async (submit: () => Promise<TransactionInfo | void>): Promise<void> => {
+    async (submit: () => Promise<TransactionInfo | void>): Promise<boolean> => {
       const info = await submit()
-      if (!info) return
+      if (!info) return false
 
       addTransactionInfo(info)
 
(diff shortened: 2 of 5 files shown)


src/components/Swap/SwapActionButton/useOnSubmit.ts
```diff
@@ -6,16 +6,20 @@ import { displayTxHashAtom, Field } from 'state/swap'
 import { TransactionInfo, TransactionType } from 'state/transactions'
 import { isAnimating } from 'utils/animations'
 
-/** Submits a transaction. */
+/**
+ * Returns a callback to submit a transaction.
+ *
+ * Returns a boolean indicating whether the transaction was submitted.
```
zzmp, line 12: Could you also specify that this will throw (and not just return false) if something besides a user rejection occurs?

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -12,11 +22,7 @@ export function swapErrorToUserReadableMessage(error: any): string {
     }
   }
 
-  let reason: string | undefined
```
zzmp, line 15: Can this be done in a separate PR? It expands the scope of this change.

src/utils/swapErrorToUserReadableMessage.tsx
```diff
@@ -12,11 +22,7 @@ export function swapErrorToUserReadableMessage(error: any): string {
     }
   }
```
zzmp, line 23: Can this initial check (for USER_REJECTED_REQUEST) be removed? Or are you keeping it for the deprecated non-UniversalRouter codepath?

src/hooks/useUniversalRouter.ts
```diff
@@ -56,7 +63,21 @@ export function useUniversalRouterSwapCallback(trade: InterfaceTrade | undefined
       }
       const gasLimit = calculateGasMargin(gasEstimate)
       response = await provider.getSigner().sendTransaction({ ...tx, gasLimit })
-    } catch (swapError: unknown) {
+    } catch (swapError) {
+      const reason = getReason(swapError)
+      if (
```
zzmp, line 68: Can this logic be moved to a helper function to improve readability/maintainability?

src/hooks/useUniversalRouter.ts
```diff
@@ -21,10 +22,16 @@ interface SwapOptions {
...
+ * Returns a callback to submit a transaction to the universal router.
+ *
+ * The callback returns the TransactionResponse if the transaction was submitted,
+ * or undefined if the user rejected the transaction.
+ **/
 export function useUniversalRouterSwapCallback(trade: InterfaceTrade | undefined, options: SwapOptions) {
   const { account, chainId, provider } = useWeb3React()
 
-  return useCallback(async (): Promise<TransactionResponse> => {
+  return useCallback(async (): Promise<TransactionResponse | undefined> => {
```
zzmp, line 34: nit: I think it would be clearer if you returned something concrete, not `undefined`. Could you create a `Symbol('Rejected transaction')` and return that instead? Or even just return `false` or `null`?

zzmp (commented):
Comments to improve readability - none of them are actually blocking, code looks good 😄

src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -90,7 +90,9 @@ export default function SwapButton({ disabled }: { disabled: boolean }) {
       })
 
       // Only close the review modal if the swap submitted (ie no-throw).
-      setOpen(false)
+      if (submitted) {
```
zzmp, line 93: I think @infredible has an opinion (for a future design) on what the UX should be for user rejection, it may be good to check in.

src/utils/swapErrorToUserReadableMessage.tsx
just-toby: i think we should keep it so that `swapErrorToUserReadableMessage` is complete - this is the message it should return in this case.

src/utils/swapErrorToUserReadableMessage.tsx
just-toby: we need this change to actually detect user rejection. the change in this file is just a small refactor so i don't duplicate code.

tinaszheng (approved):

PR #469, opened 2023-02-09
feat: new token logo component / sourcing

cartcrom:
Adds in a heavily refactored version of the token Logo component & caching table from interface. Upgraded to lazy parse all possible sources for a given token and to not re-check previously failed uris.

The widgets package now exports the Logo component & utility for interface to use. Widgets and interface can share uris via the <LogoUpdater> component, which allows passing in new URIs from a variety of sources (tokenlists, gql queries, etc)

image


src/index.tsx
```diff
@@ -3,6 +3,7 @@ import 'polyfills'
 import Swap, { SwapProps } from 'components/Swap'
 import Widget, { WidgetProps } from 'components/Widget'
 
+export * from './components/Logo'
```
zzmp, line 6: Can you explicitly export what you intend to instead of using `*`, to match the other exports?

src/components/Logo/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  width: ${({ size }) => size};
+`
+
+export type LogoBasePops = {
+  symbol?: string | null
+  backupImg?: string | null
+  size?: string
+  style?: React.CSSProperties
+}
+// type LogoProps = LogoBasePops & { isNative?: boolean; address?: string | null; chainId?: number }
```
just-toby, line 37: remove?

src/components/Logo/LogoTable.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  constructor(uri: string) {
+    this.key = uri
+    this.unparsedUri = uri
+  }
+
+  getUri() {
+    // Lazy-parse the address
+    if (this.parsedUri === null) {
+      ;[this.parsedUri, this.alternateUri] = uriToHttp(this.unparsedUri)
```
just-toby, line 27: remove semicolon?

just-toby: if this is intentional , i'm not familiar with the syntax. so just checking

just-toby (commented):
i think we should add unit tests for `LogoTable`

just-toby:
besides comments above this lgtm

src/components/Logo/LogoTable.ts
cartcrom: the linter won't allow this w/out semi colon. Do you know if theres a cleaner way do destructure into existing vars?

src/index.tsx
cartcrom: 🫡

src/components/Logo/LogoTable.ts
just-toby: maybe just using local variables? i've never seen this syntax before but that doesn't mean it's bad

just-toby (approved):

PR #470, opened 2023-02-09
fix: use the router API for RouterPreference.PRICE

just-toby:
updates the logic in `src/state/routing/slice.ts` to allow `RouterPreference.PRICE` requests to go through the Router API. There is no need for these to be client-only afaik.

this fixes a bug where the native token price isn't appearing (because we're using "ETH" as the token address).

the follow up is to fix the client-side router to support native tokens.

all other code changes in this PR are just boilerplate to set up a context for the `routerUrl`, so we don't have to pass it around as a prop.


src/hooks/swap/useRouterUrl.tsx
```diff
@@ -0,0 +1,11 @@
+import { createContext, PropsWithChildren, useContext } from 'react'
+
+const Context = createContext<string | undefined>(undefined)
+
+export function RouterUrlProvider({ children, routerUrl }: PropsWithChildren<{ routerUrl?: string }>) {
```
tinaszheng, line 5: we already have the `swapRouterUrlAtom` so you could use that instead of doing the context

seems more consistent bcuz we store other main props in atoms ^

src/state/routing/slice.ts
```diff
@@ -33,7 +33,7 @@ export const routing = createApi({
 
         if (
           // If enabled, try the routing API, falling back to client-side routing.
-          args.routerPreference === RouterPreference.API &&
+          (args.routerPreference === RouterPreference.API || args.routerPreference === RouterPreference.PRICE_API) &&
```
tinaszheng, line 36: thoughts on separating the concept of `price` query into its own argument? i feel like conflating routerPreference for both routerUrl and also use case is kinda weird

we could have `routerPreference` just reference either API | Client (and other | Gouda)

and then have a separate argument called quote type - `SwapQuote | PriceQuote | skip` and the polling interval can be determined by just that

tinaszheng (changes requested):
we should probs separate routerPreference vs quote type, but lmk if u disagree or is out of scope for this PR

src/state/routing/slice.ts
just-toby: yeah that's fair, let me try again

src/hooks/routing/types.ts
```diff
@@ -1,10 +1,19 @@
...
+export enum QuoteType {
+  TRADE = 'trade',
+  PRICE = 'price',
+  SKIP = 'skip',
+}
+
+export type QuoteConfig =
+  | { type: QuoteType.TRADE; preference: RouterPreference }
+  | { type: QuoteType.PRICE; preference: RouterPreference }
+  | { type: QuoteType.SKIP; preference?: undefined }
```
zzmp, line 15: Can this just be `| { type: QuoteType.SKIP }`? Why include the (double-)undefined `preference`?

src/hooks/routing/types.ts
```diff
@@ -1,10 +1,19 @@
 export enum RouterPreference {
-  PRICE,
-  SKIP,
   API,
```
zzmp, line 2: I really like how you used string-equivalents for the new `enum QuoteType`. It eliminates an ambiguity of default values because `0` is falsey, so you can't check for `RouterPreference | undefined` with a boolean check.

Would you add explicit string values for `enum RouterPreference` too?

src/hooks/routing/useRouterTrade.ts
```diff
@@ -29,32 +30,39 @@ export function useRouterTrade(
...
   const queryArgs = useGetQuoteArgs(
-    { provider, tradeType, amountSpecified, currencyIn, currencyOut, routerPreference, routerUrl },
-    /*skip=*/ routerPreference === RouterPreference.SKIP
+    {
+      provider,
+      tradeType,
+      amountSpecified,
+      currencyIn,
+      currencyOut,
+      routerPreference: quoteConfig.preference,
```
zzmp, line 48: At this point, could you just pass `quoteConfig` as the second arg (instead of passing `routerPreference` or `skip`), to keep it more semantic?

src/hooks/routing/useRouterTrade.ts
```diff
@@ -29,32 +30,39 @@ export function useRouterTrade(
...
-  routerPreference: RouterPreference,
-  routerUrl?: string
+  quoteConfig: QuoteConfig
 ): {
   state: TradeState
   trade?: InterfaceTrade
   gasUseEstimateUSD?: CurrencyAmount<Token>
 } {
   const { provider } = useWeb3React()
+  const routerUrl = useAtomValue(swapRouterUrlAtom)
```
zzmp, line 40: This can be moved to `useGetQuoteArgs`

zzmp (approved):
I have small changes requested, but approved assuming those changes are made.
I'm happy to look again after you make them if you'd like.

PR #472, opened 2023-02-10
feat: permit tx pending state

cartcrom:
Adds a state for when the user has finished approving/signing in their wallet, but the permit tx is still pending
image

### DIFF
--- a/src/components/Swap/Summary/index.tsx
+++ b/src/components/Swap/Summary/index.tsx
@@ -25,7 +25,7 @@ export default Summary
 
 enum ReviewState {
   REVIEWING,
-  ALLOWANCE_PROMPTED,
+  ALLOWING,
   ALLOWANCE_FAILED,
   TRADE_CHANGED,
   SWAP_PENDING,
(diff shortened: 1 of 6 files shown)


src/components/Swap/Summary/index.tsx
```diff
@@ -25,18 +25,18 @@ export default Summary
 
 enum ReviewState {
   REVIEWING,
-  ALLOWANCE_PENDING,
+  ALLOWANCE_PROMPTED,
```
just-toby, line 28: i'm having a hard time following this name. does this just mean any state before the allowance is fully set? i.e. awaiting permit2 approval, awaiting confirmation of permit2 approval on chain, or awaiting permit signature?

just-toby (approved):
lgtm but i think some of these names could be improved to make clearer exactly what we are awaiting - the permit2 contract, the signature, etc

src/components/Swap/Summary/index.tsx
cartcrom: I can change it to something else, prompted made sense at one point in my refactor but probably doesn't anymore. But yeah it's basically just any allowance state. I didn't want to call it pending anymore because I felt that could be confused with the approvalLoading variable... which isn't 1-1 with this state.

PR #473, opened 2023-02-10
fix: re-enable cd

just-toby:
remove the manual trigger, re-enable CD

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,11 +1,9 @@
 name: Release
 
 on:
-  # push:
-  #   branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -1,11 +1,9 @@
 name: Release
 
 on:
-  # push:
-  #   branches:
-  #     - main
-  # manual trigger
```
zzmp, line 7: You should retain the manual trigger just in case (eg if push trigger fails, but you want to retry).

tinaszheng (approved):

PR #474, opened 2023-02-10
fix: circular dependency

just-toby:
fixes a circular dependency and adds `yarn build` to the GH "test" action

### DIFF
--- a/src/components/Swap/SwapActionButton/SwapButton.tsx
+++ b/src/components/Swap/SwapActionButton/SwapButton.tsx
@@ -19,7 +19,7 @@ import invariant from 'tiny-invariant'
 import ActionButton from '../../ActionButton'
 import Dialog from '../../Dialog'
 import { SummaryDialog } from '../Summary'
-import { useCollapseToolbar } from '../Toolbar'
+import { useCollapseToolbar } from '../Toolbar/ToolbarContext'
 import useOnSubmit from './useOnSubmit'
 
 /**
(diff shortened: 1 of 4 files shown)


src/components/Swap/SwapActionButton/SwapButton.tsx
```diff
@@ -19,7 +19,7 @@ import invariant from 'tiny-invariant'
 import ActionButton from '../../ActionButton'
 import Dialog from '../../Dialog'
 import { SummaryDialog } from '../Summary'
-import { useCollapseToolbar } from '../Toolbar'
+import { useCollapseToolbar } from '../Toolbar/ToolbarContext'
```
zzmp, line 22: Instead of importing this, which although is not technically circular still seems circular (Toolbar iimport SwapButton imports Toolbar/ToolbarContext), could you pass the collapse toolbar callback as a prop (`onBeforeReviewSwap`?) to keep these components more distinct?

I'm also fine with this as a fast follow-up, so that you can implement a minimal change to fix the build _first_, and _then_ clean up the implementation.

zzmp (changes requested):

src/components/Swap/SwapActionButton/SwapButton.tsx
just-toby: i see your point but want to discuss this before making the change, so i'll do it in a follow up PR

zzmp (approved):

PR #479, opened 2023-02-14
feat: dialog no animation

just-toby:
adds a new option to `DialogAnimationType` to allow for *no* animation.

this would allow integrators to animation the dialog however they want, at the container level

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -203,10 +203,12 @@ const fadeAnimationCss = css`
   }
 `
 
+const EMPTY_CSS = css``
+
 const getAnimation = (animationType?: DialogAnimationType) => {
   switch (animationType) {
     case DialogAnimationType.NONE:
-      return css``
+      return EMPTY_CSS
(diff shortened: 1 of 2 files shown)


src/components/Dialog.tsx
```diff
@@ -202,9 +203,21 @@ const fadeAnimationCss = css`
   }
 `
 
-const AnimationWrapper = styled.div<{ animationType: DialogAnimationType }>`
+const getAnimation = (animationType?: DialogAnimationType) => {
+  switch (animationType) {
+    case DialogAnimationType.NONE:
+      return css``
```
JFrankfurt, line 209: can we return a build-time constant here that won't re-run the template string function call when `getAnimation` runs?

JFrankfurt (approved):

src/components/Dialog.tsx
JFrankfurt: (just shift the empty css`` call up out of the fn)

PR #481, opened 2023-02-15
fix: settings layout and behavior fixes

just-toby:
fixes the settings padding and adds a click handler to close the settings modal when the user clicks away


src/components/Swap/Settings/index.tsx
```diff
@@ -53,6 +58,7 @@ const SettingsButton = styled(IconButton)`
 export default function Settings() {
   const [open, setOpen] = useState(false)
   const [wrapper, setWrapper] = useState<HTMLDivElement | null>(null)
+  useOutsideClickHandler(wrapper, () => setOpen(false))
```
zzmp, line 62: nit: no newline needed

src/hooks/useOutsideClickHandler.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function useOutsideClickHandler(node: HTMLDivElement | null, onOutsideClick: () => void) {
+  const handleClickOutside = (event: MouseEvent) => {
+    if (node && !node.contains(event.target as Node)) {
+      onOutsideClick()
+    }
+  }
+
+  useEffect(() => {
+    document.addEventListener('mousedown', handleClickOutside)
```
zzmp, line 11: Do you have concerns that many uses of this hook will cause a lot of global mousedown handlers and eventual performance issues?

src/hooks/useOutsideClickHandler.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { fireEvent, render, renderHook } from 'test'
+
+import { useOutsideClickHandler } from './useOutsideClickHandler'
+
+describe('useOutsideClickHandler', () => {
+  it('should call onOutsideClick when clicked outside of node', () => {
```
zzmp, line 6: Move shared lines to a `beforeEach`.

```suggestion
  let node: HTMLDivElement
  beforeEach(() => {
    node = render(<div />)
    renderHook(() => useOutsideClickHandler(node.container as HTMLDivElement, onOutsideClick))
  })

  it('should call onOutsideClick when clicked outside of node', () => {
```

zzmp (approved):
This lgtm unless you think the global handlers will be an issue. I think it should be fine.

src/hooks/useOutsideClickHandler.ts
just-toby: tbh, i let copilot write this for me 😬   ..... i can optimize it for sure

just-toby:
added jira task to my backlog for optimizing the outside click handler

PR #482, opened 2023-02-15
fix: make price impact more consistent

just-toby:
addresses some [design feedback](https://www.notion.so/uniswaplabs/Widget-design-feedback-835bfad85aac4ad3a00550fcea18bf99) about the formatting and styling of the price impact.

- remove parens from the % in the review screen
- use `toString` everywhere to ensure consistent values


cartcrom (approved):

cartcrom:
@callil is this a change that should be product-wide / should mobile also do this? we could make a change to coned

PR #484, opened 2023-02-15
feat: loading state for toolbar Trade

just-toby:
adds a different loading state to the `Trade` caption in the toolbar. instead of replacing all the text with "Fetching Trade", we:
- add a loading spinner next to the exchange rate
- make the exchange rate 0.4 opacity

### DIFF
--- a/src/components/Swap/Toolbar/Caption.tsx
+++ b/src/components/Swap/Toolbar/Caption.tsx
@@ -155,7 +155,7 @@ export function Trade({
     <>
       <Caption
         caption={
-          <ThemedText.Body2 color="secondary" opacity={loading ? 0.4 : 1}>
+          <ThemedText.Body2 opacity={loading ? 0.4 : 1}>
             <Price trade={trade} outputUSDC={outputUSDC} />
           </ThemedText.Body2>
         }
--- a/src/components/Swap/Toolbar/Caption.tsx
+++ b/src/components/Swap/Toolbar/Caption.tsx
@@ -48,10 +48,11 @@ interface CaptionProps {
   icon?: Icon | null
   caption: ReactNode
   color?: Color
+  loading?: boolean
   tooltip?: CaptionTooltip
 }
 
-function Caption({ icon: Icon, caption, color = 'secondary', tooltip }: CaptionProps) {
+function Caption({ icon: Icon, caption, color = 'secondary', tooltip, loading }: CaptionProps) {
   return (
     <CaptionRow gap={0.5} shrink={0}>
       {tooltip ? (
@@ -61,7 +62,9 @@ function Caption({ icon: Icon, caption, color = 'secondary', tooltip }: CaptionP
       ) : (
         Icon && <LargeIcon icon={Icon} color={color} />
       )}
-      <ThemedText.Body2 color={color}>{caption}</ThemedText.Body2>
+      <ThemedText.Body2 color={color} opacity={loading ? 0.4 : 1}>
+        {caption}
+      </ThemedText.Body2>
     </CaptionRow>
   )
 }
@@ -132,6 +135,7 @@ export function Wrap({ inputCurrency, outputCurrency }: WrapProps) {
 
 export interface TradeProps {
   trade: InterfaceTrade
+  loading: boolean
   outputUSDC?: CurrencyAmount<Currency>
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -61,7 +62,9 @@ function Caption({ icon: Icon, caption, color = 'secondary', tooltip }: CaptionP
       ) : (
         Icon && <LargeIcon icon={Icon} color={color} />
       )}
-      <ThemedText.Body2 color={color}>{caption}</ThemedText.Body2>
+      <ThemedText.Body2 color={color} opacity={loading ? 0.4 : 1}>
```
tinaszheng, line 65: are there design tokens for opacity? 🤔

src/components/Swap/Toolbar/index.tsx
```diff
@@ -61,7 +61,7 @@ function CaptionRow() {
       default:
     }
 
-    if (state === TradeState.LOADING) {
+    if (state === TradeState.LOADING && trade == null) {
```
tinaszheng, line 64: hmm is it intentional to not show the loading state on subsequent fetches? i think as a user i'd expect there to be a loading state if i changed my input amount or currency, for example 🤔

just-toby: see the PR description or deploy preview - we do have a loading state, newly added to the `Trade` component below

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -143,10 +147,20 @@ const Expander = ({ expanded }: ExpandProps) => {
...
+  outputUSDC,
+  gasUseEstimateUSD,
+  expanded,
+  loading,
+}: TradeProps & TradeTooltip & ExpandProps) {
   return (
     <>
-      <Caption caption={<Price trade={trade} outputUSDC={outputUSDC} />} />
+      <Caption
+        caption={<Price trade={trade} outputUSDC={outputUSDC} />}
```
zzmp, line 160: Instead of generalizing to pass `loading` to `Caption` when it is only used by `Trade`, would you consider instead passing `<ThemedText.Body2 opacity={loading ? 0.4 : 1}><Price ... /></ThemedText.Body2>` as the `caption` prop?

src/components/Swap/Toolbar/index.tsx
```diff
@@ -61,7 +61,7 @@ function CaptionRow() {
...
       return { caption: <Caption.LoadingTrade gasUseEstimateUSD={gasUseEstimateUSD} /> }
     }
```
zzmp, line 67: L68 is not the same check as L64 (`trade !== null`), so this will not necessarily always be triggered. Should L68 be changed?

src/components/Swap/Toolbar/index.tsx
zzmp, line 64: You should always use `===` to avoid bugs from implicit casting. (I thought this was a lint rule 🤷 )

zzmp (changes requested):

src/components/Swap/Toolbar/index.tsx
just-toby: i actually wanted `==` because i'm checking for `undefined`

just-toby: but i can make it explicitly check for `undefined`

src/components/Swap/Toolbar/index.tsx
just-toby: no, this is intentional. we only want to show `Caption.LoadingTrade` if it's the first time a trade is loading. after that, we want to show the previously loaded trade while fetching the updated quote.

src/components/Swap/Toolbar/Caption.tsx
just-toby: works for me! i'm glad these can be nested.

src/components/Swap/Toolbar/index.tsx
zzmp: put differently, is there a case where `state === TradeState.LOADING && trade !== undefined && !(inputCurrency && outputCurrency && isAmountPopulated)`, where `Caption.Trade` should be displayed but will not?

It's hard to tell, because you're using two different checks. I think you should use the same check (negated) for these two conditionals.

just-toby: @zzmp so i think it does make sense to add an `if (trade)` check to the `Caption.Trade` case. we do expect that to be defined there (L68 in your original comment). but otherwise the checks should be different

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -143,10 +144,23 @@ const Expander = ({ expanded }: ExpandProps) => {
...
+  gasUseEstimateUSD,
+  expanded,
+  loading,
+}: TradeProps & TradeTooltip & ExpandProps) {
   return (
     <>
-      <Caption caption={<Price trade={trade} outputUSDC={outputUSDC} />} />
+      <Caption
+        caption={
+          <ThemedText.Body2 color="secondary" opacity={loading ? 0.4 : 1}>
```
zzmp, line 158: I don't think you need to specify color here, i think it will inherit, but i may be wrong

src/components/Swap/Toolbar/index.tsx
```diff
@@ -61,7 +61,7 @@ function CaptionRow() {
       default:
     }
 
-    if (state === TradeState.LOADING) {
+    if (state === TradeState.LOADING && trade === undefined) {
```
zzmp, line 64: Can this be made '&& !trade' to better mirror the next check? This way if types change we'd know all cases are still covered here.

zzmp (approved):

PR #487, opened 2023-02-16
feat: widget page-centered dialogs

just-toby:
adds an option to the Widget API which enables page-centered dialogs without any configuration. 

if they are enabled, we mount our dialogs at the page root and make them full screen. if Dialogs are page-centered, then their content needs to define their own size rather than assume the parent's size will be defined. (see my adjustments to the `SummaryDialog` and `TokenOptions` for examples).

note: i'm currently forcing errors to stay centered in the widget because it's a better UX imo.

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -258,6 +258,9 @@ const FullScreenWrapper = styled.div<{ enabled?: boolean }>`
       width: 100%;
       z-index: ${Layer.DIALOG};
     `}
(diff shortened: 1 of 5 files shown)


cartcrom (commented):
Everything looks great, just wondering if we can fix the corners in this animation

https://user-images.githubusercontent.com/39385577/219765457-f8a313da-1cb8-4a06-b202-3297ba9eb567.mov

cartcrom (approved):

PR #491, opened 2023-02-21
fix: Handle no routes state in toolbar

tinaszheng:
We are currently catching the no routes state as "missing input" instead of displaying a trade error


just-toby (approved):
yup

PR #495, opened 2023-02-22
fix: remove info icon from wrap caption

just-toby:
according to design this should be removed


src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -127,7 +127,7 @@ export function Wrap({ inputCurrency, outputCurrency }: WrapProps) {
     [inputCurrency.symbol, isWideWidget, outputCurrency.symbol]
   )
 
-  return <Caption icon={Info} caption={<Text />} />
+  return <Caption icon={null} caption={<Text />} />
```
vm, line 130: can we just make icon optional instead of passing null?

just-toby: good catch, it can just be omitted here

vm (approved):

PR #496, opened 2023-02-22
fix: border and layout issues for status dialog

just-toby:
fixes some incorrect layout for the status dialog

### DIFF
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -66,8 +66,6 @@ export function Provider({ value, children, options }: ProviderProps) {
   // If a Dialog is active, mark the main content inert
   const ref = useRef<HTMLDivElement>(null)
   const [active, setActive] = useState(false)
-  // If pageCentered dialogs are enabled, we should mount them directly to the document body
-  // and adjust the styling in the renderer below so that they are centered and sized correctly.
   const context = { element: value, active, setActive, options }
   useEffect(() => {
     if (ref.current) {
@@ -198,18 +196,18 @@ const fadeOut = keyframes`
   }
 `
 
-const HiddenWrapper = styled.div<{ hidden?: boolean; constrain?: boolean }>`
+const HiddenWrapper = styled.div<{ hideOverflow?: boolean; constrain?: boolean }>`
   border-radius: ${({ theme }) => theme.borderRadius.medium}em;
   height: ${({ constrain }) => (constrain ? 'fit-content' : '100%')};
   left: 0;
 
-  overflow: ${({ hidden }) => (hidden ? 'hidden' : 'visible')};
+  overflow: ${({ hideOverflow }) => (hideOverflow ? 'hidden' : 'visible')};
   position: ${({ constrain }) => (constrain ? 'relative' : 'absolute')};
   top: 0;
   width: ${({ constrain }) => (constrain ? 'fit-content' : '100%')};
 
   @supports (overflow: clip) {
-    overflow: ${({ hidden }) => (hidden ? 'clip' : 'visible')};
+    overflow: ${({ hideOverflow }) => (hideOverflow ? 'clip' : 'visible')};
   }
 `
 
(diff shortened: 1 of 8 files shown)


src/components/Dialog.tsx
```diff
@@ -66,8 +66,6 @@ export function Provider({ value, children, options }: ProviderProps) {
   // If a Dialog is active, mark the main content inert
   const ref = useRef<HTMLDivElement>(null)
   const [active, setActive] = useState(false)
-  // If pageCentered dialogs are enabled, we should mount them directly to the document body
-  // and adjust the styling in the renderer below so that they are centered and sized correctly.
```
JFrankfurt, line 70: confirming this comment isn't relevant any longer

src/components/Dialog.tsx
```diff
@@ -163,6 +161,7 @@ export const Modal = styled.div<{ color: Color; constrain?: boolean }>`
   height: ${({ constrain }) => (constrain ? 'fit-content' : '100%')};
   left: 0;
   outline: ${({ theme }) => `1px solid ${theme.outline}`};
+  outline: transparent;
```
JFrankfurt, line 164: the second outline value here seems wrong

src/components/Dialog.tsx
```diff
@@ -331,11 +335,8 @@ export default function Dialog({ color, children, onClose, forceContain }: Dialo
       <ThemeProvider>
         <PopoverBoundaryProvider value={popoverRef.current} updateTrigger={updatePopover}>
           <div ref={popoverRef}>
-            <FullScreenWrapper enabled={pageCentered} onClick={onClose}>
-              <HiddenWrapper
-                constrain={pageCentered}
-                hidden={context.options?.animationType === DialogAnimationType.SLIDE}
-              >
+            <FullScreenWrapper enabled={pageCentered} onClick={pageCentered ? onClose : () => null}>
```
JFrankfurt, line 338: I think this should continue to just be `onClose` and if you need a boolean behavior switch it should happen in that handler (or a wrapper, if that's a more applicable implementation)

src/components/Swap/Status/StatusDialog.tsx
```diff
@@ -34,32 +34,22 @@ function TransactionStatus({ tx, onClose }: TransactionStatusProps) {
...
-    }
-    return null
-  }, [tx.receipt?.status])
-
   return (
-    <Column flex padded gap={0.75} align="stretch" style={{ height: '100%' }} data-testid="status-dialog">
+    <Column flex padded align="stretch" style={{ height: '100%', marginTop: '3em' }} data-testid="status-dialog">
       <StatusHeader icon={Icon} iconColor={tx.receipt?.status ? 'success' : undefined}>
-        <ThemedText.H4>{heading}</ThemedText.H4>
+        <ThemedText.H4 margin="3em 0 0">{heading}</ThemedText.H4>
```
JFrankfurt, line 40: nit: pref `styled(Component)` wrapping over inline style

src/components/Dialog.tsx
```diff
@@ -320,7 +323,8 @@ export default function Dialog({ color, children, onClose, forceContain }: Dialo
           return (mountPoint?.childElementCount ?? 0) > 1 ? SlideAnimationType.PAGING : SlideAnimationType.CLOSING
       }
     },
-    modal
+    modal,
+    context.options?.animationType === DialogAnimationType.NONE
```
JFrankfurt, line 327: nit: would prefer not doing the boolean evaluation here and instead assign it to a variable and pass that it

JFrankfurt (changes requested):

src/components/Dialog.tsx
just-toby: just wanted to move it, sorry for the confusion

src/components/Dialog.tsx
just-toby: thx

src/components/Dialog.tsx
just-toby: done

src/components/Swap/Status/StatusDialog.tsx
just-toby: agree, but `ThemedText.H4` would overwrite it unless you use the prop - this is what the prop is intended for (similar to `Row`)

src/components/Dialog.tsx
just-toby: updated

JFrankfurt (approved):
LGTM

PR #498, opened 2023-02-22
feat: update copy in review screen

just-toby:
remove the "after slippage" text from a few places


src/components/Swap/Summary/Estimate.tsx
```diff
@@ -19,11 +20,19 @@ interface EstimateProps {
 }
 
 export default function SwapInputOutputEstimate({ trade, slippage }: EstimateProps) {
-  const { estimateMessage } = useMemo(() => getEstimateMessage(trade, slippage), [slippage, trade])
+  const { estimateMessage } = useMemo(() => getEstimateMessage(trade, slippage, undefined), [slippage, trade])
```
lynnshaoyu, line 23: nit: can we put an inline comment here in front of undefined like `/** priceImpact */`?

also might be worth renaming the var to priceImpact for clarity?

lynnshaoyu (approved):
just one optional nit

PR #500, opened 2023-02-23
fix: tooltip and modal updates

just-toby:
- remove the manual trigger from `Popover`, instead automatically update like we do in `interface` ([link](https://github.com/Uniswap/interface/blob/main/src/components/Popover/index.tsx#L117))

- change the min width of dialogs to not overflow small width screens

### DIFF
--- a/src/components/Popover.tsx
+++ b/src/components/Popover.tsx
@@ -1,6 +1,7 @@
 import { Options, Placement } from '@popperjs/core'
 import { globalFontStyles } from 'css/font'
 import useInterval from 'hooks/useInterval'
+import ms from 'ms.macro'
 import maxSize from 'popper-max-size-modifier'
 import React, { createContext, PropsWithChildren, useCallback, useContext, useMemo, useRef, useState } from 'react'
 import { createPortal } from 'react-dom'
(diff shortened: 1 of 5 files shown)


src/components/Popover.tsx
```diff
@@ -149,10 +151,11 @@ export default function Popover({
...
 
-  // Manually triggers an update, if prop is provided
-  useEffect(() => {
-    update?.()
-  }, [update, updateTrigger])
+  const updateCallback = useCallback(() => {
+    update && update()
+  }, [update])
+
+  useInterval(updateCallback, show ? 100 : null)
```
cartcrom, line 158: nit: could you use ms`0.1s` here for readability?

src/components/Popover.tsx
```diff
@@ -1,25 +1,27 @@
...
-import React, { createContext, PropsWithChildren, useContext, useEffect, useMemo, useRef, useState } from 'react'
+import React, { createContext, PropsWithChildren, useCallback, useContext, useMemo, useRef, useState } from 'react'
 import { createPortal } from 'react-dom'
 import { usePopper } from 'react-popper'
 import styled from 'styled-components/macro'
 import { AnimationSpeed, Layer } from 'theme'
 
-type PopoverBoundary = { boundary?: HTMLDivElement | null; updateTrigger?: any }
+type PopoverBoundary = { boundary?: HTMLDivElement | null }
 const BoundaryContext = createContext<PopoverBoundary>({})
```
cartcrom, line 13: what I meant in my last review's comment abt the provider is that we can just revert back to:
```suggestion
const BoundaryContext = createContext<HTMLDivElement | null>(null)
```
since we no longer have to worry abt updateTrigger

just-toby: gotcha

cartcrom (approved):

PR #505, opened 2023-02-23
fix: modal animations, new fade animation

just-toby:
our feedback is that these modals should behave [exactly like the other modals in the app](https://www.notion.so/uniswaplabs/The-modal-animation-when-you-close-it-feels-jumpy-sudden-6eea1bfb09964247ac958a74f2b93b9b?pvs=4)

this means removing the y-translate and scaling, and just keeping an opacity transition. also, this fixes the animation to apply to the background scrim too, so it feels less jumpy


src/components/Dialog.tsx
```diff
@@ -243,17 +249,12 @@ const getAnimation = (animationType?: DialogAnimationType) => {
...
-`
-
-const FullScreenWrapper = styled.div<{ enabled?: boolean }>`
-  ${({ enabled }) =>
+const FullScreenWrapper = styled.div<{ enabled?: boolean; animationType?: DialogAnimationType }>`
+  ${({ enabled, animationType }) =>
     enabled &&
     css`
       align-items: center;
+      ${animationType === DialogAnimationType.FADE ? fadeAnimationCss : ''}
```
just-toby, line 257: `FullScreenWrapper` is the backdrop for full screen modals - this css is only applied if we enable the backdrop via the prop

src/components/Dialog.tsx
```diff
@@ -186,13 +186,19 @@ const slideOutRight = keyframes`
 
 const fadeIn = keyframes`
```
zzmp, line 187: These are now generic enough that I think `fadeIn`/`fadeOut`/`fadeAnimationCss` can all be moved to `theme/animations.ts`.

src/components/Dialog.tsx
zzmp, line 257: Are there other `DialogAnimationType`s, or should this be a `boolean`?

src/components/Dialog.tsx
```diff
@@ -339,7 +348,12 @@ export default function Dialog({ color, children, onClose, forceContain }: Dialo
       <ThemeProvider>
         <PopoverBoundaryProvider value={popoverRef.current} updateTrigger={updatePopover}>
           <div ref={popoverRef}>
-            <FullScreenWrapper enabled={pageCentered} onClick={closeOnBackgroundClick}>
+            <FullScreenWrapper
+              enabled={pageCentered}
+              onClick={closeOnBackgroundClick}
+              ref={fullScreenWrapperRef}
+              animationType={context.options?.animationType}
```
zzmp, line 355: Should this be a boolean - `fade={context.options?.animationType === DialogAnimationType.FADE}`?

src/components/Dialog.tsx
```diff
@@ -327,7 +336,7 @@ export default function Dialog({ color, children, onClose, forceContain }: Dialo
           return (mountPoint?.childElementCount ?? 0) > 1 ? SlideAnimationType.PAGING : SlideAnimationType.CLOSING
       }
     },
-    modal,
+    [fullScreenWrapperRef, modal],
```
zzmp, line 339: Would it be safer to use nesting CSS instead of passing an array here?

zzmp (approved):
lgtm - there are minor nits that could simplify the code, but nothing blocking, and the functionality looks good.

src/components/Dialog.tsx
just-toby: i think this is worth looking into later. i actually tried that approach at first but couldn't get it working. can't spend more time on this atm but i'll circle back.

PR #506, opened 2023-02-24
feat: add narrow variant of ActionButton

just-toby:
[see design here](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=12156%3A118039&t=b0V5aEN9cKtpFUZH-0)

adds a `narrow` prop to the ActionButton which makes the button shorter and have smaller text

### DIFF
--- a/src/components/ActionButton.tsx
+++ b/src/components/ActionButton.tsx
@@ -117,7 +117,7 @@ export default function ActionButton({
     }
   }, [color, disabled])
 
-  const buttonSize = useMemo(() => narrow ? 'small' : action ? 'medium' : 'large', [narrow, action])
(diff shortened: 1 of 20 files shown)


cartcrom (approved):

src/components/ActionButton.tsx
```diff
@@ -115,17 +117,35 @@ export default function ActionButton({
...
+  const buttonSize = useMemo(() => {
+    if (narrow) {
+      return 'small'
+    }
+    if (action) {
+      return 'medium'
+    }
+    return 'large'
+  }, [narrow, action])
+
```
cartcrom, line 129: nit
```suggestion
  const buttonSize = useMemo(() => narrow ? 'small' : action ? 'medium' : 'large', [narrow, action])

```

PR #507, opened 2023-02-24
feat: add close button to error dialog

just-toby:
see design here: https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=12156%3A118039&t=b0V5aEN9cKtpFUZH-0

adds an "x" icon to close the error dialog

### DIFF
--- a/src/components/Error/ErrorDialog.tsx
+++ b/src/components/Error/ErrorDialog.tsx
@@ -2,7 +2,9 @@ import { Trans } from '@lingui/macro'
 import ActionButton from 'components/ActionButton'
 import Column from 'components/Column'
 import Expando from 'components/Expando'
-import { AlertTriangle, Icon, LargeIcon } from 'icons'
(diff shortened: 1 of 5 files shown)


lynnshaoyu (approved):

src/components/Error/ErrorDialog.tsx
```diff
@@ -51,6 +53,10 @@ const ExpandoContent = styled(ThemedText.Code)`
   margin: 0.5em 0;
 `
 
+const StyledXButton = styled(X)`
```
lynnshaoyu, line 56: could you possibly reuse the StyledXButton component / export it instead of declaring twice in 2 separate files? although not sure what widget codebase convention is, so this is more of a suggestion / nit

just-toby: thanks, i'll dedup these

PR #508, opened 2023-02-24
feat: close dialog if user rejects token allowance

vm:


just-toby (changes requested):
the idea lgtm,  but RC for the lint errors and maybe using a custom `WidgetError`

just-toby (approved):

PR #510, opened 2023-02-24
feat: styled expando wrapper and update callsites

just-toby:
see the new expando design [here](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=12156%3A118039&t=be4Y6vrVFLpyQQf8-0)

this refactors the Expando component styling - instead of just offering a styling wrapper around the `title`, we wrap the entire title+content in a styled container.

for callsites that don't want this, (i.e. settings expandos), they must opt out and style their title manually with the legacy design


src/components/Expando.tsx
```diff
@@ -5,12 +5,22 @@ import Rule from 'components/Rule'
...
+`
+
+const StyledWrapper = styled(Column)<{ expanded: boolean }>`
+  background-color: ${({ theme }) => theme.module};
+  border-radius: ${({ theme }) => theme.borderRadius.medium}rem;
+  overflow: hidden;
+
+  @supports (overflow: clip) {
+    overflow: clip;
+  }
```
JFrankfurt, line 23: I had no idea we could do this

src/components/Swap/Settings/MaxSlippageSelect.tsx
```diff
@@ -142,10 +143,8 @@ export default function MaxSlippageSelect() {
   return (
     <Column gap={0.75}>
       <Expando
-        hideRulers
-        showBottomGradient={false}
         title={
-          <Row grow>
+          <Row style={{ cursor: 'pointer' }} grow justify="space-between" onClick={() => setOpen(!open)}>
```
JFrankfurt, line 147: nit: useCallback on open toggle

src/components/Swap/Settings/TransactionTtlInput.tsx
```diff
@@ -44,20 +46,26 @@ export default function TransactionTtlInput() {
...
+        styledWrapper={false}
         title={
-          <Label
-            name={<Trans>Transaction deadline</Trans>}
-            // TODO (tina): clicking on this tooltip on mobile shouldn't open/close expando
-            tooltip={
-              <Trans>Your transaction will revert if it has been pending for longer than this period of time.</Trans>
-            }
-          />
+          <Row style={{ cursor: 'pointer' }} onClick={() => setOpen(!open)}>
```
JFrankfurt, line 54: I thiiiiink the title already handles click-to-close? Maybe we're losing that w/ your refactor and this is adding it back?


Also, useCallback for toggle would be nice here also since you're using the anon fn implementation twice

JFrankfurt (approved):
some nits and minor suggestions but lgtm

src/components/Swap/Settings/TransactionTtlInput.tsx
just-toby: >  losing that w/ your refactor

that's right, this component is replacing the default title bar

PR #511, opened 2023-02-24
feat: new error dialog design

just-toby:
[design spec here](https://www.figma.com/file/kNSDBMpOzxSTOP6MerohLm/Web-Design-Spec?node-id=10575%3A122757)

implements a new ErrorDialog design, building on the changes linked below.

note that I'm renaming `ErrorDialog` to `ErrorView` because it's not a dialog anymore. it will just be a regular inline-rendered component.

depends on #507 , #508 , #509 , and #510 .


src/components/Error/ErrorView.tsx
```diff
@@ -0,0 +1,91 @@
...
+      </Column>
+    </>
+  )
+}
+
+const ExpandoContent = styled(ThemedText.Code)`
+  margin: 0.5em;
+`
+
+const StyledXButton = styled(X)`
```
lynnshaoyu, line 41: I feel like i've seen this component declared a couple times across the widget repo now, could we extract it into a single file and re-use via exporting?

src/components/Error/ErrorBoundary.tsx
```diff
@@ -53,30 +53,30 @@ export default class ErrorBoundary extends Component<PropsWithChildren<ErrorBoun
...
-        />
-      </Dialog>
+      <ErrorView
+        message={header}
+        error={error}
+        action={t`Get support`}
+        onDismiss={
+          error instanceof WidgetError && error.dismissable
+            ? () => {
+                this.setState({ error: undefined })
```
lynnshaoyu, line 66: is there a reason behind why we're using a mix of class and functional react components in widget repo?

lynnshaoyu (approved):
non-blocking comments

src/components/Error/ErrorBoundary.tsx
just-toby: yes - the ErrorBoundary needs to be a class component. this should be the only one afaik.

> Only class components can be error boundaries. In practice, most of the time you’ll want to declare an error boundary component once and use it throughout your application.

https://reactjs.org/docs/error-boundaries.html#introducing-error-boundaries

PR #513, opened 2023-02-27
fix: don't propogate tooltip clicks to their parents

JFrankfurt:
There are some more issues w/ this tooltip/popper setup that I'd like to fix eventually, but this change makes the tooltip usable with our `Expando` component at least


just-toby (approved):
thanks for fixing! btw, i have another tooltip PR open which may address some of the other issues you're seeing: https://github.com/Uniswap/widgets/pull/500

PR #516, opened 2023-02-28
fix: loading state post-error

just-toby:
make sure we show a loading state while fetching the next quote, even if the most recent cached result is a "not route found" error !

### DIFF
--- a/src/hooks/routing/useRouterTrade.ts
+++ b/src/hooks/routing/useRouterTrade.ts
@@ -59,8 +59,8 @@ export function useRouterTrade(
     }
   }, [amountSpecified, quoteConfig])
 
-  // Get the cached state *immediately* to update the UI without sending a request - using useGetQuoteQueryState -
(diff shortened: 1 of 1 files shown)


src/hooks/routing/useRouterTrade.ts
```diff
@@ -82,7 +82,7 @@ export function useRouterTrade(
   return useMemo(() => {
     if (!amountSpecified || isError || queryArgs === skipToken) {
       return TRADE_INVALID
-    } else if (data === NO_ROUTE) {
+    } else if (data === NO_ROUTE && isValid) {
```
just-toby, line 85: `isValid` represents whether the cached result (`data`) is the actual result for the current `queryArgs` if it's not, then we're fetching a new result for the most relevant args.

tinaszheng: helpful comment, ty!

tinaszheng (approved):

PR #519, opened 2023-03-01
fix: token selector design issues

just-toby:
- move X close button to the right side of the header, and make it heavier
- remove side padding from token list section
- adjust font weights, and spacings in the TokenOptions and CommonBases
- add ETH to the common bases for mainnet
- make token selector a bottom sheet on mobile widths

feedbacks here: https://www.notion.so/uniswaplabs/Token-selector-issues-536b091ac80d474c884602e541287e09?pvs=4

### DIFF
--- a/src/components/TokenSelect/index.tsx
+++ b/src/components/TokenSelect/index.tsx
@@ -1,9 +1,11 @@
 import { t, Trans } from '@lingui/macro'
 import { Currency } from '@uniswap/sdk-core'
 import { useWeb3React } from '@web3-react/core'
+import { BottomSheetModal } from 'components/BottomSheetModal'
 import { inputCss, StringInput } from 'components/Input'
 import { useConditionalHandler } from 'hooks/useConditionalHandler'
 import { useCurrencyBalances } from 'hooks/useCurrencyBalance'
+import { useIsMobileWidth } from 'hooks/useIsMobileWidth'
 import useNativeCurrency from 'hooks/useNativeCurrency'
 import useTokenList, { useIsTokenListLoaded, useQueryTokens } from 'hooks/useTokenList'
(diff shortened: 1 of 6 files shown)


src/components/TokenSelect/index.tsx
```diff
@@ -149,10 +162,23 @@ export default memo(function TokenSelect({ field, value, approved, disabled, onS
...
     [onSelect, setOpen]
   )
+  const isMobile = useIsMobileWidth()
+  const pageCenteredDialogsEnabled = useIsDialogPageCentered()
+
   return (
     <>
       <TokenButton value={value} approved={approved} disabled={disabled} onClick={onOpen} />
-      {open && <TokenSelectDialog value={value} onSelect={selectAndClose} onClose={() => setOpen(false)} />}
+      {isMobile && pageCenteredDialogsEnabled ? (
```
vm, line 171: nit: this might be better with actual if statements?

src/components/TokenSelect/index.tsx
```diff
@@ -27,6 +29,15 @@ const SearchInputContainer = styled(Row)`
   ${inputCss}
 `
 
+const TokenSelectContainer = styled.div`
+  border-radius: ${({ theme }) => theme.borderRadius.medium}em;
+  overflow: hidden;
+  padding: 0.5em 0 0;
+  @supports (overflow: clip) {
+    overflow: 'clip';
```
vm, line 37: is there a downside to just putting this without the @supports? just curious

vm (approved):

src/components/TokenSelect/index.tsx
just-toby: yes, it would override the "hidden" value

PR #520, opened 2023-03-01
feat: upgrade ethers, sdk-core, universal-router, smart-order-router

vm:
- Removes support for kovan, goerli and ropsten from sdk-core
- Namehash now works with emojis in ENS names, so don't need a safe wrapper
- Smart-order-router upgraded to support BNB chain in the future


zzmp (approved):

package.json
```diff
@@ -68,10 +68,10 @@
     "@uniswap/permit2-sdk": "^1.2.0",
     "@uniswap/redux-multicall": "^1.1.8",
     "@uniswap/router-sdk": "^1.3.0",
-    "@uniswap/sdk-core": "^3.0.1",
-    "@uniswap/smart-order-router": "^2.10.0",
+    "@uniswap/sdk-core": "^4.0.0",
+    "@uniswap/smart-order-router": "^3.6.0",
```
zzmp, line 72: This requires ethers@^5.7.2, so ethers should be updated as well.
This will likely require some of our users to upgrade their ethers.

zzmp (changes requested):

zzmp (approved):

PR #522, opened 2023-03-01
build: clean up unused i18n

zzmp:
Cleans up usage of i18n, including removing it from cosmos.
Translations are only generated as part of the release, so they are untestable with the current setup (ie without adding crowdin CLI commands).


package.json
```diff
@@ -28,10 +28,8 @@
     "contracts:compile:abi": "typechain --target ethers-v5 --out-dir src/abis/types \"./src/abis/**/*.json\"",
     "contracts:compile:v3": "typechain --target ethers-v5 --out-dir src/types/v3 \"./node_modules/@uniswap/**/artifacts/contracts/**/*.json\"",
     "contracts:compile": "yarn contracts:compile:abi && yarn contracts:compile:v3",
-    "prei18n:extract": "touch src/locales/en-US.po",
```
just-toby, line 31: are you sure we don't need this step anymore?

just-toby (changes requested):
rc to sanity check the missing step

package.json
zzmp: yes - this file is created by `lingui extract --locale en-US`

just-toby (approved):

PR #525, opened 2023-03-02
revert "fix: remove USD value impact"

just-toby:
Reverts Uniswap/widgets#457


src/components/Swap/Settings/MaxSlippageSelect.tsx
```diff
@@ -190,7 +193,7 @@ export default function MaxSlippageSelect() {
                 size={Math.max(maxSlippageInput.length, 4)}
                 value={maxSlippageInput}
                 onChange={(input) => processInput(input)}
-                placeholder={'0.10'}
+                placeholder={allowedSlippage.allowed.toFixed(2)}
```
just-toby, line 196: we should show the actual "auto" value we're using here, which can change based on the current trade

src/components/Swap/Output.tsx
```diff
@@ -43,7 +42,7 @@ export default function Output() {
       <OutputWrapper
         isWide={isWideWidget}
         field={Field.OUTPUT}
-        impact={impact}
+        impact={fiatValueChange}
```
just-toby, line 45: we should show the fiat value change here, because right next to the USD value of the output.

src/components/Swap/Summary/Estimate.tsx
```diff
@@ -49,10 +51,10 @@ export function getEstimateMessage(
       descriptor: (
         <ThemedText.Body2>
           <Trans>Minimum output after slippage</Trans>
-          {priceImpact && (
+          {slippage && (
             <ThemedText.Body2 $inline color={slippage?.warning ?? 'secondary'}>
               {' '}
-              ({priceImpact?.toString()})
+              ({slippage?.toString()})
```
just-toby, line 57: we should be showing the SLIPPAGE % here, next to the word "slippage"

src/components/Swap/Toolbar/index.tsx
```diff
@@ -115,6 +114,7 @@ function CaptionRow() {
 
   const tradeSummaryRows: SummaryRowProps[] = useMemo(() => {
     const currencySymbol = trade?.outputAmount?.currency.symbol ?? ''
+    const { descriptor, value } = getEstimateMessage(trade, slippage)
```
just-toby, line 117: Toolbar should be re-using this util that we built for the summary dialog

src/components/Swap/index.tsx
```diff
@@ -48,21 +48,21 @@ export default function Swap(props: SwapProps) {
 
   return (
     <>
-      <Header title={<Trans>Swap</Trans>}>
```
just-toby, line 51: this change is just moving `SwapInfoProvider` higher up, to wrap the settings component too. we need this so that the settings view has access to the current slippage

src/hooks/swap/useSwapInfo.tsx
```diff
@@ -105,6 +106,7 @@ function useComputeSwapInfo(): SwapInfo {
   // Wait until the trade is valid to avoid displaying incorrect intermediate values.
   const slippage = useSlippage(trade)
   const impact = usePriceImpact(trade.trade)
+  const fiatValueChange = useFiatPriceImpact(trade.trade)
```
just-toby, line 109: this is a new metric that shouldn't be confused with the 2 above

src/hooks/useAllCurrencyCombinations.ts
```diff
@@ -1,71 +0,0 @@
-import { Currency, Token } from '@uniswap/sdk-core'
```
just-toby, line 1: this file is not used and was giving me typecheck errors

src/utils/computeFiatValuePriceImpact.tsx
```diff
@@ -0,0 +1,14 @@
+import { Currency, CurrencyAmount, Percent } from '@uniswap/sdk-core'
+import { ONE_HUNDRED_PERCENT } from 'constants/misc'
+import JSBI from 'jsbi'
+
+export function computeFiatValuePriceImpact(
```
just-toby, line 5: this is a function i removed in the previous PR, i'm just adding it back here

src/hooks/useSlippage.ts
```diff
@@ -16,9 +18,14 @@ export interface Slippage {
   auto: boolean
   allowed: Percent
   warning?: 'warning' | 'error'
+  toString: () => string
```
just-toby, line 21: we're displaying this value in more places, so we should use the same pattern as price impact - a built in string formatter.

src/hooks/usePriceImpact.ts
```diff
@@ -22,6 +25,21 @@ export function usePriceImpact(trade?: InterfaceTrade): PriceImpact | undefined
   }, [trade])
 }
 
+export function useFiatPriceImpact(trade?: InterfaceTrade) {
```
just-toby, line 28: this logic was removed in the previous PR, i'm adding it back here

src/components/Swap/Output.tsx
tinaszheng: as a followup we should probably just rename the prop here

src/hooks/swap/useSwapInfo.tsx
tinaszheng: i'd rename this to just `useFiatValueChange`

src/hooks/useSlippage.ts
tinaszheng: you should follow the other usages of formatter and define a formatting function that takes in Slippage instead of defining an object method 

if you add a function as a method here, slippage no longer becomes a serializable object, which may pose problems if you want to store them in any modern redux-based library

tinaszheng (changes requested):

just-toby:
addressed comments

tinaszheng (approved):

PR #528, opened 2023-03-02
fix: stop blocking swap on balance fetching

just-toby:
https://uniswaplabs.atlassian.net/browse/WEB-2766?atlOrigin=eyJpIjoiNTA4OGVmZWI5ZTJhNDMyMThhMjkwOGRkZjc0NjBhMmUiLCJwIjoiaiJ9

we want to stop blocking swap if the balances aren't loaded to match the current behavior in interface.

if the balance IS loaded, we should still disable the button if the user doesn't have enough funds

### DIFF
--- a/src/components/Swap/SwapActionButton/index.tsx
+++ b/src/components/Swap/SwapActionButton/index.tsx
@@ -28,8 +28,8 @@ export default function SwapActionButton() {
       (!permit2Enabled && approval.state !== SwapApprovalState.APPROVED) ||
       error !== undefined ||
       (!isWrap && !trade) ||
-      !(inputCurrencyAmount && inputCurrencyBalance) ||
-      inputCurrencyBalance.lessThan(inputCurrencyAmount),
+      !inputCurrencyAmount ||
+      (Boolean(inputCurrencyBalance) && Boolean(inputCurrencyBalance?.lessThan(inputCurrencyAmount))),
     [permit2Enabled, approval.state, error, isWrap, trade, inputCurrencyAmount, inputCurrencyBalance]
   )
 
--- a/src/components/Swap/SwapActionButton/index.tsx
+++ b/src/components/Swap/SwapActionButton/index.tsx
@@ -29,7 +29,8 @@ export default function SwapActionButton() {
       error !== undefined ||
       (!isWrap && !trade) ||
       !inputCurrencyAmount ||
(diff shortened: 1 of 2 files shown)


src/components/Swap/SwapActionButton/index.tsx
```diff
@@ -28,8 +28,8 @@ export default function SwapActionButton() {
       (!permit2Enabled && approval.state !== SwapApprovalState.APPROVED) ||
       error !== undefined ||
       (!isWrap && !trade) ||
-      !(inputCurrencyAmount && inputCurrencyBalance) ||
-      inputCurrencyBalance.lessThan(inputCurrencyAmount),
+      !inputCurrencyAmount ||
```
just-toby, line 31: i'm not sure if we still need this check - if we're requiring `inputCurrencyAmount`, shouldn't we also be requiring `outputCurrencyAmount`?

zzmp: We're not checking it against a balance, so we don't need to check it at all: `!(inputCurrencyAmount && inputCurrencyBalance)` acted as a guard to ensure that `inputCurrencyBalance.lessThan(inputCurrencyAmount)` was valid. I _think_ this should be covered by `(!isWrap && !trade)`.

src/components/Swap/SwapActionButton/index.tsx
```diff
@@ -28,8 +28,8 @@ export default function SwapActionButton() {
...
+      (Boolean(inputCurrencyBalance) && Boolean(inputCurrencyBalance?.lessThan(inputCurrencyAmount))),
```
zzmp, line 32: You either need the guard check (`inputCurrencyBalance &&`) _or_ the optional operator (`?.`) - I think you can simplify it like this:

```suggestion
      inputCurrencyBalance?.lessThan(inputCurrencyAmount),
```

zzmp (changes requested):

src/components/Swap/SwapActionButton/index.tsx
just-toby: this is true, but we avoid some type warnings if we keep it.

src/components/Swap/SwapActionButton/index.tsx
just-toby: still need to cast to Boolean so that `isDisabled` isn't undefined

zzmp (approved):

PR #529, opened 2023-03-02
fix: collapse toolbar for wrap transactions

just-toby:
collapse the toolbar when we switch to a wraping swap

# before

https://user-images.githubusercontent.com/66155195/223544975-b35b4a95-263b-4a58-8569-24064dfd20c5.mov

# after

https://user-images.githubusercontent.com/66155195/223544801-105fac31-89ef-4e7d-916d-32bfda5227c6.mov

### DIFF
--- a/src/components/Swap/Toolbar/ToolbarContext.tsx
+++ b/src/components/Swap/Toolbar/ToolbarContext.tsx
@@ -1,4 +1,5 @@
-import { createContext, PropsWithChildren, useContext, useState } from 'react'
+import { useIsWrap } from 'hooks/swap/useWrapCallback'
+import { createContext, PropsWithChildren, useContext, useEffect, useState } from 'react'
 
 export const Context = createContext<{
   open: boolean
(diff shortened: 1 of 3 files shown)


src/components/Swap/Toolbar/ToolbarContext.tsx
```diff
@@ -14,6 +15,14 @@ export const Provider = ({ children }: PropsWithChildren) => {
   const [open, setOpen] = useState(false)
   const onToggleOpen = () => setOpen((open) => !open)
   const collapse = () => setOpen(false)
+
+  const isWrap = useIsWrap()
```
vm, line 19: will it re-open if it's not a wrap?

vm (approved):

src/components/Swap/Toolbar/ToolbarContext.tsx
just-toby: no, and we don't want it to

PR #531, opened 2023-03-03
fix: contained tokenlist height

just-toby:
fixes a bug where the token list was being cut off in the non-page-centered version of the widget.

### DIFF
--- a/src/components/TokenSelect/index.tsx
+++ b/src/components/TokenSelect/index.tsx
@@ -8,6 +8,7 @@ import { useCurrencyBalances } from 'hooks/useCurrencyBalance'
 import { useIsMobileWidth } from 'hooks/useIsMobileWidth'
 import useNativeCurrency from 'hooks/useNativeCurrency'
 import useTokenList, { useIsTokenListLoaded, useQueryTokens } from 'hooks/useTokenList'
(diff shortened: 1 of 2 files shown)


src/components/TokenSelect/index.tsx
```diff
@@ -98,7 +102,12 @@ export function TokenSelectDialogContent({ value, onSelect, onClose }: TokenSele
     )
   }
   return (
-    <TokenSelectContainer>
+    <TokenSelectContainer
+      style={{
+        minWidth: isPageCentered ? Math.min(400, width) : 'auto',
+        minHeight: isPageCentered ? 'unset' : '100%',
+      }}
```
JFrankfurt, line 109: can you do this with `$isPageCentered` as a transient styled-component prop and something like `min(400px, 100vw)`?

JFrankfurt (changes requested):
see comment

src/components/TokenSelect/index.tsx
just-toby: yep

just-toby:
https://user-images.githubusercontent.com/66155195/222813938-9665811f-3722-4120-a2c2-68dd86833c37.mov

JFrankfurt (approved):

PR #533, opened 2023-03-03
feat: use rem instead of em

just-toby:
# Background

### what's the difference?
- px stands for pixels and is an absolute unit of measurement. It is fixed and does not change with respect to the size of the parent element or the viewport.
- em is a relative unit of measurement that is based on the font-size of the parent element. If the font-size of the parent element is 16px, then 1em would be equal to 16px. If the font-size of the parent element is changed, then the size of the element will also change proportionally.
- rem is also a relative unit of measurement but is based on the font-size of the root element. The root element is typically the <html> tag, and the default font-size is usually 16px. If the font-size of the root element is changed, then the size of the element will also change proportionally.

### what's better?

We should be using a unit that scales with font size (rem or em). However, pixels are required for certain use cases: 
- Border widths: Since border widths are often very small, using "px" units can provide more precise control over their size than using "rem" units.
- Background images: When using background images, "px" units can be more appropriate because the size of the image is not affected by the font-size of the element.
- Fixed layouts: In some cases, you may want to use a fixed layout where all elements are sized using "px" units. This can be useful for designs where exact pixel values are important and the layout is not intended to be responsive.
- Third-party components: If you are using third-party components or libraries that are designed to use "px" units, it may be easier to stick with "px" units for consistency and to avoid conflicts with the component's design.

We should be using REM for most other cases, because...
- Responsive design: Rem units can be used to create responsive designs that adapt to different screen sizes and devices. By setting the font-size of the root element to a percentage, the entire layout can be scaled up or down as needed.
- Consistency: Using "rem" units ensures that all elements on a webpage are proportionally sized based on the same font-size, which creates a consistent layout and improves the overall design of the webpage. With "em" units, the size of an element is relative to the font-size of its parent element, which can create inconsistencies if the font-size of the parent element changes.
  - This can cause hard-to-find bugs in which some ancestor in the hierarchy, at any depth, can mess up your measurement. 
- Ease of maintenance: Since "rem" is based on the root font-size, changing the font-size of the entire page can be done by modifying just one value. This makes it much easier to maintain and modify the codebase in the future. With "em" units, changes to the font-size of a parent element can require updates to the font-size of all its child elements.
- Accessibility: "Rem" is a more accessible unit of measurement because it can be scaled up or down globally, making it easy to improve readability for users who may have visual impairments or disabilities. With "em" units, the size of an element is relative to the font-size of its parent element, which can make it more difficult to adjust the font-size of an entire webpage for accessibility purposes.
- Responsive design: "Rem" units can be used to create responsive designs that adapt to different screen sizes and devices. By setting the font-size of the root element to a percentage, the entire layout can be scaled up or down as needed. "Em" units can also be used for responsive design, but can be more difficult to work with because the size of an element is relative to the font-size of its parent element.

# This PR

is basically just a find-and-replace. the only interesting thing to check would be our `icons/index.tsx` file - I actually left these using EM, because I think they should scale with the font size of their container. I'm open to feedback here.

https://uniswaplabs.atlassian.net/browse/WEB-2740?atlOrigin=eyJpIjoiODYxNjcyNWMyNjdkNGY4OWIyYTk4NGUwODg1ODg3OTUiLCJwIjoiaiJ9

### DIFF
--- a/src/components/ActionButton.tsx
+++ b/src/components/ActionButton.tsx
@@ -8,9 +8,9 @@ import Row, { RowProps } from './Row'
 import Tooltip from './Tooltip'
 
 const StyledButton = styled(Button)<{ shouldUseDisabledColor?: boolean; narrow?: boolean }>`
-  border-radius: ${({ theme, narrow }) => (narrow ? theme.borderRadius.small : theme.borderRadius.medium)}em;
+  border-radius: ${({ theme, narrow }) => (narrow ? theme.borderRadius.small : theme.borderRadius.medium)}rem;
   flex-grow: 1;
-  max-height: ${({ narrow }) => (narrow ? '2.5em' : '3.5em')};
+  max-height: ${({ narrow }) => (narrow ? '2.5rem' : '3.5rem')};
   transition: background-color ${AnimationSpeed.Medium} ease-out, border-radius ${AnimationSpeed.Medium} ease-out,
     flex-grow ${AnimationSpeed.Medium} ease-out;
   ${({ theme, disabled, shouldUseDisabledColor }) =>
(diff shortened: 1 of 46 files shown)


zzmp (approved):
This PR description is basically MDN documentation. I think you should try to PR this upstream:
https://github.com/mdn/content/blob/main/files/en-us/learn/css/styling_text/fundamentals/index.md#font-size

An emphatic lgtm

tinaszheng:
border-radius seems like it should also be px?

PR #534, opened 2023-03-04
fix: deep link by swapping sooner after user interaction

zzmp:
Uses `@uniswap/conedison/provider` implementation of `sendTransaction`, which includes a gas margin computation.

It also omits the gas margin computation for Uniswap Wallet, and uses sendUncheckedTransaction under the hood. This allows the app to be linked to within 300ms, which allows iOS to deep-link through WalletConnect to the mobile app instead of navigating the to app's webpage.

To test (this may not work on cloudflare/vercel deploys - I'm not sure why, but it may be because they add instrumentation):
- Using your iOS device, connect to the App using Uniswap Wallet via WalletConnect
- Initiate a swap/wrap/etc on the app
- The application should deep-link directly to the wallet

Requires https://github.com/Uniswap/conedison/pull/13 to poll before merging.

### DIFF
--- a/src/hooks/useUniversalRouter.ts
+++ b/src/hooks/useUniversalRouter.ts
@@ -7,6 +7,7 @@ import { SwapRouter, UNIVERSAL_ROUTER_ADDRESS } from '@uniswap/universal-router-
 import { FeeOptions, toHex } from '@uniswap/v3-sdk'
 import { useWeb3React } from '@web3-react/core'
 import { ErrorCode } from 'constants/eip1193'
+import { TX_GAS_MARGIN } from 'constants/misc'
 import { SwapError } from 'errors'
 import { useCallback } from 'react'
 import { InterfaceTrade } from 'state/routing/types'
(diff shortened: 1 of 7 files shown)


src/hooks/useUniversalRouter.ts
```diff
@@ -68,15 +68,7 @@ export function useUniversalRouterSwapCallback(trade: InterfaceTrade | undefined
...
-      let gasEstimate: BigNumber
-      try {
-        gasEstimate = await provider.estimateGas(tx)
-      } catch (gasError) {
-        console.warn(gasError)
-        throw new SwapError({ header: t`Swap Error`, message: t`Your swap is expected to fail` })
-      }
-      const gasLimit = calculateGasMargin(gasEstimate)
-      response = await provider.getSigner().sendTransaction({ ...tx, gasLimit })
+      response = await sendTransaction(provider, tx, 0.2)
```
tinaszheng, line 71: same here about moving the 0.2 to a constant

tinaszheng (approved):

PR #536, opened 2023-03-06
fix: show connect to ethereum button for unsupported chains

just-toby:
show "Connect to Ethereum" button when on an unsupported chain

### DIFF
--- a/src/components/Swap/SwapActionButton/index.tsx
+++ b/src/components/Swap/SwapActionButton/index.tsx
@@ -36,9 +36,7 @@ export default function SwapActionButton() {
 
   if (!account || !isActive) {
     return <ConnectWalletButton />
-  } else if (error === ChainError.UNSUPPORTED_CHAIN) {
-    return <SwitchChainButton chainId={SupportedChainId.MAINNET} />
-  } else if (error === ChainError.MISMATCHED_CHAINS) {
+  } else if (error === ChainError.MISMATCHED_CHAINS || error === ChainError.UNSUPPORTED_CHAIN) {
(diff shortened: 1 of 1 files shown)


src/components/Swap/SwapActionButton/index.tsx
```diff
@@ -36,6 +36,8 @@ export default function SwapActionButton() {
 
   if (!account || !isActive) {
     return <ConnectWalletButton />
+  } else if (error === ChainError.UNSUPPORTED_CHAIN) {
+    return <SwitchChainButton chainId={SupportedChainId.MAINNET} />
```
JFrankfurt, line 40: I think this is an improvement, but I think the expected behavior is that the target chain should match the tokens passed into the widget. I.e., if the user is connected to (unsupported network) and the token input/outputs are on Polygon, then the widget should say "switch to polygon"

JFrankfurt (changes requested):

src/components/Swap/SwapActionButton/index.tsx
```diff
@@ -36,7 +36,7 @@ export default function SwapActionButton() {
 
   if (!account || !isActive) {
     return <ConnectWalletButton />
-  } else if (error === ChainError.MISMATCHED_CHAINS) {
+  } else if (error === ChainError.MISMATCHED_CHAINS || error === ChainError.UNSUPPORTED_CHAIN) {
     const tokenChainId = inputCurrency?.chainId ?? outputCurrency?.chainId ?? SupportedChainId.MAINNET
```
JFrankfurt, line 40: Unless it's happening elsewhere, I think you'll probably want to confirm that the input/output currency chainId is supported instead of trusting it to be supported if it is defined.

JFrankfurt (approved):
one comment that might req change or might not

PR #542, opened 2023-03-07
fix: price impact tostring

just-toby:
remove the custom `toString` field on Price Impact objects because it makes them unserializable. prefer to use a helper function.

### DIFF
--- a/src/hooks/usePriceImpact.ts
+++ b/src/hooks/usePriceImpact.ts
@@ -33,7 +33,6 @@ export function useFiatValueChange(trade?: InterfaceTrade) {
     return {
       percent: fiatPriceImpact,
       warning: getPriceImpactWarning(fiatPriceImpact),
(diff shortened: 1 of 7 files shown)


src/hooks/usePriceImpact.ts
```diff
@@ -49,3 +46,8 @@ export function toHumanReadablePercent(priceImpact: Percent): string {
   const number = parseFloat(priceImpact.multiply(-1)?.toFixed(2))
   return `${sign}${number}%`
 }
+
+export function formatPriceImpact(impact: PriceImpact | undefined): string {
```
tinaszheng, line 50: maybe copy the mobile version to be consistent? or we could move to conedison: https://github.com/Uniswap/mobile/blob/c4f3ab31b2b9422230e0e308c8b9db1ddf9125c6/src/utils/format.ts#L295

just-toby: updated to use conedison

PR #548, opened 2023-03-10
fix: stricter typing for router query

zzmp:
Narrows typings for router queries, which avoids having to coerce types in useRouterTrade.
This will enable cleaner code for a planned perf event hook (see #549).

Also uncovers and fixes a bug for swaps with no quote (`QuoteState.NOT_FOUND`), which do not include a `blockNumber` but should still be rendered.

### DIFF
--- a/src/hooks/routing/clientSideSmartOrderRouter.ts
+++ b/src/hooks/routing/clientSideSmartOrderRouter.ts
@@ -13,7 +13,7 @@ import {
 } from '@uniswap/smart-order-router'
 import { nativeOnChain } from 'constants/tokens'
 import JSBI from 'jsbi'
-import { GetQuoteArgs, GetQuoteError, INITIALIZED, NO_ROUTE, QuoteResult } from 'state/routing/types'
+import { GetQuoteArgs, QuoteResult, QuoteState } from 'state/routing/types'
 import { isExactInput } from 'utils/tradeType'
 
 import { transformSwapRouteToGetQuoteResult } from './transformSwapRouteToGetQuoteResult'
@@ -82,7 +82,7 @@ function getRouter(chainId: ChainId, provider: BaseProvider): AlphaRouter {
   return router
 }
 
-async function getQuote(
+async function getQuoteResult(
   {
     tradeType,
     tokenIn,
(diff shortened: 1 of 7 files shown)


src/state/routing/types.ts
```diff
@@ -84,17 +90,28 @@ export interface QuoteResult {
...
+    }
+  | {
+      state: QuoteState.SUCCESS
+      data: QuoteData
+    }
 
-export type TradeQuoteResult = TradeResult | GetQuoteError
+export type TradeResult =
+  | {
+      state: QuoteState.INITIALIZED | QuoteState.NOT_FOUND
```
tinaszheng, line 105: it looks like `state: QuoteState.NOT_FOUND` is only coupled with the rest of the varaibles being undefined, so i would narrow down the type definition of `TradeResult` even further:

```
 {
   state: QuoteState.INITIALIZED | QuoteState.NOT_FOUND
   trade: undefined 
   gasUseEstimateUSD: undefined 
   blockNumber: undefined
  
```

src/hooks/routing/clientSideSmartOrderRouter.ts
```diff
@@ -111,13 +111,13 @@ async function getQuote(
   const amount = CurrencyAmount.fromRawAmount(baseCurrency, JSBI.BigInt(amountRaw ?? '1')) // a null amountRaw should initialize the route
   const route = await router.route(amount, quoteCurrency, tradeType, /*swapConfig=*/ undefined, routerConfig)
 
-  if (!amountRaw) return INITIALIZED
-  if (!route) return NO_ROUTE
+  if (!amountRaw) return { state: QuoteState.INITIALIZED }
```
tinaszheng, line 114: when does this case happen? shouldn't this be part of the skipToken case?

tinaszheng (approved):

src/hooks/routing/clientSideSmartOrderRouter.ts
zzmp: the slow part of CSSOR is initialization - that first time a token pair is checked. but it's not even getting the quote, it's initializing the data to get the quote.
so as soon as we have a pair, we make a query with a null amount, signifying that it's just an initialization. the client-side code will replace null with 1 so that it fetches the necessary initializing data, and then return QuoteState.INITIALIZED.
then, when we make a real query, it's nearly instant!

tinaszheng: 🤦‍♀️ oy VEY!

src/state/routing/types.ts
zzmp: If I type them as `undefined` then I'm required to define them (as `undefined` 🤮 ).
So I'm going to narrow them to optional `undefined`, so I can continue to actually leave them undefined.

tinaszheng: That's fine!

PR #549, opened 2023-03-10
feat: onSwapQuote handler

zzmp:
Adds a handler for swap quote timings, to be used by consumers for performance metrics.

### DIFF
--- a/src/state/routing/args.ts
+++ b/src/state/routing/args.ts
@@ -6,17 +6,15 @@ import { QuoteConfig, QuoteType } from 'hooks/routing/types'
 import useIsWindowVisible from 'hooks/useIsWindowVisible'
 import { useAtomValue } from 'jotai/utils'
 import { useMemo } from 'react'
-import { swapRouterUrlAtom } from 'state/swap'
+import { swapEventHandlersAtom, swapRouterUrlAtom } from 'state/swap'
 
 import { GetQuoteArgs } from './types'
 import { currencyAddressForSwapQuote } from './utils'
 
-const NON_SERIALIZABLE_KEYS = ['provider']
+const NON_SERIALIZABLE_KEYS = ['provider', 'onQuote']
 
-/** Omits the non-serializable keys from GetQuoteArgs' cache key. */
-export function serializeGetQuoteArgs({ endpointName, queryArgs }: { endpointName: string; queryArgs: GetQuoteArgs }) {
-  // same as default serializeQueryArgs, but ignoring non-serializable keys.
-  return `${endpointName}(${JSON.stringify(queryArgs, (key, value) => {
+export function serializeGetQuoteArgs(args: GetQuoteArgs) {
+  return JSON.stringify(args, (key, value) => {
     if (NON_SERIALIZABLE_KEYS.includes(key)) {
       return undefined
     }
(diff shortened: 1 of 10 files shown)


src/state/routing/slice.ts
```diff
@@ -23,9 +23,32 @@ const baseQuery: BaseQueryFn<GetQuoteArgs, TradeResult> = () => {
...
+        } finally {
+          const duration = performance.now() - timestamp
+          args.onQuote?.(
+            {
+              timestamp,
+              duration,
+              args: JSON.parse(serializeGetQuoteArgs(args)),
+              state: result?.state,
+            },
+            error
```
zzmp, line 48: This can be narrowed to `FetchBaseQueryError`

src/state/routing/args.ts
```diff
@@ -6,17 +6,15 @@ import { QuoteConfig, QuoteType } from 'hooks/routing/types'
...
 import { currencyAddressForSwapQuote } from './utils'
 
-const NON_SERIALIZABLE_KEYS = ['provider']
+const NON_SERIALIZABLE_KEYS = ['provider', 'onQuote']
 
-/** Omits the non-serializable keys from GetQuoteArgs' cache key. */
-export function serializeGetQuoteArgs({ endpointName, queryArgs }: { endpointName: string; queryArgs: GetQuoteArgs }) {
-  // same as default serializeQueryArgs, but ignoring non-serializable keys.
-  return `${endpointName}(${JSON.stringify(queryArgs, (key, value) => {
+export function serializeGetQuoteArgs(args: GetQuoteArgs) {
```
just-toby, line 16: is this function specific to `onQuote`/`GetQuote` or can we just rename it to be more generic (might help disambiguate between it and the new function below)

src/state/swap/index.ts
```diff
@@ -85,6 +85,17 @@ interface InputEventHandlers {
 /** An integration hook called when the user approves a token, either through allowance or permit. */
 export type OnSwapApprove = () => void
 
+/** An integration hook called when a new quote is fetched. */
+export type OnSwapQuote = (
+  event: {
+    timestamp: number
```
just-toby, line 91: is this the timestamp of the query start or end? (unclear to integrator)

src/state/swap/index.ts
```diff
@@ -85,6 +85,17 @@ interface InputEventHandlers {
 /** An integration hook called when the user approves a token, either through allowance or permit. */
 export type OnSwapApprove = () => void
 
+/** An integration hook called when a new quote is fetched. */
+export type OnSwapQuote = (
```
just-toby, line 89: do you think there's any benefit to making `event` fields optional? i'm thinking of situations where we need to change the API (adding/removing fields without disrupting integrators). maybe this only applies to when we add or remove later?

just-toby (approved):
lgtm

src/state/routing/args.ts
zzmp: These are specific to `GetQuote`:
- `serializeGetQuoteArgs` serializes a `GetQuoteArgs` object
- `serializeGetQuoteQueryArgs` is plugged into `serializeQueryArgs` for the `GetQuote` query

I think they're both named appropriately, just unfortunately very similarly.

src/state/swap/index.ts
zzmp: I don't think there's a benefit - I don't think there's a large issue with changing this down the road.

PR #553, opened 2023-03-12
feat: perf event handlers

zzmp:
Adds new perf event handlers so that relevant events can be instrumented by integrators (eg to track performance metrics):
- `onSwapQuote`
- `onTokenAllowance`
- `onPermit2Allowance`
- `onSwapSend`
- `onWrapSend`

Deprecates existing handlers whose functionality is shared by the new handlers: `onSwapApprove`, `onSubmitSwapClick`.

Perf event handlers should record the entire event, so they all follow a similar pattern and use a new helper hook: `usePerfEventHandler`. All perf event hooks take two parameters: `args` and an `event`. The `event` is a promise that resolves (or rejects) when the instrumented event has completed. `usePerfEventHandler` wraps `event` in a resolved Promise, so that it is not executed until after the perf event handler has executed. This allows integrators to instrument the full event (eg even if it triggers a fetch).

---

INFRA-144


src/state/routing/slice.ts
```diff
@@ -23,9 +23,34 @@ const baseQuery: BaseQueryFn<GetQuoteArgs, TradeResult> = () => {
...
     getTradeQuote: build.query({
+      async onQueryStarted(args: GetQuoteArgs | SkipToken, { queryFulfilled }) {
+        if (args === skipToken) return
+
+        args.onQuote?.(
+          JSON.parse(serializeGetQuoteArgs(args)),
+          queryFulfilled
+            .catch((error) => {
+              const { error: queryError } = error
+              if (queryError && typeof queryError == 'object' && 'status' in queryError) {
```
just-toby, line 37: should this be a `===`

src/state/routing/slice.ts
```diff
@@ -23,9 +23,34 @@ const baseQuery: BaseQueryFn<GetQuoteArgs, TradeResult> = () => {
...
+                const parsedError = queryError as FetchBaseQueryError
+                switch (parsedError.status) {
+                  case 'CUSTOM_ERROR':
+                  case 'FETCH_ERROR':
+                  case 'PARSING_ERROR':
```
just-toby, line 42: type/enum these?

src/state/routing/slice.ts
```diff
@@ -23,9 +23,34 @@ const baseQuery: BaseQueryFn<GetQuoteArgs, TradeResult> = () => {
...
+                    throw parsedError.error
+                    break
+                  default:
+                    throw parsedError.status
+                }
+              }
+              throw queryError
```
just-toby, line 49: is there a way to indicate in `perf.ts` that these can throw certain types of errors? should the integrator be aware so they can handle them?

src/hooks/usePerfEventHandler.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useCallback } from 'react'
+import { swapEventHandlersAtom } from 'state/swap'
+import { PerfEventHandlers } from 'state/swap/perf'
+
+export function usePerfEventHandler<
+  K extends keyof PerfEventHandlers,
+  P extends Parameters<NonNullable<PerfEventHandlers[K]>>,
+  A extends P[0],
+  E extends Awaited<P[1]>,
+  H extends PerfEventHandlers[K] & ((args: A, event: Promise<E>) => void)
```
just-toby, line 11: can you add some comments to explain what these are

src/hooks/swap/useWrapCallback.tsx
```diff
@@ -43,19 +43,26 @@ export default function useWrapCallback(): UseWrapCallbackReturns {
...
+        return {
+          response: await wrappedNativeCurrencyContract.withdraw(`0x${parsedAmountIn.quotient.toString(16)}`),
+          type: TransactionType.WRAP,
+          amount: parsedAmountIn,
+        }
+      case undefined:
+        return undefined
     }
   }, [parsedAmountIn, wrappedNativeCurrencyContract, wrapType])
+  const callback = usePerfEventHandler('onWrapSend', wrapCallback, parsedAmountIn && { amount: parsedAmountIn })
```
just-toby, line 65: `{ amount: parsedAmountIn }` is inputs to `wrapCallback` right? kinda hard to tell without comments in `usePerfEventHandler` and without labelling these inputs

src/state/routing/slice.ts
zzmp: these are literal strings in FetchBaseQueryError, but I can instead check for `typeof parsedError.status === 'number'`

src/hooks/usePerfEventHandler.ts
zzmp: I can name these with words rather than letters 🤷

src/hooks/swap/useWrapCallback.tsx
zzmp: refactored to (hopefully) make this more clear. `usePerfEventHandler` is meant to mirror the arguments of the perf event handler, so I want it to be ordinal inputs (ie not an options object).

src/state/routing/slice.ts
zzmp: typechecking requires that this be done without checking for `=== 'number'` - this is the most elegant way I could get past the typechecker :(

src/state/routing/slice.ts
zzmp: I think a good follow up would be for _us_ to export an enum (or extend `Promise`) and only throw these promises for integration hooks, but I don't think that that should be a part of this PR. It could be a future improvement.

just-toby: sounds good - i was imagining some generic "throws" descriptor on the function , not even typing it. maybe something worth calling out in the docs if we update them for the PerfEventHandlers

just-toby (approved):

src/hooks/swap/useWrapCallback.tsx
just-toby: this makes sense now thanks

PR #554, opened 2023-03-12
test: add filter to event feed

zzmp:
Adds a filter to the cosmos event feed to aid debugging.

### DIFF
--- a/src/cosmos/EventFeed.tsx
+++ b/src/cosmos/EventFeed.tsx
@@ -1,5 +1,6 @@
 import { defaultTheme, SwapEventHandlers, TransactionEventHandlers, WidgetEventHandlers } from '@uniswap/widgets'
 import Row from 'components/Row'
+import { useState } from 'react'
 import styled from 'styled-components/macro'
 import * as Type from 'theme/type'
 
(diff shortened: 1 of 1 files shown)


src/cosmos/EventFeed.tsx
```diff
@@ -66,15 +70,32 @@ export default function EventFeed({ events, onClear }: EventFeedProps) {
           <Type.Subhead1>clear</Type.Subhead1>
         </button>
       </Row>
+      <Row>
+        <Type.Subhead2>Filter: </Type.Subhead2>
+        <select
+          onChange={(e) => {
+            console.log(e.target.value)
```
lynnshaoyu, line 77: is this console.log needed?

src/cosmos/EventFeed.tsx
```diff
@@ -66,15 +70,32 @@ export default function EventFeed({ events, onClear }: EventFeedProps) {
...
+            onSelectEventType(e.target.value)
+          }}
+          value={selectedEventType}
+        >
+          <option>{SHOW_ALL_EVENTS}</option>
+          {HANDLERS.map((name) => (
+            <option key={name}>{name}</option>
```
lynnshaoyu, line 84: any chance the handlers have the same name? could also use i (index) as key value for consistency with what you do on line 91

lynnshaoyu (commented):
some comments, not sure if they should be blocking or not so didnt approve yet🙃

src/cosmos/EventFeed.tsx
zzmp: leaving this as a runtime check - left a comment stating as such

lynnshaoyu (approved):

Issue #374, opened 2023-01-11
Could not find dependency: 'jotai-immer' when using SwapWidget


cnasc:
**Bug Description**
After installing `@uniswap/widgets` and importing `SwapWidget`, a `DependencyNotFound` error for `jotai-immer` throws.

**Steps to Reproduce**

[Codesandbox here](https://codesandbox.io/s/cool-sun-kridbg?file=/src/App.tsx)

1. install `@uniswap/widgets` per [the readme](https://github.com/Uniswap/widgets#installation)
2. try to use `SwapWidget`
3. error occurs

**Expected Behavior**
After following the installation steps in the readme, `SwapWidget` is usable without error.

**Additional Context**
![image](https://user-images.githubusercontent.com/9300702/211826839-6d4d4f3e-b8e9-4c84-b3e9-1cf08a04fb4c.png)

- noticed this at first on our Next.js app
- Tried in a clean react codesandbox to reproduce the issue and confirm it wasn't local to my machine
- manually adding `jotai-immer` and `qs` as dependencies to my project appears to resolve the issue

Tomiblanchard:
Same!

rharkor:
Same, and when I fix it it says error - Error: Cannot find module '***/node_modules/@uniswap/conedison/dist/format'

GitFede100:
**Had the same issue, solved by doing:**
```
-- npm i @uniswap/widgets@1.1.1
-- npm i react/redux
-- npm i jotai-immer
```

and it should work.

zyrm:
Same

just-toby:
see #404 for a suggestion on a fix/workaround. this should help resolve the conedison issue until we can update that library.

gilles-hemmerle:
> **Had the same issue, solved by doing:**
> 
> ```
> -- npm i @uniswap/widgets@1.1.1
> -- npm i react/redux
> -- npm i jotai-immer
> ```
> 
> and it should work.

This makes the project building properly and run.

However, in my application it seems that after this I get thousands of deprecated error occuring in an infinite loop.
Then the swap button is always greyed out. I don't know if it is linked but I see no other reason for this button to stay deactivated.

![image](https://user-images.githubusercontent.com/2291800/221542924-991fa717-d6af-4db6-ad6a-585160d43e3b.png)

dai-ki-web:
After manually adding jotai-immer,the console loops indefinitely warning:use `useSetAtom` from `jotai` instead. 
[...]

PR #555, opened 2023-03-13
fix: lock jotai version to 1.4.0

marcinciarka:
The current `"jotai": "^1.3.7"` works well when testing with the react-cosmos stack cause it's using the locked version from the project `yarn.lock` file (which is 1.4.0). When using the widget somewhere else (eg. anywhere tbh) it's using the newer version (1.13.1) which is _not compatible_.

This PR makes sure that the proper version (exact same as when running the project from this repo) is used and will resolve this issue: https://github.com/Uniswap/widgets/issues/374


piotrkonowrocki (approved):

vm:
we'll re-run the tests and then merge this 👍

just-toby (approved):

vm:
https://github.com/Uniswap/widgets/releases/tag/v2.47.10 should solve your issue. unfortunately, this branch's tests keep failing for an unknown reason, so i merged the same diff in #556. we'll figure that out separately. thanks for the effort here!

PR #556, opened 2023-03-13
fix: lock jotai version

vm:
This is the same as https://github.com/Uniswap/widgets/pull/555, but the tests won't run there. Making this to see if it's related to being from a non-organization contributor.


just-toby (approved):

marcinciarka:
Looks like it was due to outside help :)

PR #560, opened 2023-03-15
fix: use standardized WidgetPromise for perf event handlers

zzmp:
Track user rejections, and include them in perf events under a concrete type: `UserRejectedRequestError` (which extends `WidgetError`).
Reacts to the `UserRejectedRequestError` at the `ErrorBoundary` level, to avoid misuse in the future.
Wraps perf events in a `WidgetPromise` so that they are typed to reject with a `WidgetError`.


src/utils/jsonRpcError.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function isUserRejection(error: any): boolean {
+  const reason = getReason(error)
+  if (
+    error?.code === ErrorCode.USER_REJECTED_REQUEST ||
+    // These error messages have been observed in the listed wallets:
+    (reason?.match(/request/i) && reason?.match(/reject/i)) || // Rainbow
+    reason?.match(/declined/i) || // Frame
+    reason?.match(/cancell?ed by user/i) || // SafePal
```
zzmp, line 19: Note that I've updated this list to include both spellings of "cancelled", Trust, and Fireblocks. This brings the checks up-to-date with interface.

just-toby (changes requested):
i think this is looking good, but i have one request: can we explicitly ignore `UserRejectedError`s in the `ErrorBoundary` (or however you want to implement it) so we're completely sure that we don't show the Error Modal in response to user rejections? 

 with the current implementation, we have to catch these errors at each callsite. someone could miss this in the future (not to mention potentially missing one here).
 
 we've gotten more than enough feedback from internal testing to know that we should never be showing the Error modal UI in response to a rejection

just-toby (approved):

PR #565, opened 2023-03-16
fix: narrow WidgetPromise typings

zzmp:
Narrows `WidgetPromise` typings such that we can be certain we are only triggering the events with promises that will reject in a `WidgetError`.

This is primarily a refactor, but it adds type safety to the internals of the `PerfEventHandlers`.


mikeki:
Types look good, it's quite a complex typing though but seems like it is necessary to accomplish what you need.

just-toby (approved):

PR #567, opened 2023-03-17
feat: add success color to fiat value change

just-toby:
make the USD value change GREEN if it's positive / profitable

depends on: https://github.com/Uniswap/conedison/pull/17

### DIFF
--- a/src/hooks/usePriceImpact.ts
+++ b/src/hooks/usePriceImpact.ts
@@ -8,7 +8,7 @@ import { useUSDCValue } from './useUSDCPrice'
 
 export interface PriceImpact {
   percent: Percent
-  warning?: 'warning' | 'error'
+  warning?: 'warning' | 'error' | 'success'
 }
 
 export function usePriceImpact(trade?: InterfaceTrade): PriceImpact | undefined {
@@ -32,7 +32,8 @@ export function useFiatValueChange(trade?: InterfaceTrade) {
     }
     return {
       percent: fiatPriceImpact,
-      warning: getPriceImpactWarning(fiatPriceImpact),
+      // Fiat value should be green when it's positive to indicate a profitable trade.
+      warning: fiatPriceImpact.greaterThan(0) ? 'success' : getPriceImpactWarning(fiatPriceImpact),
     }
   }, [inputUSDCValue, outputUSDCValue])
 }
--- a/src/hooks/useSlippage.ts
+++ b/src/hooks/useSlippage.ts
@@ -1,4 +1,3 @@
-import { formatPriceImpact } from '@uniswap/conedison/format'
 import { CurrencyAmount, Percent, Token } from '@uniswap/sdk-core'
 import useAutoSlippageTolerance, { DEFAULT_AUTO_SLIPPAGE } from 'hooks/useAutoSlippageTolerance'
 import { useAtomValue } from 'jotai/utils'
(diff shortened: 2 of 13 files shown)


src/hooks/usePriceImpact.ts
```diff
@@ -32,7 +32,8 @@ export function useFiatValueChange(trade?: InterfaceTrade) {
     }
     return {
       percent: fiatPriceImpact,
-      warning: getPriceImpactWarning(fiatPriceImpact),
+      // Fiat value should be green when it's positive to indicate a profitable trade.
+      warning: fiatPriceImpact.greaterThan(0) ? 'success' : getPriceImpactWarning(fiatPriceImpact),
```
just-toby, line 36: we only want 'success' state for fiat value change, not the pool price impact

src/hooks/useSlippage.ts
```diff
@@ -53,5 +53,7 @@ export function getSlippageWarning(slippage?: Percent): 'warning' | 'error' | un
 }
 
 export function formatSlippage(slippage: Slippage): string {
-  return formatPriceImpact(slippage.allowed)
+  // Allowed slippage should always be displayed as a positive number,
+  // and conedison flips the sign when formatting price impact.
+  return formatPriceImpact(slippage.allowed.multiply(-1))
```
just-toby, line 58: i could just avoid conedison here, but it's nice to have consistent formatting?

src/utils/computeFiatValuePriceImpact.tsx
```diff
@@ -9,6 +9,10 @@ export function computeFiatValuePriceImpact(
   if (!fiatValueOutput || !fiatValueInput) return undefined
   if (!fiatValueInput.currency.equals(fiatValueOutput.currency)) return undefined
   if (JSBI.equal(fiatValueInput.quotient, JSBI.BigInt(0))) return undefined
-  const pct = ONE_HUNDRED_PERCENT.subtract(fiatValueOutput.divide(fiatValueInput))
+  // Example when output is greater than input:
+  // 1 - (100 / 80) = 1 - 1.25 = -0.25 = 25% profit (positive)
+  // Example when output is less than input:
+  // 1 - (80 / 100) = 1 - 0.8 = 0.2 = 20% loss (negative)
+  const pct = ONE_HUNDRED_PERCENT.subtract(fiatValueOutput.divide(fiatValueInput)).multiply(-1)
```
just-toby, line 16: this makes more sense if we flip the sign at the end

src/utils/prices.ts
```diff
@@ -23,9 +23,14 @@ export function computeRealizedPriceImpact(trade: InterfaceTrade): Percent {
   return trade.priceImpact.subtract(realizedLpFeePercent)
 }
 
-export function getPriceImpactWarning(priceImpact: Percent): 'warning' | 'error' | undefined {
+/**
+ * Returns the warning level for a given % change. The change can be either positive or negative.
```
tinaszheng, line 27: lets figure out the +- price impact issue first before committing this comment to code because it could be confusing

and also we should probably decouple using this helper function for both price impact and for fiat value change

src/utils/computeFiatValuePriceImpact.tsx
tinaszheng: image
dont think this is right

src/hooks/useSlippage.ts
tinaszheng: you could add a `formatSlippage` helper to conedison

tinaszheng (changes requested):

src/utils/prices.ts
just-toby: sounds good, removed from this PR

just-toby:
this is still a valid change, but closing since we're not prioritizing new widget work

PR #568, opened 2023-03-20
feat: new warning UI in toolbar

just-toby:
adds a warning UI to the expander on the right when the toolbar is collapsed  - IF there is a price impact warning to view in the expanded state.

Screenshot 2023-03-20 at 11 14 22 AM

Screenshot 2023-03-20 at 11 14 16 AM

### DIFF
--- a/src/components/Swap/PriceImpactRow.tsx
+++ b/src/components/Swap/PriceImpactRow.tsx
@@ -19,7 +19,9 @@ export function PriceImpactRow({ impact, reverse, tooltipText }: PriceImpactProp
   return (
     <Row gap={0.25} flex align="center" flow={reverse ? 'row-reverse' : 'row wrap'}>
       <ThemedText.Body2 userSelect={false} color={impact.warning ?? 'hint'}>
-        <TooltipText text={`(${formatPriceImpact(impact?.percent)})`}>{tooltipText}</TooltipText>
+        <TooltipText text={`(${formatPriceImpact(impact?.percent)})`}>
+          <ThemedText.Caption>{tooltipText}</ThemedText.Caption>
+        </TooltipText>
       </ThemedText.Body2>
       {impact?.warning && (
         <Tooltip icon={AlertTriangle} iconProps={{ color: impact.warning }} data-testid="alert-tooltip">
--- a/src/components/Swap/Toolbar/Caption.tsx
+++ b/src/components/Swap/Toolbar/Caption.tsx
@@ -1,16 +1,14 @@
 import { Trans } from '@lingui/macro'
 import { Placement } from '@popperjs/core'
-import { formatPriceImpact } from '@uniswap/conedison/format'
 import { Currency, CurrencyAmount } from '@uniswap/sdk-core'
 import Row from 'components/Row'
 import Tooltip from 'components/Tooltip'
 import { loadingCss } from 'css/loading'
-import { PriceImpact as PriceImpactType } from 'hooks/usePriceImpact'
 import { useIsWideWidget } from 'hooks/useWidgetWidth'
 import { AlertTriangle, ChevronDown, Icon, Info, LargeIcon, Spinner } from 'icons'
 import { ReactNode, useCallback } from 'react'
 import { InterfaceTrade } from 'state/routing/types'
-import styled from 'styled-components/macro'
+import styled, { css } from 'styled-components/macro'
 import { AnimationSpeed, Color, ThemedText } from 'theme'
 
 import Price from '../Price'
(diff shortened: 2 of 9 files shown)


src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -151,59 +174,38 @@ export function Trade({
   gasUseEstimateUSD,
   expanded,
   loading,
+  warning,
 }: TradeProps & TradeTooltip & ExpandProps) {
+  const widgetWidth = useWidgetWidth()
+  const shouldHideUSD = widgetWidth < 360 && warning && !expanded
```
just-toby, line 180: this behavior is based on a conversation with Fred  - at narrow widget width, we need to shorten the Trade Price string length so the new button fits

src/components/Swap/PriceImpactRow.tsx
```diff
@@ -19,7 +19,9 @@ export function PriceImpactRow({ impact, reverse, tooltipText }: PriceImpactProp
   return (
     <Row gap={0.25} flex align="center" flow={reverse ? 'row-reverse' : 'row wrap'}>
       <ThemedText.Body2 userSelect={false} color={impact.warning ?? 'hint'}>
-        <TooltipText text={`(${formatPriceImpact(impact?.percent)})`}>{tooltipText}</TooltipText>
+        <TooltipText text={`(${formatPriceImpact(impact?.percent)})`}>
+          <ThemedText.Caption>{tooltipText}</ThemedText.Caption>
```
zzmp, line 23: Is this change related to the warning UI? Is it necessary?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -139,10 +137,35 @@ export interface TradeProps {
 
 interface ExpandProps {
   expanded: boolean
-}
+  warning?: 'warning' | 'error'
+}
+
+const ExpanderRow = styled(Row)<{ $expanded: boolean; warning?: 'warning' | 'error' }>`
+  ${({ warning, $expanded }) =>
+    warning &&
```
zzmp, line 145: This is hard to follow because it's using defaulting - could you use an if-statement instead?

```
  ${({ warning, $expanded }) => {
    if (!warning) return undefined
    return css`
```

or

```
  if (warning) {
    return css`
      ...
    `
  }

File: src/components/Swap/Toolbar/Caption.tsx [hunk 4]
```diff
@@ -139,10 +137,35 @@ export interface TradeProps {
...
+`
 
-function Expander({ expanded }: ExpandProps) {
-  return <ExpandIcon $expanded={expanded} />
+function Expander({ expanded, warning }: ExpandProps) {
+  return (
+    <ExpanderRow $expanded={expanded} warning={warning} gap={0.5}>
+      {warning && !expanded && (
+        <Tooltip icon={AlertTriangle} iconProps={{ color: warning }} placement="auto">
+          <ThemedText.Caption>
```
zzmp, line 161: Is this needed, or is it already supplied by `Tooltip`?

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -139,10 +137,35 @@ export interface TradeProps {
...
+            <Trans>Your trade will have a high impact on the market price of this pool.</Trans>
+          </ThemedText.Caption>
+        </Tooltip>
+      )}
+      <ExpandIcon $expanded={expanded} color={expanded ? undefined : warning} />
+    </ExpanderRow>
+  )
 }
 
 export function Trade({
```
zzmp, line 171: Modifications to Trade should go in a separate PR, I think. Are they related?

src/components/Swap/Toolbar/Caption.tsx
zzmp: Is there a constant that can be used instead of `360`?
You should include a comment above this line explaining this behavior. A good rule of thumb is that if you leave a PR comment, it should probably be noted in the code.

zzmp (changes requested):

src/components/Swap/Toolbar/Caption.tsx
just-toby: this is needed

src/components/Swap/Toolbar/Caption.tsx
just-toby: updated to use a share constant that already exists

src/components/Swap/PriceImpactRow.tsx
just-toby: it's semi-related because all the tooltips should have consistent font size and the warning UI includes tooltips

src/components/Swap/Toolbar/Caption.tsx
```diff
@@ -151,59 +176,38 @@ export function Trade({
   gasUseEstimateUSD,
   expanded,
   loading,
+  warning,
 }: TradeProps & TradeTooltip & ExpandProps) {
+  const widgetWidth = useWidgetWidth()
+  const shouldHideUSD = widgetWidth < WIDGET_BREAKPOINTS.EXTRA_SMALL && warning && !expanded
```
zzmp, line 182: Add a comment above this line explaining why it is necessary to hide USD.

zzmp (approved):

PR #569, opened 2023-03-22
feat: upgrade sor

just-toby:
upgrades the Smart Order Router dependency because we want this bug fix: https://github.com/Uniswap/smart-order-router/pull/211

this fixes an issue Opensea is having with their integration related to `provider` compatibility.

-----

also, remove the dynamic import for the client side routing code. this code splitting is not working for third party integrators using next.js.


vm:
Need to run `yarn deduplicate`. I'm also not sure how to test thing. Is there a way to re-create the env OpenSea has locally?

vm (approved):

PR #571, opened 2023-03-24
feat: check lockfile changes during the setup action

just-toby:
adds a checksum check to the yarn.lock file during the install step.

this is needed because yarn's built-in `--frozen-lockfile` option isn't working in all the cases we need it to. see this issue: https://github.com/yarnpkg/yarn/issues/5840

### test plan:

the current version of the PR has an unnecessary `package.json` change just to make sure the script is working. i'll remove that change and fix the lock file before merging.


package.json
```diff
@@ -67,7 +68,7 @@
     "@uniswap/redux-multicall": "^1.1.8",
     "@uniswap/router-sdk": "^1.3.0",
     "@uniswap/sdk-core": "^3.2.0",
-    "@uniswap/smart-order-router": "^3.10.0",
+    "@uniswap/smart-order-router": "^3.9.0",
```
cmcewen, line 71: did you mean to revert this?

scripts/yarn-install.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
cmcewen, line 1: i'm generally open to scripts but i think sometimes bash is a little harder for people and you may run into env issues. also - i've seen this in the past where people do a general git diff to check for changes instead of yarn install, which may catch things like autogenerated changes from types, etc. do you have a preference between checksums and git diffs?

package.json
just-toby: i plan to revert this before merging, this is just to prove that it's working

scripts/yarn-install.sh
just-toby: now that you mention it, i'm in favor of using `git diff` because it will help us deal with gql codegen issues in the interface repo too.

.github/actions/setup/action.yml
```diff
@@ -14,8 +14,11 @@ runs:
...
         path: node_modules/
-        key: ${{ runner.os }}-install-${{ hashFiles('**/yarn.lock') }}
+        key: ${{ runner.os }}-install-${{ hashFiles('**/yarn.lock', '**/package.json') }}
 
     - if: steps.install-cache.outputs.cache-hit != 'true'
-      run: yarn install --frozen-lockfile --ignore-scripts
+      run: yarn install --ignore-scripts
+      shell: bash
+
+    - run: git diff --quiet -- yarn.lock
```
cmcewen, line 23: i think the intended behavior we want is to fail this step if there are any git diffs (throw an error). will that happen with this command?

just-toby: yes, i added a comment to clarify

just-toby:
this is still a valid change, but i'm closing for now since we're not prioritizing widget work

PR #574, opened 2023-04-05
fix: replace missing return value from swap callback

just-toby:
the callback we pass to `onSubmit` requires the transaction info be returned if successful. this was removed in #553 , probably because it seemed unnecessary? cc @zzmp 

this re-adds the missing return statement with a comment indicating it's required

### test plan

verified that the StatusDialog appears properly now when you submit a transaction


zzmp (approved):
Seems like a mistake that it was removed.
Should there be a test confirming that it is returned?

just-toby:
@zzmp i'll create a ticket for writing a test for this and it can get prioritized against interface work

talonZack:
Thank ya'll was struggling with the transaction submission and success callback

PR #578, opened 2023-04-13
feat: add swap summary info to OnReviewSwapClick

talonZack:
Trying to add a non-hack way to ensure the user has adequate slippage set in the case of something like a fee on transfer token


eastedgeZack:
Anybody there for review?

PR #579, opened 2023-04-19
fix: upgradable web3-react

zzmp:
Upgrades web3-react to the stable 8.2.0 version, and allows higher versions to replace it (`^8.2.0`).
There were no functional changes between `8.1.2-beta.0` and `8.2.0`, so this should be a nop.

### Test plan

Manually confirmed wallet connection still works.


grabbou (approved):
I confirm no changes were made!

PR #604, opened 2023-06-29
feat(wallet-connect): add support for wcv2 

JFrankfurt:
retires walletconnect v1 in favor of v2


src/cosmos/useProvider.ts
```diff
@@ -12,15 +13,22 @@ enum Wallet {
...
       actions,
       options: {
-        rpc: Object.entries(JSON_RPC_FALLBACK_ENDPOINTS).reduce((rpcMap, [chainId, rpcUrls]) => ({
+        rpcMap: Object.entries(JSON_RPC_FALLBACK_ENDPOINTS).reduce((rpcMap, [chainId, rpcUrls]) => ({
           ...rpcMap,
           [chainId]: rpcUrls.slice(0, 1),
         })),
+        showQrModal: true,
+        projectId: WALLET_CONNECT_PROJECT_ID,
+        chains: [SupportedChainId.MAINNET],
```
zzmp, line 29: This will refuse to connect to wallets which do not exist on MAINNET (eg a Safe that was created on an L2).
Can you leave that as a comment, so it's at least a known issue?

zzmp (approved):

PR #615, opened 2023-08-11
feat: base and avax

just-toby:
adds support for base network

also removes the locally-defined SupportedChainId and replaces it with the `ChainId` enum from `sdk-core`


JFrankfurt:
image
Looks like the QR that should be rendering here is missing now

JFrankfurt:
I'm unable to trigger walletconnect in the preview build

Issue #620, opened 2023-09-15
Sepolia integration


simonmullaney:
We are implementing the Uniswap widget on our Development environment that runs on the Sepolia network. The Uniswap widget does not work with this network.

When we use the Sepolia network we are only given the option to switch back to Ethereum mainnet network.
 
We would like the widget to work with the Sepolia network in the same fashion as it does with other testnet networks like Mumbai etc.

Is this a feature that will be implemented soon? If not is there any way to add ourselves manually?

Many thanks,
Simon

duccadhv1:
i have same problem. any update?

LunaJet:
Running into the same problem

xrdavies:
any update about sepolia?

fabdarice:
following

waqaswahid1:
following

steoniy:
same issue

KevinMiller97:
Any news? There is still no support for Sepolia after over a year.

alexflorence:
Could this me merged / fixed please? It is not consistent that Sepolia is the only supported testnet by the Uniswap UI (https://support.uniswap.org/hc/en-us/articles/14580495154445-Testnets-on-Uniswap) but its not supported by the swap widget.

PR #635, opened 2024-06-21
Added support for sepolia Chain | issue : Sepolia integration #620

kritik-sah:
# Sepolia integration #620
We are implementing the Uniswap widget on our Development environment that runs on the Sepolia network. The Uniswap widget does not work with this network.

When we use the Sepolia network we are only given the option to switch back to Ethereum mainnet network.

We would like the widget to work with the Sepolia network in the same fashion as it does with other testnet networks like Mumbai etc.


steoniy:
With goerli deprecated most platforms test on Sepolia - can we get this in motion pls

longmup:
Sepolia cannot intergrate with uniswap/widget? 
Please @mr-uniswap , have any news for this?

PR #636, opened 2024-07-03
chore(infra): add npm deploy

hodl4fun:
Updates the npm deploy to use the new token


just-toby (changes requested):
can you explain why we need this

mr-uniswap:
@just-toby Sorry, modified this to use the new npm token

